PR #5, opened 2024-04-12
Create translation.json

vanja-san:
ru


hydralauncher:
@vanja-san we changed the structure of the repo, so we do not force push the main branch no more. Could you please possibly make a new pull request with those changes? That would be very appreaciated! Thank you for your contribution and sorry for the inconvinience.

PR #12, opened 2024-04-14
Add back button on game's page

LuanRoger:
- [x] Create the button on the games page
- [ ] Possible to back to search result when it come from there

### Demo
demo


hydralauncher:
This implementation was used as part of #10

PR #14, opened 2024-04-14
Added modal to warning user if Wine or Lutris is not installed

fhilipecrash:
- Refactor `openGame` function with early returns because this function have too much chained if else
- Using `exec` function instead `spawn` to start Wine/Lutris process
- Added Wine binary verification
- Added the `BinaryNotFoundModal` component to inform the user that they do not have the necessary programs installed
- New `i18n` section to new modal

Note: Checking whether the binaries exist in the system's `$PATH` is still occurs with the `spawnSync` function because it simply returns 0 if it was found and 1 if it was not, using the `exec` function it throws an exception and I see no reason to handle this.


src/renderer/pages/shared-modals/binary-not-found-modal.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <Modal 
+      visible={visible}
+      title={t("title")}
+      description={t("description")}
+      onClose={onClose}
+    >
+      {t("instructions")}
+    </Modal>
+  )
+}
```
hydralauncher, line 25: Please add a newline

src/renderer/pages/game-details/hero-panel.tsx
```diff
@@ -46,7 +49,8 @@ export function HeroPanel({
   const isGameDownloading = isDownloading && gameDownloading?.id === game?.id;
 
   const openGame = (gameId: number) =>
-    window.electron.openGame(gameId).then(() => {
+    window.electron.openGame(gameId).then(res => {
```
hydralauncher, line 52: Please use a more descriptive name for `res`

src/renderer/pages/game-details/hero-panel.tsx
```diff
@@ -27,6 +28,8 @@ export function HeroPanel({
 }: HeroPanelProps) {
   const { t } = useTranslation("game_details");
 
+  const [showBinaryNotFoundModal, setShowBinaryNotFoundModal] = useState<boolean>(false);
```
hydralauncher, line 31: No need tao pass `boolean` as a generic, since TypeScript can infer the type from the default value

src/renderer/pages/downloads/downloads.tsx
```diff
@@ -37,7 +39,8 @@ export function Downloads() {
   }, [library]);
 
   const openGame = (gameId: number) =>
-    window.electron.openGame(gameId).then(() => {
+    window.electron.openGame(gameId).then(res => {
```
hydralauncher, line 42: Please use a more descriptive name than `res`

src/renderer/pages/downloads/downloads.tsx
```diff
@@ -18,6 +19,7 @@ export function Downloads() {
   const navigate = useNavigate();
 
   const [filteredLibrary, setFilteredLibrary] = useState<Game[]>([]);
+  const [showBinaryNotFoundModal, setShowBinaryNotFoundModal] = useState<boolean>(false);
```
hydralauncher, line 22: TypeScript can infer boolean from the default value

.github/workflows/sync.yml
```diff
@@ -0,0 +1,26 @@
+name: Sync
```
hydralauncher, line 1: You can remove this file when syncing with main; no longer necessary

hydralauncher (changes requested):

PR #22, opened 2024-04-15
fix: dropping algolia usage

### LINKED ISSUE #20: Find another way to scrap through Steam games
SteamDB seems to be aware of what Hydra's currently doing. Their website traffic might not be matching with the amount of requests that their Algolia account is currently receiving.

For now, they simply started to use a base64 encoded string as the Algolia API Key:

image

Even a small change like this is enough to break Hydra's search routine.

ghost:
This closes #20

There's a LOT of tests to be made regarding this feature


hydralauncher:
Closing this for now. Steam search mechanism doesn't suffice for Hydra.

PR #26, opened 2024-04-15
build: add linux .desktop entry

fhilipecrash:
- Change MakerRpm and MakerDeb functions to create a valid .desktop file


hydralauncher:
LGTM

PR #27, opened 2024-04-16
feat: move surprise me button to header

fernandozanutto:
- Moves the `✨ Surprise me` button to app header, so it is always visible, allowing the user to continuously navigate to random games

image


hydralauncher:
Please discuss these implementations on Hydra's Discord server:
https://discord.gg/hydralauncher

For now, I don't like this idea of moving to the header much. It's only useful on the catalogue/game details context.

PR #28, opened 2024-04-16
feat: update README.md to add note about Python 3.9

fernandozanutto:
- Updates README.md to add a note about why Python 3.9 is important to Windows users


hydralauncher:
Python 3.9 is MANDATORY for any OS in order to run Hydra. I see no need for this disclaimer since it's already marked in the README.

PR #40, opened 2024-04-18
refactor: remove pretty bytes

JackEnx:
pretty bytes show KiB, MiB, GiB... format in base 2 values, it's correct but is usual to see the format in KB, MB, GB...

Brave
image

Edge
image

Windows
image

Github
image


hydralauncher:
LGTM

PR #52, opened 2024-04-21
Save database to userData instead of appData

thwwy:
According to Electron's documentation, the `userData` folder is the place to store configuration files, and, by default, it already is the `appData` appended with the app name, so this PR should not move the file to another folder. (source: https://www.electronjs.org/docs/latest/api/app#appgetpathname)

The problem of doing the concatenation manually, is that it prevents us to launch the app with the `--user-data-dir` (to save data in a different folder than %APPDATA%), as all the other electron files go to the provided user data directory, except for the hydra.db file, as it does not use the userDir parameter.

e.g.
|Command|Before this PR|After this PR|*|
|-----|----|----|-----|
|Hydra.exe|%APPDATA%\Hydra\\\<electron files> <br> %APPDATA%\Hydra\hydra.db|%APPDATA%\Hydra\\\<electron files> <br> %APPDATA%\Hydra\hydra.db|No change|
|Hydra.exe <br>&nbsp; --user-data-dir=D:\Data| D:\Data\\\<electron files> <br> %APPDATA%\Hydra\hydra.db|D:\Data\\\<electron files> <br> D:\Data\\hydra.db||

### DIFF
--- a/src/main/constants.ts
+++ b/src/main/constants.ts
@@ -46,8 +46,7 @@ export enum GameStatus {
 export const defaultDownloadsPath = path.join(os.homedir(), "downloads");
 
 export const databasePath = path.join(
-  app.getPath("appData"),
-  app.getName(),
+  app.getPath("userData"),
   "hydra.db"
 );
 


hydralauncher:
I think this makes perfect sense, but this migration is not this trivial. If you suddenly update the database path, every single Hydra user from previous versions will have their databases wiped out when they update.

Please consider making a *migration script* to support the migration from appdata to userdata.

hydralauncher:
@thwwy do you have plans to work on this? We'll be closing the roadmap for 1.1.1 today.

thwwy:
@hydralauncher I did make a table that shows that the path does not change at all. Why would we need a migration if the file is at the same place?

PR #66, opened 2024-04-22
Update translation.json

lepsk:
Fixed English Translation


hydralauncher:
I don't see any improvement on simply capitalizing every single string and most translations are logically incorrect (assuming that all downloads are .rar, for instance). Closing for now.

PR #79, opened 2024-04-24
process watcher performance

zamitto:


src/main/helpers/ps.ts
```diff
@@ -1,12 +1,32 @@
...
-export const getProcesses = async () => {
-  if (process.platform === "win32") {
-    return tasklist().then((tasks) =>
-      tasks.map((task) => ({ ...task, name: task.imageName }))
-    );
-  }
+const TEN_MEGABYTES = 1000 * 1000 * 10;
+const execFile = promisify(childProcess.execFile);
+
+export const getProcesses = async (isPackaged: boolean) => {
```
hydralauncher, line 9: Why don't you access `app.isPackaged` from this function directly instead of passing it as an argument? I think you should be good.

src/main/helpers/ps.ts
```diff
@@ -1,12 +1,32 @@
 import psList from "ps-list";
-import { tasklist } from "tasklist";
+import path from "node:path";
```
hydralauncher, line 2: Could you please uninstall `tasklist`? This library will no longer be necessary.

postinstall.py
```diff
@@ -0,0 +1,9 @@
+import shutil
+import platform
+import os
+
+if platform.system() == "Windows":
```
hydralauncher, line 5: I think you can move this script to a simply common js script instead. It doesn't really matter to be honest, but it would make more sense considering that this script will be executed from the node package manager

hydralauncher (changes requested):

src/main/helpers/ps.ts
zamitto: sure, done

postinstall.py
zamitto: makes sense

hydralauncher (approved):
LGTM

PR #82, opened 2024-04-24
[Locales] Fix french translation

### LINKED ISSUE #80: Problem with the french translation
Hi, while reviewing a PR I noticed that in the french translation, the home screen is translated "meson", which would be correct if you were talking about a home as the building.
I'm not a french native speaker but that looked off to me so i did some research.
Apparently in software development the home screen is referred as "Accueil".

Magrid0:
Hi, while reviewing a PR I noticed that in the french translation, the home screen is translated "maison", which would be correct if you were talking about a home as the building.
I'm not a french native speaker but that looked off to me so i did some research.
Apparently in software development the home screen is referred as "Accueil".

[opinion of a french native speaker](https://www.reddit.com/r/FrenchLearning/s/CRlWAGydKU)

Fix #80

### DIFF
--- a/src/locales/fr/translation.json
+++ b/src/locales/fr/translation.json
@@ -25,7 +25,7 @@
     "downloads": "Téléchargements",
     "search_results": "Résultats de la recherche",
     "settings": "Paramètres",
-    "home": "Maison"
+    "home": "Accueil"
   },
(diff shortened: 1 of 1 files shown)


hydralauncher (approved):
LGTM

PR #84, opened 2024-04-25
fix: fix french terms in translation

Netflixyapp:
Fixed some terms that were not having the right meaning in french. Using "repack" in french is a more direct translation that adopts the English term "repack". It's understood by gamers familiar with the concept.


Magrid0 (approved):
Looks good to me

hydralauncher (approved):
LGTM

oussama1touda:
bro is not oben is lancher netflixy chroome is bannde

PR #92, opened 2024-04-28
Feat/enhance get random game

zamitto:
### Refactor logic of getting random games

1. On the first call of getRandomGame, Hydra makes all the requests to steam250 pages
2. Removes duplicates games from the list
3. Removes games that are not indexed on Hydra
4. And shuffles the resulting list
5. Now, on each new call to getRandomGame, there's no need to make a new request, it just gets the next game on the list
6. If the end of the list is reached, it is shuffled again and moves the index to the start of the list

### Other adjustments
- Makes the Surprise me button disabled until the game details page finishes loading, preventing a click spam on the button that could cause some strange behaviours on the page

### DIFF
--- a/src/main/events/catalogue/get-random-game.ts
+++ b/src/main/events/catalogue/get-random-game.ts
@@ -4,37 +4,35 @@ import { Steam250Game, getSteam250List } from "@main/services";
 
 import { registerEvent } from "../register-event";
 import { searchGames, searchRepacks } from "../helpers/search-games";
-import { formatName } from "@main/helpers";
 
 let gamesList: Steam250Game[] = [];
 let nextGameIndex = 0;
 
 const getRandomGame = async (_event: Electron.IpcMainInvokeEvent) => {
   if (gamesList.length == 0) {
-    console.log("fetching steam 250 pages");
-    gamesList = shuffle(await getSteam250List());
-  } else {
-    console.log("getting cached list");
+    const steam250List = await getSteam250List();
+
+    const filteredSteam250List = steam250List.filter((game) => {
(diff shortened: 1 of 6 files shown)


src/main/events/catalogue/get-random-game.ts
```diff
@@ -1,27 +1,41 @@
 import shuffle from "lodash/shuffle";
 
-import { getRandomSteam250List } from "@main/services";
+import { Steam250Game, getSteam250List } from "@main/services";
 
 import { registerEvent } from "../register-event";
 import { searchGames, searchRepacks } from "../helpers/search-games";
-import { formatName } from "@main/helpers";
+
+let gamesList: Steam250Game[] = [];
```
hydralauncher, line 8: Consider using an object instead:
```ts
const state = { games: [], index: 0 };
```

I think it would be clearer, and you wouldn't use variables

zamitto: done

src/main/events/catalogue/search-games.ts
```diff
@@ -1,11 +1,15 @@
...
-  {
-    name: "searchGames",
-    memoize: true,
-  }
-);
+const searchGamesEvent = async (
+  _event: Electron.IpcMainInvokeEvent,
+  query: string
+): Promise<CatalogueEntry[]> => {
+  return Promise.all(searchGames({ query, take: 12 }));
```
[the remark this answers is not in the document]
zamitto: done

hydralauncher (approved):
LGTM

PR #93, opened 2024-04-28
feat: add debrid services caching/downloading support

Netflixyapp:
This PR adds support for Debrid services to cache and download games/torrents

Debrid services are used to generate direct HTTP/S links for torrent files, this makes downloading games much faster since it doesn't rely on seeders anymore. It also improves security in countries or places where torrenting is illegal or traceable.

In this PR, I first added a settings section in the launcher to enable/disable using Debrid services, as well as an input field to specify an API key for Real-Debrid (Eventually, more services can be added soon)
On Python's end, I separated the _libtorrent_ logic into a separate class (TorrentDownloader) and added a new class (DebridDownloader) that handles communication with Debrid services as well as downloading from HTTP/S links using _pySmartDL_ library
A new state was added to the front end, which is DebridCaching. When using Debrid services, you usually get a direct link for a torrent file almost instantly when it's cached in the servers. However, when using a new torrent, Debrid services take a while to download the torrent on the servers to be able to generate a direct link. During this time, DebridCaching status is sent, alongside the progress of this caching process.


src/main/events/torrenting/resume-game-download.ts
```diff
@@ -1,6 +1,6 @@
 import { registerEvent } from "../register-event";
 import { GameStatus } from "../../constants";
-import { gameRepository } from "../../repository";
+import {gameRepository, userPreferencesRepository} from "../../repository";
```
hydralauncher, line 3: Are you using prettier? Please use prettier as a formatter or run `npm run format` before uploading
There should be whitespaces in between the brackets

src/main/services/torrent-client.ts
```diff
@@ -22,6 +22,7 @@ enum TorrentState {
   Downloading = 3,
   Finished = 4,
   Seeding = 5,
+  DebridCaching = 6,
```
hydralauncher, line 25: Make sure not to mix this with `libtorrent` statuses
If you want to create such thing as a virtual status, use a constant that is not being used by libtorrent
https://libtorrent.org/reference-Torrent_Status.html

hydralauncher (changes requested):

src/main/events/torrenting/resume-game-download.ts
Netflixyapp: Done ✅

src/main/services/torrent-client.ts
Netflixyapp: Done ✅

PR #128, opened 2024-04-30
Add check for null or undefined in get-random-game.ts 

eliascoding365:
Check condition to prevent potential runtime errors


mades-g:
Heey, I think we the check is done there, might as well do it here too.
https://github.com/hydralauncher/hydra/pull/128/files#diff-545fb414b7456cf9f81090710e21e4c6b14fd0d960f8e544e292551b9c14533bL11

eliascoding365:
Nice, thank you for pointing that out

zamitto:
These checks don't really make sense
The `state` object and its properties are being defined and initialized on line 8
They will never be null or undefined

PR #137, opened 2024-04-30
feat: Option to choose the number of results on the Catalog page in the configuration menu.

Dryrtan:
My idea initially was to do the same with the search page, but my knowledge in Electro is limited and prevented me from doing this, but it is open to anyone who wants to implement it on the search page.

### DIFF
--- a/src/renderer/src/pages/catalogue/catalogue.tsx
+++ b/src/renderer/src/pages/catalogue/catalogue.tsx
@@ -38,17 +38,10 @@ export function Catalogue() {
 
   useEffect(() => {
     async function fetchUserPreferences() {
-      try {
-        const [, userPreferences] = await Promise.all([
-          window.electron.getDefaultDownloadsPath(),
-          window.electron.getUserPreferences(),
-        ]);
+        const [ userPreferences ] = await Promise.all([window.electron.getUserPreferences()]);
         const resultsPerPage = userPreferences?.resultsPerPage ?? 30;
         setPaginationSettings({ resultsPerPage });
-      } catch (error) {
-        console.error("Failed to fetch user preferences:", error);
-      }
-    };
+    }
 
     fetchUserPreferences();
   }, []);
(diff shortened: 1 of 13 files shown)


hydralauncher:
Could you attach a screenshot?

Dryrtan:
> Thanks for the contribution! Could you please attach a screenshot?

Of course

resultsPerPage

zamitto:
These radio button seems more like check boxes. I think they should have a round style that radio buttons usually have

Dryrtan:
> These radio button seems more like check boxes. I think they should have a round style that radio buttons usually have

Alright, I'll make the change

Dryrtan:
> These radio button seems more like check boxes. I think they should have a round style that radio buttons usually have

The new style of the radio component

image

src/renderer/src/pages/catalogue/catalogue.tsx
```diff
@@ -34,13 +34,34 @@ export function Catalogue() {
     navigate(`/game/${game.shop}/${game.objectID}`);
   };
 
+  const [paginationSettings, setPaginationSettings] = useState({ resultsPerPage: 30 });
+
+  useEffect(() => {
+    async function fetchUserPreferences() {
+      try {
+        const [, userPreferences] = await Promise.all([
+          window.electron.getDefaultDownloadsPath(),
```
zamitto, line 43: do you need to call getDefaultDownloadsPath? you dont seem to be using its result

src/renderer/src/pages/catalogue/catalogue.tsx
```diff
@@ -34,13 +34,34 @@ export function Catalogue() {
     navigate(`/game/${game.shop}/${game.objectID}`);
   };
 
+  const [paginationSettings, setPaginationSettings] = useState({ resultsPerPage: 30 });
+
+  useEffect(() => {
+    async function fetchUserPreferences() {
+      try {
```
zamitto, line 41: i'm not sure if this try catch block is needed

Dryrtan:
Alright, refactoring...

hydralauncher:
Closing this in wishes for a more intuitive approach to handle this.

PR #151, opened 2024-05-01
add most played, trending 7day, trending 30days, trending quarterly.

gylli251:
image

I felt like something was missing with only having 30 day as the trending tab.

I added most played, then trending based on days.

Let me know if you want to implement this!

maybe this belongs in the catalog instead.

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -2,7 +2,10 @@
   "home": {
     "featured": "Featured",
     "recently_added": "Recently added",
-    "trending": "Trending",
+    "trending_7day": "Weekly Trending",
+    "trending_30day": "Monthly Trending ",
+    "trending_90day": "Quarter Trending",
+    "most_played": "Most Played",
     "surprise_me": "Surprise me",
     "no_results": "No results found"
   },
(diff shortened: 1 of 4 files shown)


Hachi-R:
I liked your idea, very good.

I think that if it ends up like this, the redundancy of image 1 should be removed, keeping the text only on the button.

or perhaps it would be better to remove the buttons there and place them in a dropdown menu that could be accessed by clicking on the heading

just ideas

image
image

hydralauncher:
Could you attach a screenshot of how it looks when you shrink the width of the window down?

gylli251:
image

This is when i shrink it to the minimum

gylli251:
> I liked your idea, very good.
> 
> I think that if it ends up like this, the redundancy of image 1 should be removed, keeping the text only on the button.
> 
> or perhaps it would be better to remove the buttons there and place them in a dropdown menu that could be accessed by clicking on the heading
> 
> just ideas
> 
> image image

i agree, i think having the "weekly trending" twice seems pointless.

hydralauncher:
@gylli251 please find a better way to display them when the screen is small like that. Looks good overall though

gylli251:
Not sure what to change.

do you mean that it shows weekly then newline trending?
i think its better this way if we make them truncate together the boxes its much more ugly.

this is what it normally looks like:
image

chubbygrannychaser:
Closing this. The idea has potential but it should be groomed a little bit better.

PR #154, opened 2024-05-01
Feat/format playtime in hours

zamitto:
- Changes how the playtime is formated to show in minutes (until 120 minutes) and in hours
- In the previous version of this PR I had increased the sleep time of process-watcher loop to 5 seconds. But we can't right now increase it too much as the game-details use this event to update the isGamePlaying status.

image

image


src/renderer/src/pages/game-details/hero-panel.tsx
```diff
@@ -47,27 +47,29 @@ export function HeroPanel({
...
-      return () => {
-        clearInterval(interval);
-      };
-    }
-  }, [game?.lastTimePlayed, updateLastTimePlayed]);
+      if (minutes < 120) {
+        return minutes.toFixed(0) + " " + t("minutes")
+      }
+
+      const numberFormat = new Intl.NumberFormat(i18n.language, { maximumFractionDigits: 1 })
```
hydralauncher, line 68: You should create the `NumberFormat` instance outside of the component so it doesn't have to instantiated every single time

src/renderer/src/pages/game-details/hero-panel.tsx
```diff
@@ -47,27 +47,29 @@ export function HeroPanel({
...
+
+      const hours = minutes / 60
+      return numberFormat.format(hours)  + " " + t("hours")
```
hydralauncher, line 71: Please use i18n interpolation instead:
```ts
t("playtime", { hours: minutes / 60 })
```

And on the translation file:
```json
{
"playtime": "{{hours}} hours"
}
```

This way you can handle multiple languages better

src/renderer/src/pages/game-details/hero-panel.tsx
```diff
@@ -47,27 +47,29 @@ export function HeroPanel({
...
+  const formatPlayTime = (milliseconds: number) => {
+      const seconds = milliseconds / 1000
+      const minutes = seconds / 60
 
-      return () => {
-        clearInterval(interval);
-      };
-    }
-  }, [game?.lastTimePlayed, updateLastTimePlayed]);
+      if (minutes < 120) {
```
hydralauncher, line 64: `120` is a magic number here. Please assign it to a constant outside of the function

src/renderer/src/pages/game-details/hero-panel.tsx
```diff
@@ -47,27 +47,29 @@ export function HeroPanel({
...
+        return minutes.toFixed(0) + " " + t("minutes")
```
hydralauncher, line 65: Consider using the same logic as I have mentioned for the hours

hydralauncher (changes requested):

src/renderer/src/pages/game-details/hero-panel.tsx
zamitto: Not sure what's the best way to solve this as I don't know much of React, but I used a `useMemo` hook and put the `NumberFormat` inside it, what do you think?

hydralauncher (approved):
LGTM

PR #155, opened 2024-05-01
Allow us to start games on linux using hydra and lutris.

gylli251:
Allow us to start games on linux using hydra and lutris.

So far just works, if somebody else can test it with me it would be great.

it uses the lutris runtime thats default to run the executable. sadly it needs to make the .exe execute permissions but don't see a way around that.

also changed the way the select and executable its not windows only.

Note that maybe its better to do the detection of what OS is running some way else then using "which lutris"


src/main/events/library/open-game.ts
```diff
@@ -1,16 +1,42 @@
...
 
 const openGame = async (
   _event: Electron.IpcMainInvokeEvent,
   gameId: number,
   executablePath: string
 ) => {
-  await gameRepository.update({ id: gameId }, { executablePath });
+  try {
+    await gameRepository.update({ id: gameId }, { executablePath });
+    await exec(`chmod +x "${executablePath}"`);
```
lilezek, line 21: Not an expert, but Isn't this a breaking change in windows? Is this code path executed only on linux?

src/renderer/src/pages/game-details/hero-panel-actions.tsx
```diff
@@ -47,7 +47,7 @@ export function HeroPanelActions({
         filters: [
           {
             name: "Game executable",
-            extensions: window.electron.platform === "win32" ? ["exe"] : [],
+            extensions: ["exe"],
```
hydralauncher, line 50: Not sure about this one. Not all of the game binaries are `.exe` on Windows

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
...
+    }
+  } catch (err) {
+    console.error('Failed operation:', err);
+  }
 };
 
 registerEvent(openGame, {
   name: "openGame",
-});
+});
```
hydralauncher, line 52: No newline

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
...
+        } catch (err) {
+          console.error('Error launching with Lutris:', err);
+        }
+      }
+    } else {
+      // For non-Linux systems, just open the path
+      shell.openPath(executablePath);
+    }
+  } catch (err) {
+    console.error('Failed operation:', err);
```
hydralauncher, line 46: Don't use `console.error` use the `logger` instead

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
...
+    if (os.platform() === 'linux') {
+      const { stdout: lutrisPath } = await exec('which lutris');
+      if (lutrisPath.trim()) {
+        // Lutris is installed
+        await exec(`chmod +x "${executablePath}"`); // Make executable only if Lutris is found
+        const escapedPath = escapeShellPath(executablePath);
+        try {
+          await exec(`lutris -e "${escapedPath}"`); // Use Lutris to open the game
+        } catch (err) {
+          console.error('Error launching with Lutris:', err);
```
hydralauncher, line 38: Don't use `console.error` use the `logger` instead

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
...
   gameId: number,
   executablePath: string
 ) => {
-  await gameRepository.update({ id: gameId }, { executablePath });
+  try {
+    await gameRepository.update({ id: gameId }, { executablePath });
 
-  shell.openPath(executablePath);
+    // Check for Lutris only on Linux systems
+    if (os.platform() === 'linux') {
```
hydralauncher, line 29: You can use `process.platform` here, it'll generate less changes since you don't have to import the `os` and it'll work just as fine

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
...
+        await exec(`chmod +x "${executablePath}"`); // Make executable only if Lutris is found
+        const escapedPath = escapeShellPath(executablePath);
+        try {
+          await exec(`lutris -e "${escapedPath}"`); // Use Lutris to open the game
+        } catch (err) {
+          console.error('Error launching with Lutris:', err);
+        }
+      }
+    } else {
+      // For non-Linux systems, just open the path
```
hydralauncher, line 42: No need for this comment here

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
...
-  await gameRepository.update({ id: gameId }, { executablePath });
+  try {
+    await gameRepository.update({ id: gameId }, { executablePath });
 
-  shell.openPath(executablePath);
+    // Check for Lutris only on Linux systems
+    if (os.platform() === 'linux') {
+      const { stdout: lutrisPath } = await exec('which lutris');
+      if (lutrisPath.trim()) {
+        // Lutris is installed
```
hydralauncher, line 32: I see no need for this comment

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
...
-
 import { registerEvent } from "../register-event";
 import { shell } from "electron";
+import { promisify } from 'util';
+import os from 'os';
+
+const exec = promisify(execCallback);
+
+const escapeShellPath = (path: string) => {
+  if (os.platform() === 'win32') {
```
hydralauncher, line 11: Use `process.platform` instead

src/main/events/library/open-game.ts
```diff
@@ -1,18 +1,52 @@
+import { exec as execCallback } from 'child_process';
 import { gameRepository } from "@main/repository";
-
 import { registerEvent } from "../register-event";
 import { shell } from "electron";
+import { promisify } from 'util';
```
hydralauncher, line 5: Import from `node:util` instead

hydralauncher (changes requested):
Please remove left comments

src/renderer/src/pages/game-details/hero-panel-actions.tsx
gylli251: Do you have any examples? We could have no filter but i think it makes the experience annoying.

hydralauncher: Not really to be honest, but I think that it's safer to left it open wide on anything else than Windows

hydralauncher (approved):
LGTM

gylli251:
Done. not sure about the merge conflicts can you take a look @hydralauncher ?

hydralauncher:
@gylli251 could you resolve them?

src/main/events/library/open-game.ts
```diff
@@ -1,16 +1,45 @@
...
 ) => {
-  await gameRepository.update({ id: gameId }, { executablePath });
+  try {
+    await gameRepository.update({ id: gameId }, { executablePath });
 
-  shell.openPath(executablePath);
+    if (process.platform === 'linux') {
+      const { stdout: lutrisPath } = await exec('which lutris');
+      if (lutrisPath.trim()) {
+        await exec(`chmod +x "${executablePath}"`);
```
scrlkx, line 29: What's the point of having `chmod +x` in here? Give execution permission to a file that can be downloaded from the internet can lead to major security issues and as far as I can see we are doing it without even telling the user about it.

As we do not handle game installation we should not handle file permissioning.

scrlkx (changes requested):

src/main/events/library/open-game.ts
gylli251: Like i said in the original PR:

"it uses the lutris runtime thats default to run the executable. sadly it needs to make the .exe execute permissions but don't see a way around that."

Lutris does not run games without making the .exe an executable. dont see a way around it unless we want to make it a popup for the user to select which .exe to make executable?

Or we can look into using some other tool then lutris

zamitto: Any new ideas with this @scrlkx @gylli251?
I fixed the conflict, the PR is now updated with main branch

scrlkx: I've tried this branch but I couldn't make it work. For some reason Lutris is not doing anything. Maybe one of you guys can share a sucefull scenario with version details? Anyway, as I will try to figure it out as soon as I have time to look into ti again.

PR #180, opened 2024-05-03
Fix/onlinefix

pmenta:
changed got-scraping and tough-cookie to fetch-cookie, also created a helper function to decode non utf-8 fetch responses like windows-1251 and removed windows-1251 dep.


hydralauncher (approved):
LGTM

PR #187, opened 2024-05-04
feat: add debrid services caching/downloading support

Netflixyapp:
_I felt like I had to reopen this PR after resolving the conflicts since the electron-builder migration. I've seen there is already another open PR that adds Debrid support but lacks some stuff such as Caching status or handling invalid responses from Real-Debrid._

This PR adds support for Debrid services to cache and download games/torrents

Debrid services are used to generate direct HTTP/S links for torrent files, this makes downloading games much faster since it doesn't rely on seeders anymore. It also improves security in countries or places where torrenting is illegal or traceable.

In this PR, I first added a settings section in the launcher to enable/disable using Debrid services, as well as an input field to specify an API key for Real-Debrid (Eventually, more services can be added soon)
On Python's end, I separated the libtorrent logic into a separate class (TorrentDownloader) and added a new class (DebridDownloader) that handles communication with Debrid services as well as downloading from HTTP/S links using pySmartDL library
A new state was added to the front end, which is DebridCaching. When using Debrid services, you usually get a direct link for a torrent file almost instantly when it's cached in the servers. However, when using a new torrent, Debrid services take a while to download the torrent on the servers to be able to generate a direct link. During this time, DebridCaching status is sent, alongside the progress of this caching process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ out
 
 # IntelliJ
 .idea
+
+# Webpack
+.webpack
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ out
 
 # Sentry Config File
 .env.sentry-build-plugin
+
+# IntelliJ
+.idea
--- a/postinstall.js
+++ b/postinstall.js
@@ -1,0 +1,12 @@
+const fs = require("fs");
+
+if (process.platform === "win32") {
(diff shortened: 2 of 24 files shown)


hydralauncher:
Closing this due to the fact that another implementation has been merged.
Thank you so much for your contribution.

PR #198, opened 2024-05-05
Japanese Locale and Bugs Fix (Locale)

tysaiwofc:
This pull request is to resolve translation bugs and add the translation for the Japanese language, I hope you like it.


zamitto:
It seems your PR is removing some strings that are in the main branch. Could you check it?

tysaiwofc:
I overwrote the files and then the pull request bugged, but everything is practically the same. It only changes the variables that were translated and now it is no longer

PR #211, opened 2024-05-06
Improvements over the real debrid integration.

lilezek:
- Changes the name of http downloader to real debrid, given that the class **only** supports real debrid and no other http source.
- Fixes the notification triggering before the download has been joined from parts into a single file.
- Unarchives the .rar file into the game files to the proper folder.

### DIFF
--- a/src/main/services/downloaders/real-debrid.downloader.ts
+++ b/src/main/services/downloaders/real-debrid.downloader.ts
@@ -3,6 +3,7 @@ import { QueryDeepPartialEntity } from "typeorm/query-builder/QueryPartialEntity
 import path from "node:path";
 import EasyDL from "easydl";
 import { GameStatus } from "@shared";
+import { fullArchive } from "node-7z-archive";
 
 import { Downloader } from "./downloader";
 import { RealDebridClient } from "../real-debrid";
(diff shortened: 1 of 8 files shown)


src/main/services/downloaders/real-debrid.downloader.ts
```diff
@@ -22,52 +26,14 @@ export class HTTPDownloader extends Downloader {
     return 1;
   }
 
-  static async getDownloadUrl(game: Game) {
```
lilezek, line 25: I have moved this method to the real debrid client, given that all it does is to turn a magnet into a downloadable url by purely calling all the real debrid methods.

Netflixyapp:
Very cool implementation 👍 Could we rebase on main to stay up-to-date?
Quick suggestions:
- Handle errors when Real-Debrid fails to download a torrent or transform a magnet to a torrent, there are error codes in the API.
- Handle caching state where Real-Debrid takes a couple of minutes to cache/download a torrent before it's available to download. This could be another state to add to the logic with progress percentage, the API returns that.
- Add a checkbox in settings (**Debrid services** section) where you can enable/disable the use of Debrid because currently, we have to copy/paste the API each time we want to switch between Torrenting/RealDebrid.

hydralauncher (approved):
LGTM

PR #226, opened 2024-05-07
 fix: when .rd doesn't exist, EasyDL downloads the file as if it where the folder.

lilezek:


hydralauncher:
Closing this since there are no changes.

PR #230, opened 2024-05-07
Improved organization and functionality of the hydra-download-manager

shad0wDevv:


thegrannychaseroperation:
Thanks a lot for your contribution, but we will be dropping Python usage in favor of using aria2.

PR #232, opened 2024-05-07
Correction of showOpenDialog without the expected return for a promise

daniielamaro:
Without this change, it is impossible to change the default download folder due to not having the appropriate return for a promise

### DIFF
--- a/src/preload/index.d.ts
+++ b/src/preload/index.d.ts
@@ -99,7 +99,7 @@ contextBridge.exposeInMainWorld("electron", {
   getVersion: () => ipcRenderer.invoke("getVersion"),
   getDefaultDownloadsPath: () => ipcRenderer.invoke("getDefaultDownloadsPath"),
   openExternal: (src: string) => ipcRenderer.invoke("openExternal", src),
-  showOpenDialog: (options: Electron.OpenDialogOptions) =>
-    ipcRenderer.invoke("showOpenDialog", options),
+  showOpenDialog: async (options: Electron.OpenDialogOptions) =>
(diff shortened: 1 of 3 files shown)


src/preload/index.d.ts
zamitto, line 1: the changes on this file are not needed, if you check the return type it i already a Promise

src/preload/index.ts
zamitto, line 1: the changes on this file are not needed, if you check the return type it i already a Promise

src/preload/index.d.ts
daniielamaro: Yes, it is a promise, but the objects that are using this promise are not using the appropriate return for a promise, if you have any doubts, do a test and try changing your download folder in settings

src/preload/index.ts
daniielamaro: Yes, it is a promise, but the objects that are using this promise are not using the appropriate return for a promise, if you have any doubts, do a test and try changing your download folder in settings

src/preload/index.d.ts
zamitto: yes, but the changes on the index.d.ts and index.ts files are not needed. The changes on show-open-dialog.ts are the changes that fix the problem. 
I just tested it

daniielamaro: Você está certo, enviei mais mudanças do que o necessario, já corrigi e deixei so o necessário para corrigir o problema

daniielamaro: You're right, I sent more changes than necessary, I already corrected them and left only what was necessary to fix the problem

zamitto:
This is fixed now on main branch. Thank you for your collaboration anyways

PR #249, opened 2024-05-09
Feature change language & Fullscreen options in settings

darkjontex:
Including options in settings to allow:

- Open Hydra Launcher in Fullscreen mode or custom resolution.
- Change the Language system for a custom one.

Everything is saved as userPreferences, so if you clear your Hydra Launcher Cache, everything is reset.

I hope you guys can implement or use part of the code.
Once again, thanks for the project Hydra Launcher.

### DIFF
--- a/.hintrc
+++ b/.hintrc
@@ -1,0 +1,8 @@
+{
+  "extends": [
+    "development"
+  ],
+  "hints": {
+    "axe/forms": "off"
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,10671 @@
+{
+  "name": "hydra",
+  "version": "1.1.1",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "hydra",
+      "version": "1.1.1",
+      "hasInstallScript": true,
+      "dependencies": {
+        "@electron-toolkit/preload": "^3.0.0",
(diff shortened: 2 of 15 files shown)


darkjontex:
Someone from Los Broxas quem review my PR? I would like to contribute with those features for Hydra Launcher.

darkjontex:
- Fixed points about Style and Select component for Language.
- Removed Input Fields for Resolution.

image

image

zamitto:
Ok, i took a better look into your PR. There are some problems:
1. A lot of comments
2. Your PR aims to do 2 things (persist screen resolution and change language)
3. None of these things seems done
4. Your options inside the Select are hard coded. If you want to work on this feature, you need to figure out a way to dinamically load your options so we dont have to revisit this component every time a new localization is added

I will have to close this PR, and I suggest you do only one thing per PR.

piradata:
does not seems so hard to make it dynamic, its vasically do the sabe thaat the localization plugin does, reading the folders at the localization directory. u can use the same var it uses and convert the local code back to language name

piradata:
does not seems so hard to make it dynamic, its vasically do the sabe thaat the localization plugin does, reading the folders at the localization directory. u can use the same var it uses and convert the local code back to language name

piradata:
will take a look at the language changer later this week if no one does, seems funny

darkjontex:
ok

PR #255, opened 2024-05-09
Standardizing the code and adding a feature to enhance performance

cardosource:
Using the Singleton pattern will ensure that only one instance of the Session class is created. Replacing threading with asyncio now makes it asynchronous and will execute multiple asynchronous tasks on a single thread and WILL NOT have a large memory consumption.


thegrannychaseroperation:
Thanks a lot for this contribution, but we will be migrating to aria2 implementation instead of keeping libtorrent.

PR #267, opened 2024-05-10
add belarusian translation

Latgardi:


zamitto:
I don't speak Belarusian but I check some string on Google Translate and it seems fine

Issue #225, opened 2024-05-07
Create a more friendly experience when presenting repack options


hydralauncher:
A few things could be simply inferred by the repacker and the title of the repack:
- Whether it has multiplayer or not (every OnlineFix game; repacks with "Multiplayer" in the name)
- Whether it has support for multiple languages (MULTiXX)
- The language of the installer (always Russian for Xatab, for instance)
- Show quantity of seeds/peers using something like https://checker.openwebtorrent.com/ (this requires the magnet to have trackers)

ChristoferMendes:
I would like to contribute on this one. Can you add me as an assignee?

ChristoferMendes:
WDYT?

![seeders-peers](https://github.com/hydralauncher/hydra/assets/107426464/0b8eacab-5cb4-469c-a502-1689a216fecc)

ChristoferMendes:
Or like this instead...

![image](https://github.com/hydralauncher/hydra/assets/107426464/8386b74f-613f-49d0-88f0-45ca948ef73a)

xbozo:
> Or like this, instead...
> 
[...]

PR #270, opened 2024-05-10
More friendly experience when presenting repack options

ChristoferMendes:
## CHANGES MADE

- Add a new file for getting info from title (multiplayer/multixx)
- Add new components (Tag, Tooltip)
- Add fetch of peers/seeders (with 2 minutes of cache)
- Add language of installer ('ru' for Xatab, 'en' for others)

<details>
<summary>UI</summary>

repacks-modal
</details>

#225


JackEnx:
run "yarn lint", there some errors to fix.

image

src/main/services/open-web-torrent.ts
```diff
@@ -0,0 +1,50 @@
...
+      const response = await axiosRef.get(endpoint, { params });
+      const { seeds, peers } = this.transformResponseError(response);
+
+      return {
+        seeders: seeds,
+        peers: peers,
+      };
+    } catch (e) {
+      const error = e as Error;
+      throw new Error(error.message);
```
JackEnx, line 47: doesn't this have the same effect as not using try/catch?

src/main/services/repack-tracker/helpers.ts
```diff
@@ -5,7 +5,7 @@ import { repackRepository } from "@main/repository";
 
 import { QueryDeepPartialEntity } from "typeorm/query-builder/QueryPartialEntity";
 
-export const savePage = async (repacks: QueryDeepPartialEntity<Repack>[]) =>
+export const saveRepacks = async (repacks: QueryDeepPartialEntity<Repack>[]) =>
```
JackEnx, line 8: is there any specific reason for changing the name of function in this pr?

ChristoferMendes: Yeah, sorry, I didn't specify this. But my goal with this change is to improve readability, as the function does not save any page, but instead saves the repacks on the database. While I was reading the code I found this suitable. I will enhance the description to contain this

src/main/services/open-web-torrent.ts
ChristoferMendes: Yes, correct. Updated

ChristoferMendes:
@JackEnx, could you revisit this? Updated the code with requested changes

src/main/services/repack-tracker/helpers.ts
JackEnx: It does indeed save the page. The reason behind is that we keep the page column in order to track and resume from the page that the scraper left at if it crashes or simply cannot move any forward.

For now I think it's best to leave the name as it is.

image

package.json
```diff
@@ -53,6 +53,7 @@
     "jsdom": "^24.0.0",
     "lodash-es": "^4.17.21",
     "lottie-react": "^2.4.0",
+    "lucide-react": "^0.378.0",
```
hydralauncher, line 56: Do you really need a whole library for two icons? Can't you use the svgs individiually instead?

src/locales/en/translation.json
```diff
@@ -96,7 +96,20 @@
     "dont_show_it_again": "Don't show it again",
     "copy_to_clipboard": "Copy",
     "copied_to_clipboard": "Copied",
-    "got_it": "Got it"
+    "got_it": "Got it",
+    "repack_language_code": {
```
hydralauncher, line 100: I don't think this is really necessary. Plase check this out:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames
(This applies for all the translations)

src/main/events/index.ts
```diff
@@ -27,6 +27,7 @@ import "./torrenting/start-game-download";
 import "./user-preferences/get-user-preferences";
 import "./user-preferences/update-user-preferences";
 import "./user-preferences/auto-launch";
+import "./repacks/magnet-data"
```
hydralauncher, line 30: Please consider this as being part of the catalogue instead

src/main/services/open-web-torrent.ts
```diff
@@ -0,0 +1,45 @@
...
+      error: {
+        code: number;
+        message: string;
+      };
+    };
+
+const httpsAgent = new Agent({
+  rejectUnauthorized: false,
+});
+const axiosRef = axios.create({
```
hydralauncher, line 19: You don't need a whole axios instance for a single request;
Plus, you should use https://www.npmjs.com/package/webtorrent-health instead (it's the same library used by the checker).

src/main/services/repack-tracker/1337x.ts
```diff
@@ -3,7 +3,7 @@ import { JSDOM } from "jsdom";
 import { formatUploadDate } from "@main/helpers";
 
 import { Repack } from "@main/entity";
-import { requestWebPage, savePage } from "./helpers";
+import { requestWebPage, saveRepacks } from "./helpers";
```
hydralauncher, line 6: Please keep `savePage` instead

src/preload/index.ts
```diff
@@ -112,4 +112,5 @@ contextBridge.exposeInMainWorld("electron", {
   showOpenDialog: (options: Electron.OpenDialogOptions) =>
     ipcRenderer.invoke("showOpenDialog", options),
   platform: process.platform,
+  getMagnetData: (magnet: string) => ipcRenderer.invoke("getMagnetData", magnet),
```
hydralauncher, line 115: I think it's worth changing the event name to `getMagnetHealth` or something

src/renderer/src/components/tag/tag.css.ts
```diff
@@ -0,0 +1,11 @@
+import { style } from "@vanilla-extract/css";
+
+export const tagStyle = style({
```
hydralauncher, line 3: No formatting on this file

src/renderer/src/components/tag/tag.css.ts
```diff
@@ -0,0 +1,11 @@
...
+  borderRadius: '3px',
+  border: '1px solid #FFFFFF33',
```
hydralauncher, line 5: Use the border color from the theme vars please

src/renderer/src/components/tag/tag.css.ts
```diff
@@ -0,0 +1,11 @@
...
+  padding: '2px 5px',
```
hydralauncher, line 6: Don't use arbitrary spacing, use `SPACING_UNIT` from the vars instead

src/renderer/src/pages/game-details/seeders-and-peers/useMagnetData.tsx
```diff
@@ -0,0 +1,78 @@
+import { useEffect, useState } from "react";
```
hydralauncher, line 1: Do you really think that it's necessary to create a hook that is going to be used in a single place?

hydralauncher (changes requested):

src/main/services/open-web-torrent.ts
ChristoferMendes: I didn't know about that, thanks!

src/renderer/src/pages/game-details/seeders-and-peers/useMagnetData.tsx
ChristoferMendes: I prefer to organize components with custom hooks and utilities to keep them focused on rendering, making the code more modular and maintainable. What do you think?

src/main/services/repack-tracker/helpers.ts
ChristoferMendes: Thank you for the information, reverting

package.json
ChristoferMendes: Done

src/locales/en/translation.json
ChristoferMendes: Perfect, thanks. Done

zamitto:
@ChristoferMendes , just a heads up, there are lint errors to fix as well

zamitto:
Repacks from onlinefix should have the multiplayer tag. Could you check this feature? It does not seem to be working here

image

ChristoferMendes:
> Repacks from onlinefix should have the multiplayer tag. Could you check this feature? It does not seem to be working here
> 
> image

Will check

ChristoferMendes:
> Repacks from onlinefix should have the multiplayer tag. Could you check this feature? It does not seem to be working here
> 
> image

Fixed! I was not checking for the repacker itself, only title

zamitto:
You last thing I think.
Are all magnets returning 0 seeders and 0 peers for you as well?
image

ChristoferMendes:
> You last thing I think.
> Are all magnets returning 0 seeders and 0 peers for you as well?
> image
> 

Strange, for me it was okay. I'll check it out

zamitto:
> > You last thing I think.
> > Are all magnets returning 0 seeders and 0 peers for you as well?
> > image
> 
> Strange, for me it was okay. I'll check it out

Yeah, the log says timeout for all trackers. It's strange.

JackEnx:
earlier I was checking this timeout issue, one way I did it was to check when empty, then i triggering a second request with a longer time, below is an example, its just a example

image

ChristoferMendes:
> earlier I was checking this timeout issue, one way I did it was to check when empty, then i triggering a second request with a longer time, below is an example, its just a example
> 
> image
> 
> first request, 1 second
> 
> image
> 
> second request, 10 seconds after the first request
> 
> image

Applied retry logic in here: https://github.com/hydralauncher/hydra/pull/270/commits/7a2b693397ef051ec50aee0a8428d824b11a567e

src/main/services/web-torrent-data.ts
```diff
@@ -0,0 +1,32 @@
...
+            return resolve(
+              webTorrentData.getSeedersAndPeers(magnet, retry + 1, timeout * 2)
+            );
+          }
+
+          return resolve({ peers, seeders: seeds });
+        }
+      );
+    });
+  },
```
zamitto, line 31: i don't like the recursive function

could you replace with a loop solution?
maybe something like this (just an example):

```
  async getSeedersAndPeers(magnet: string) {
    let peers = 0;
    let seeds = 0;
    let retry = 0;

    while (retry < 3 && (!peers || !seeds)) {
      try {
        const data = await webTorrentHealth(magnet, {
          timeout: 1500 * (retry + 1),
        });

        peers = data.peers;
        seeds = data.seeds;
      } catch (e) {
        retry++;
      }
    }

    return { peers, seeders: seeds };
  },
```

zamitto (changes requested):

src/main/services/web-torrent-data.ts
ChristoferMendes: Done

ChristoferMendes:
> will try to finish reviewing your pr todat, we are kinda busy this week

Thank you!

zamitto (approved):

zamitto:
@ChristoferMendes, I had merged your PR earlier, but we will actually  test it a little more before an actual merge

zamitto:
If you could recreate the PR that would be nice

ChristoferMendes:
> If you could recreate the PR that would be nice

Recreate it? Without changes?

PR #276, opened 2024-05-11
better ru localization

delyfss4796:


delyfss4796:
close cuz https://github.com/hydralauncher/hydra/pull/278 impove it

PR #287, opened 2024-05-11
fix(window): windows beign placed on the wrong display

scrlkx:
This solves the issue with placing new windows. When running on GNOME/Wayland it happens of placing the windows on the wrong display as it doesn't consider the primary display position.

```typescript
      x: primaryDisplay.bounds.x,
      y: primaryDisplay.bounds.y,
```

### DIFF
--- a/src/main/services/window-manager.ts
+++ b/src/main/services/window-manager.ts
@@ -1,10 +1,13 @@
-import { BrowserWindow, Menu, Tray, app } from "electron";
+import { BrowserWindow, Menu, Tray, app, screen } from "electron";
 import { is } from "@electron-toolkit/utils";
 import { t } from "i18next";
 import path from "node:path";
 import icon from "@resources/icon.png?asset";
 import trayIcon from "@resources/tray-icon.png?asset";
 import { userPreferencesRepository } from "@main/repository";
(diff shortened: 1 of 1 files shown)


src/main/services/window-manager.ts
```diff
@@ -27,12 +30,16 @@ export class WindowManager {
...
     this.mainWindow = new BrowserWindow({
-      width: 1200,
-      height: 720,
-      minWidth: 1024,
-      minHeight: 540,
+      height: height,
+      width: width,
+      minHeight: WINDOW_MIN_HEIGHT,
+      minWidth: WINDOW_MIN_WIDTH,
+      x: primaryDisplay.bounds.x,
```
scrlkx, line 41: This is the relevant part: making sure the new window will be placed at the right (x, y) based on the bounds of the primary display.

chubbygrannychaser:
It doesn't actually start maximized, but rather filling the screen; in order to maximize, you would have to simply call this Electron event:
https://www.electronjs.org/docs/latest/api/browser-window#winmaximize
I don't think it makes sense to force Hydra to go fullscreen when opening it. Closing this for now.

PR #304, opened 2024-05-12
fix: unable to install games on linux

### LINKED ISSUE #299: [BUG] Installers are not handled correctly on Linux (using Lutris or Wine)
**Describe the bug**
We already have the ability to handle running game installers using Lutris or Wine as you can see [here](https://github.com/hydralauncher/hydra/blob/main/src/main/events/library/open-game-installer.ts#L41) and [here](https://github.com/hydralauncher/hydra/blob/main/src/main/events/library/open-game-installer.ts#L48).

However, there is something wrong on the lines above it. We have an extra logic that appends a `setup.exe` to the file path which results in an invalid path to the game file. And, when this path is not found, we just call `shell.openPath(gamePath)` and, as Linux doesn't know how to open .exe files, it just does it's best opening with Archive Manager or something like that.

That means the lines that handle opening the installer with Lutris or Wine never reach. Removing this piece of code seems to fix the issue, but I assume they are there for a reason?

```typescript
const setupPath = path.join(gamePath, "setup.exe");
if (!fs.existsSync(setupPath)) {
    shell.openPath(gamePath);
    return true;
}
```

On the lines above, `gamePath` is a valid path (pointing to the installer file), but `setupPath` it's not. `fs.existsSync(setupPath)` will always be `false` so we will always do `shell.openPath(gamePath);`.

**To Reproduce**

1. Open Hydra on a Linux distribution
2. Download any game
3. After download is finish, click on Install

Note that, what will happend after you click install depends on which distribution/desktop environment you are running. In case of GNOME it should open Archive Manager.

**Expected behavior**
In case Lutris or Wine are installed, the game installer should open on them. I'm not so sure about what should happen if they are not.

**Screenshots**

**Additional context**

- OS: Debian 12/GNOME
- Hydra Version: v1.1.1
- Additional information and context of your problem:

scrlkx:
There is a issue when installing games on Linux. I described it in #299.

Fixes #299


zamitto:
Finally got able to install a linux disto on my PC, will test your PR today

zamitto:
Gonna post here the same commend I posted on #299:


May I suggest something?

1. If there is a setup.exe inside the gamePath, we call wine ${setupPath} and return
2. If there is not a setup.exe but there is any other .exe file, we call wine ${gamePath}
3. If not, we open the directory using shell.openPath(gamePath)

The Windows conditional remains the same as it is in your PR. These checks should happen after that.

zamitto:
Hey @scrlkx any changes you can work on those changes today? We'd like to merge this PR today to start the tests for the next release
If you are busy or anything, no problem, I probably can commit the changes we talked about on #299

scrlkx:
I have implemented the changes you suggested with one single consideration: the first step of the installation process is now to check if the game path actually leads to a file (not a directory). If it does, we proceed to execute it, assuming it is the game installer. This is important for cases like [18 Wheels of Steel: Across America](https://store.steampowered.com/app/514160/18_Wheels_of_Steel_Across_America/), where the game comes as a single file, and it also makes all other handling more reasonable.

Additionally, I created a function to execute files based on the operating system. This way, we avoid duplicating the Wine logic.

I tested these changes using different games that come in various formats, including:

    Rabbit and Stell - which comes in a single compressed file.
    18 Wheels of Steel: Across America - which comes as a single executable file.
    A Little to the Left - which comes in a folder with a setup.exe file.
    BIT.TRIP Rerunner - which comes in a folder with a setup.exe file that I renamed to installer.exe to stage the scenario where there is an installer with a different name.

Everything seems to be working well.

It's important to note that the Lutris logic has become less useful as we now employ a Wine-first approach. However, this doesn't seem like a problem since the current Lutris implementation relies on having a setup.exe file available, which seems out of place. I do think we can make it more useful based on the insights we brought up in #505.

.gitignore
zamitto, line 1: Since #521 the logs folder won´t be here, so this shouldn't be needed

zamitto:
The only thing I'm worried is if there are any game that can have more than one .`exe` file inside de download folder. I've never seen an example like that, but i don't know...

scrlkx:
It's a fair concern. We could improve what we have by looking for something similar to "setup.exe", "{gamename}.exe", you name it. There is no silver bullet though. I do think this PR improve some cases, I had one with NFS (I think) where it comes with a "nfs-something.exe" and it was well handled, so...

scrlkx:
Ideally we should know what to run, but as it's not possible right now we will have to do some assumptions.

src/main/events/library/open-game-installer.ts
```diff
@@ -29,15 +44,22 @@ const openGameInstaller = async (
...
-  if (process.platform === "win32") {
-    shell.openPath(setupPath);
-    return true;
+  const gamePathFileNames = fs.readdirSync(gamePath);
+  const gameAlternativeSetupPath = gamePathFileNames.find(
+    (fileName: string) => path.extname(fileName).toLowerCase() === ".exe"
+  );
+
+  if (gameAlternativeSetupPath) {
+    return executeGameInstaller(path.join(gamePath, gameAlternativeSetupPath));
```
zamitto, line 62: What do you think about this? If for some reason there is more than 1 .exe the code won't select an arbitrary one.
The rest of the login remains the same


```suggestion
  const gameAlternativeSetupPath = gamePathFileNames.filter(
    (fileName: string) => path.extname(fileName).toLowerCase() === ".exe"
  );

  if (gameAlternativeSetupPath.length == 1) {
    return executeGameInstaller(path.join(gamePath, gameAlternativeSetupPath[0]));
```

scrlkx: Yeah, we can do this.

zamitto (approved):

datingesp-hash:
im using arch linux with hyprland and I cant use this launcher, when I choose a .exe and a wine prefix and click play, it opens firefox asking to save or open the .exe file, whatever u click it does nothing

datingesp-hash:
im gettting this error 05:46:02.756                › Unhandled Error: spawn /home/archie/Games/silksong/Hollow Knight Silksong/Hollow Knight Silksong.exe EACCES
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
    
    I spent hours but I cant make it work

s0mebodyhelpme:
Exact same situation here except on Cachyos. Can't launch anything with this.

PR #324, opened 2024-05-12
Refactor code to encapsulate global variables and simplify structure

BalajiJBcs:
- moved downloading_game_id and torrent_handle variables into the listen_to_socket function scope to avoid global usage
- few structure and function changes


thegrannychaseroperation:
Thanks a lot for your contribution, but we will be dropping Python usage in favor of using aria2 instead.

PR #296, opened 2024-05-11
Readme modifications for Discord change to Telegram

piradata:
Hello guys, my first PR here, so basically the changes are:

- Make the list of contributors in README dynamic using contrib.rocks
- Add pt-bBR README
- Add uk-AU README
- Add uk-AU localization
- Add README links to other languages
- Remove discord tag from READMEs
- Change discord link to telegram on README
- Change link and logo on launcher sidebar from discord to telegram

Here is how it will be:

image

image

image

image

PR #328, opened 2024-05-12
Add Ukrainian localization.

DenisCooper09:


zamitto:
Ukranian localization has already been added in #296 
@piradata and @squidsdev, could you double check if it should be "uk"or "ua" characters for the localization?

piradata:
its `uk` according to https://en.m.wikipedia.org/wiki/IETF_language_tag

piradata:
i checked the lib we use and it has this link as a reference @zamitto

zamitto:
> i checked the lib we use and it has this link as a reference @zamitto 

Thanks! I will leave the library doc page that references the link you said in the previous comment here as well:

https://www.i18next.com/principles/translation-resolution

PR #348, opened 2024-05-13
Belarusian language translation

ruth3n:


zamitto:
I see #347 is your PR as well. Could you update them so they have only the required modifications of each PR? 
Like, this one must have only the modifications related to the belarusian localization

zamitto (approved):

PR #352, opened 2024-05-13
Using GitHub forms for Issues

ruth3n:
GitHub forms can provide a more structured way to gather information when issues are created.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: screenshots
+    attributes:
+      label: Screenshots
+      description: If applicable, add screenshots to help explain your problem.
+    validations:
+      required: false
+  - type: input
```
zamitto, line 38: could you update it so this type is a dropdown with the SO options?

ryeviz (commented):
The form is well-structured, covering all necessary elements

.github/ISSUE_TEMPLATE/bug_report.yml
ruth3n: Will Windows and Linux be sufficient or do we need to specify the version/distribution?

piradata: windows, linux and mac, and maube a optional field below so the user can say say the version (win7, 10, 11, or arch, on manjaro, or ubuntu, etc)

zamitto: Yeah, when I commented I forgot about the version. I think the way you did is good for now

zamitto (approved):

PR #361, opened 2024-05-14
feat: search results pagination

### LINKED ISSUE #292: [BUG] Problems with searching
Some games are very difficult to find in the launcher. For example, I want to download the RAGE game, which is not in the list because I can not turn the page.
image

- OS: [Windows 11]
- Hydra Version: 1.1.0

### LINKED ISSUE #357: [REQUEST] A button to "show more" when searching for games
For example, when I searched for *Raft*, I found a bunch of *aircraft* games and not what I wanted, there are two solutions to this problem, either improve the app's search method or add a button to "show more" games.

Captura de tela 2024-05-13 184018
Captura de tela 2024-05-13 184029

mohannadzidan:
This PR will add pagination buttons to search results to show more search results 

closes #357, and closes #292


zamitto:
Please, always add a screenshot when your PR is releated to UI stuff.

mohannadzidan:
image

piradata:
could you please add the translation for `en` at least? the others can be added later

piradata:
image
@mohannadzidan u can use it actually

mohannadzidan:
I have updated the code to use `next_page`, `previous_page` keys from the catalogue namespace

image

mohannadzidan:
@zamitto hey, this feature is ready to merge, can you please review the code?

src/renderer/src/pages/home/search-results.tsx
```diff
@@ -50,10 +55,40 @@ export function SearchResults() {
...
     <SkeletonTheme baseColor={vars.color.background} highlightColor="#444">
+      <section
+        style={{
+          padding: `${SPACING_UNIT * 3}px ${SPACING_UNIT * 4}px`,
+          display: "flex",
+          width: "100%",
+          justifyContent: "space-between",
+          alignItems: "center",
+          borderBottom: `1px solid ${vars.color.border}`,
+        }}
```
zamitto, line 76: you could move this style object to home.css.ts file

src/renderer/src/pages/home/search-results.tsx
```diff
@@ -50,10 +55,40 @@ export function SearchResults() {
...
+          disabled={page === 0 || isLoading}
+        >
+          <ArrowLeftIcon />
+          {t("previous_page", { ns: "catalogue" })}
+        </Button>
+
+        <Button onClick={handleNextPage} theme="outline" disabled={isLoading}>
+          {t("next_page", { ns: "catalogue" })}
+          <ArrowRightIcon />
+        </Button>
```
zamitto, line 90: this button will never get disabled? even if the page has only one result?

piradata:
hey @mohannadzidan , need some help here?

mohannadzidan:
Hello @piradata and @zamitto

 I've implemented the changes requested in the previous review:

1. Moved pagination inlined styles to home.css for better separation of concerns.
2.  Updated the searchGames function signature for improved clarity and reusability.
3. Added functionality to disable the "next" button when there are no more results to prevent unexpected behavior.

Please take another look at the code and let me know if everything looks good or if you have any further suggestions. Thanks!

src/renderer/src/pages/home/search-results.tsx
```diff
@@ -16,6 +20,8 @@ import { useNavigate, useSearchParams } from "react-router-dom";
 import * as styles from "./home.css";
 import { buildGameDetailsPath } from "@renderer/helpers";
 
+const PAGE_SIZE = 12;
```
piradata, line 23: why set the pagesize here if you already set the default on the function itself to 12?

src/main/events/catalogue/search-games.ts
```diff
@@ -1,12 +1,19 @@
 import { registerEvent } from "../register-event";
 import { searchGames } from "../helpers/search-games";
-import { CatalogueEntry } from "@types";
+import { CatalogueEntry, PaginationArgs } from "@types";
+import { defaults } from "lodash-es";
+
+const DEFAULT_SEARCH_ARGS: PaginationArgs = {
+  take: 12,
+  skip: 0,
```
piradata, line 8: instead of skip, you could also just pass the page as a default for zero, and calculate the skip inside the function itself, then you dont need to know the default value for take when you just wannt to go to the page like 5 using the default "take" parameter

src/renderer/src/pages/home/search-results.tsx
mohannadzidan: I made it like this to increase the code readability and avoid hard coding magic numbers in the code while calculating whether this is the last results page or not

For example if a new contributor saw this line
```ts
setIsEnd(results.length < 12)
```
He will be confused and wonder what does this number  12  mean and why we put it here, right? But by declaring this constant and use it properly in the query it is clearer and will make since to him

piradata: but you can fet this number from the search params the same way you does to get the page no?

src/main/events/catalogue/search-games.ts
mohannadzidan: Actually your suggestion was exactly the previous version of the function when i modified it initially for this pagination feature, but I thought that it is more reusable like this? Like if in the future we had to use this function to display different number of results we will able to do this without revising the code again and modify it

piradata: +/-, like, how it is now, you need to know the `take` parameter to send the `skip` correctly considering the number of the `page` you want, so you basically dont make use of the default value for `take`, as you need to consider it anyway to send the `skip`.

if you send only `take` and `page`, both optional to 12 an 0 respectively, you dont need to care about any of this logic, and can even just send the `page` arg for any request. and you also has access to the `take` (default or not) used on the query to see if you are on the last page (query could also return a parameter with the number of pages tho)

what do you think @zamitto?

henriqueleite42: I kinda agree with @piradata on this one, but the default to 12 on `take` I think that should be a global default value, indeed of having one here, another one in `src/renderer/src/pages/home/search-results.tsx` and so on.

piradata:
@zamitto what do you think on this one?

piradata:
@zamitto is this one stil relevant?

zamitto:
> @zamitto is this one stil relevant?
Not anymore, we moved the search to our api
We might eventually add pagination to it
But, as now the results are much more relevant and gives what the user is search for, pagination is not a top priority right now

PR #373, opened 2024-05-14
Adding recently played games to tray context menu

mag1yar:
Before starting the game:
image
After:
image

---

The only thing I ask you to do is test downloading any game, because for some reason I get an error:
```
TypeError: Cannot read properties of null (reading 'replace')
```
I think this is my local problem. Otherwise everything seems to be working fine.


zamitto:
Cool! Could you add a screenshot of your feature?

mag1yar:
added
> Cool! Could you add a screenshot of your feature?

ChristoferMendes:
@mag1yar 

Such a nice feature! I will test this TypeError locally and see if I can find the error

src/main/services/window-manager.ts
```diff
@@ -77,33 +86,66 @@ export class WindowManager {
...
+        },
+      });
+
+      const recentlyPlayedGames: Array<MenuItemConstructorOptions | MenuItem> =
+        games.map(({ title, id, executablePath }) => ({
+          label: title,
+          type: "normal",
+          click: async () => {
+            if (!executablePath) return;
+            await gameRepository.update({ id }, { executablePath });
```
zamitto, line 107: just a question, why is it needed to update the executablePath on click?

mag1yar: The same thing happens here as in the sidebar when we select a game in the library.
Without the update, we will not know which game the user launched last time and the game will not be first in the list.

zamitto: Ok, code looks fine, I will run and test it today. Really cool feature, thanks!

zamitto: This update was still in my mind so I investigated and this update is not needed right here. The process-watcher is the responsible to update the lastTimePlayed. 

We update the executablePath in the openGame function because there may be a case where the openGame receives the executablePath as an argument.

For your feature, this isn't really needed. I just tested and your feature works correctly removing this call. Could you just remove it? Then all should be fine to merge!

mag1yar: Yes, you are right, I removed update, but in addition I discovered that lastTimePlayed is a utc string and it was not sorted correctly, I had to sort by updateAt, but I don’t know if it’s correct to sort by it.

zamitto: Yes, I think your idea is good for now. We will probably work on a migration and update so the DB saves the lastTimePlayed as timestamp. This will solve some issues that will happen sorting by updateAt too

zamitto (approved):

PR #387, opened 2024-05-15
build(linux): add flatpak/rpm package and remove snap

fhilipecrash:
- Remove snap package from build because flatpak and appimage are more used
- Add mimeType to handle `hydralauncher://` protocol


zamitto:
I think you would also need to update the build.yml file

image

zamitto:
Closing as done in #487
I took your branch, rebased and added the commented change

zamitto:
I actually removed the flatpak from my PR as the build was failing, so it is still open to do

iMonZ:
> I actually removed the flatpak from my PR as the build was failing, so it is still open to do

Hey could you then reopen this PR? That would be nice

Issue #365, opened 2024-05-14
[REQUEST] Russification of the launcher and website


ghost:
Hello, I suggest adding Russian and Ukrainian languages. Both on the website and in the launcher.
Why?
Because I believe that in such regions your resource will be in demand.

zamitto:
It's already added in the launcher on the lastest release

aleks1timiya:
![image](https://github.com/hydralauncher/hydra/assets/149929281/52eaf3ac-6a07-4654-bf29-5754e9a79564)

zamitto:
You are not running the latest release

aleks1timiya:
How can I update?

zamitto:
https://github.com/hydralauncher/hydra/releases

KMozor:
Hmm, I today installed on Windows the latest version and Russian language I did not observe there
Can somehow through the config to fix it?

zamitto:
Is your OS configure with russian language? Hydra should get the default language of your OS and use it.
Anyway, the next update should have a language selector

KMozor:
Yes, my OS is completely in Russian, anyway, I'll wait for the language selection item

aleks1timiya:
Да , русификация отличная, но сайт на португальском и английском.

piradata:
strange, if the OS is in ru the language should have been detected automatically here:

![image](https://github.com/hydralauncher/hydra/assets/16529503/4b0a8ff5-61e2-4323-947a-9fffa4e03fca)

where `@locales` are

![image](https://github.com/hydralauncher/hydra/assets/16529503/00667a72-d305-4649-ba56-e42bc5994ea9)

piradata:
addedlanguage selector menu to solve this

PR #402, opened 2024-05-16
Add option to choose locale from settings menu

piradata:
Someone can help translating the 'language' for the other locales? like, how do i write "language" in ru or be?

Thanks @darkjontex for the initial work in #249 💯 

There is how it will be:

image

image


piradata:
close #371 and possibly some others

userMacieG:
pl: `"language": "Język"`

src/renderer/src/pages/settings/settings-general.tsx
```diff
@@ -79,28 +102,40 @@ export function SettingsGeneral({
         </Button>
       </div>
 
+      <h3>{t("language")}</h3>
+      <>
+        <Select value={form.language} onChange={handleLanguageChange}>
+          {Object.keys(languageResources).map((language) => (
+            <option key={language} value={language}>
+              {ISO6391.getName(language)}
```
zamitto, line 110: Could you try with getNativeName instead of getName?

I think it would be better to show the languages names ready in their native language

zamitto (approved):

zamitto (commented):
There are probably some tweaks to do but i will take a better look later

package-lock.json
zamitto, line 1: You probable used npm, we use yarn as dependency manager tool.
Could you delete this file?

src/renderer/src/pages/settings/settings-general.tsx
```diff
@@ -15,34 +18,41 @@ export function SettingsGeneral({
...
+  const { t } = useTranslation("settings");
+
   const [form, setForm] = useState({
     downloadsPath: "",
     downloadNotificationsEnabled: false,
     repackUpdatesNotificationsEnabled: false,
+    language: "",
   });
 
+  const [defaultDownloadsPath, setdefaultDownloadsPath] = useState("");
```
zamitto, line 30: rename `setdefaultDownloadsPath` to `setDefaultDownloadsPath` to follow camelCase pattern

src/renderer/src/pages/settings/settings.tsx
```diff
@@ -27,32 +27,31 @@ export function Settings() {
...
+          />
+        );
+      case "real_debrid":
+        return (
+          <SettingsRealDebrid
+            userPreferences={userPreferences}
+            updateUserPreferences={handleUpdateUserPreferences}
+          />
+        );
+      default:
```
zamitto, line 46: If you want to refactor to a switch case (which makes sense to me) I think it would be better to use the "behaviour" string here instead of default

package-lock.json
piradata: ha, yes, i use mostly npm so forgot about yarn. is there any way to warn about it or just run yarn install if someone accidentally uses npm install?

src/renderer/src/pages/settings/settings.tsx
piradata: Can I return the "return" of the case statement btw without having to write return for every case? Tried to do that but does not work, i am more familiar with ruby that has explicit return

piradata:
nice, will do these revisions later tomorrow

piradata:
> pl: `"language": "Język"`

thanks, will add

package-lock.json
zamitto: yes, i just approved a PR that force the use of yarn, so this should be fine from now on

src/renderer/src/pages/settings/settings.tsx
zamitto: AFAIK, in javasript you have to return on each case

src/renderer/src/pages/settings/settings-general.tsx
```diff
@@ -15,34 +18,41 @@ export function SettingsGeneral({
...
+
+  const handleLanguageChange = (event) => {
+    const value = event.target.value;
+
+    handleChange({ language: value });
+    changeLanguage(value);
+  };
 
   const handleChange = (values: Partial<typeof form>) => {
+    // TODO: why is the setForm needed if updateUserPreferences already changes the
```
zamitto, line 53: later today i will test this

mohannadzidan:
Awesome work, you can find all languages with their codes and native name in this json file

[languages.json](https://github.com/hydralauncher/hydra/files/15338709/languages.json)

piradata:
> Awesome work, you can find all languages with their codes and native name in this json file
> 
> [languages.json](https://github.com/hydralauncher/hydra/files/15338709/languages.json)
> 
> 

thanks @mohannadzidan , but the `ISO6391.getNativeName` already does that automagically :) will just change the method later

piradata:
Done guys, ready to merge :)

piradata:
image

piradata:
also /closes #365

src/renderer/src/pages/settings/settings-general.tsx
```diff
@@ -15,34 +18,41 @@ export function SettingsGeneral({
...
```
zamitto, line 53: Could you just remove these comments?

src/renderer/src/pages/settings/settings-general.tsx
```diff
@@ -79,28 +102,40 @@ export function SettingsGeneral({
         </Button>
       </div>
 
+      <h3>{t("language")}</h3>
+      <>
+        <Select value={form.language} onChange={handleLanguageChange}>
+          {Object.keys(languageResources).map((language) => (
```
zamitto, line 108: Could you move this to a useState? This way you can also sort the languageOptions. Something like this:

image

zamitto: Another thing that I noticed it that, the current string for portuguese language is "pt-BR", so the Select component won't have the correct language selected by default for who has portuguese language. I'm not sure the best way to fix this. Maybe something using the startsWith function. If you have an ideia let us know

zamitto (changes requested):

src/renderer/src/pages/settings/settings-general.tsx
piradata: is there already an issue about this? if not, i can create one and remove the comment

src/renderer/src/pages/settings/settings-general.tsx
piradata: where did you see it is `pt-br`? could you send a pint of the file that sets this or file name and line? because where i looked also used the `import * as languageResources from "@locales";`, and there is `pt`

piradata: > Could you move this to a useState? This way you can also sort the languageOptions. Something like this:
> 
> image
> 

makes sense to sort if the languages wil be shown as native name tho? the alfaphabets are not the same and it is not guaranteed that the lang codes sorted, will keep the native names sorted. i can sort after calling the toNativeName tho

piradata: one thing that i was thinking is, if there are a LOT of languages available, the select will be massive, and i have not tested a scroll there. one option is a maxheight and scroll, another option is to use smt like a select2 with a max shown opts

src/renderer/src/pages/settings/settings-general.tsx
zamitto: There's not, you may create one

src/renderer/src/pages/settings/settings-general.tsx
zamitto: I think a scroll should be fine, could you try it?
And yes, the sort should happen after you get the native names

piradata: the thing is, i will have ti sort by native name and covert back tolang code... will i think there is a good way to do this, will try here. will also try the scroll

piradata: > Another thing that I noticed it that, the current string for portuguese language is "pt-BR", so the Select component won't have the correct language selected by default for who has portuguese language. I'm not sure the best way to fix this. Maybe something using the startsWith function. If you have an ideia let us know

did you found where says the stribg is `pt-BR` @zamitto?

zamitto: you can use a clean install to test and see
image

piradata: only way i can think to fix it is to convert the `pt-BR` to `pt` with split before trying to set the default, as we dont have pt-PT anyway, but that seems too "gambit" to me...

sorry, crazy week at work, only could look at it now

piradata: lol, seems that you have done just that hahah

image

well, if you dont see a problem with it, them LGTM :)

src/renderer/src/pages/settings/settings-general.tsx
```diff
@@ -79,28 +123,40 @@ export function SettingsGeneral({
         </Button>
       </div>
 
+      <h3>{t("language")}</h3>
+      <>
+        <Select value={form.language} onChange={handleLanguageChange}>
+          {languageOptions.map((language) => (
```
thegrannychaseroperation, line 129: You can potentially convert this into an `options` prop, since the `Select` children will always be `<option />`
For instance:
```tsx
<Select options={[ { value: "en", label: "English" } ]} />
```

src/renderer/src/pages/settings/settings-general.tsx
```diff
@@ -15,32 +23,53 @@ export function SettingsGeneral({
...
+    async function fetchdefaultDownloadsPath() {
+      setDefaultDownloadsPath(await window.electron.getDefaultDownloadsPath());
     }
-  }, [userPreferences]);
 
-  const { t } = useTranslation("settings");
+    fetchdefaultDownloadsPath();
+
+    setLanguageOptions(
+      Object.keys(languageResources)
```
thegrannychaseroperation, line 47: Please use lodash's `orderBy` instead
https://www.geeksforgeeks.org/lodash-_-orderby-method/

thegrannychaseroperation (changes requested):

piradata:
@thegrannychaseroperation done, used lodash-se instead of just lodash tho

thegrannychaseroperation (approved):
LGTM

zamitto (approved):

PR #407, opened 2024-05-16
Feat/index file for re export

itsOdell:
added an index.ts file within every component folder inside of `renderer/src/components` for two reasons:
1. For shorter import paths (e.g `import { Backdrop } from ../backdrop` instead of `import { Backdrop } from ../backdrop/backdrop`)
2. For more power and flexibility. Ability to export multiple files from the same UI component while being modular and reusable while still having a short import path, for example:

Components/
  /button
    button.css.ts
    button.ts
    utils.ts
    actions.ts
    ...


piradata:
lgtm

PR #411, opened 2024-05-16
feat: force developer to use yarn

itsOdell:
This solves alot of issues in PR's that have package-lock.json becuase they did npm install, this update forces an error and exits out when developer runs npm install

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict = true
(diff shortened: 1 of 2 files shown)


zamitto (approved):

piradata:
nice one, i ahve used npm more than once myself in a lotta projects kek

PR #419, opened 2024-05-16
Steam user rating

wansleynery:
- Added Steam User Rating
- Redesigned on software standards (color, size and style)
- Removed obsoleted style files

image

image


piradata:
why so much things on the commit history? did you merge instead of rebasing?

piradata:
the rating looks pretty good tho, good work :) just think would be nice to also shou half stars instead of only full stars

PR #441, opened 2024-05-18
probability of a race condition....

cardosource:
Analyzing the code, there is a probability of a "race condition" occurring. This condition can happen because the thread accesses and modifies variables while they are still in use. The global variables torrent_handle and downloading_game_id, which are being modified by the threads, are not protected, which can cause strange behaviors, such as: one thread may read the result while the variable being executed simultaneously by another thread is being modified. While one thread hasn't finished the modification, it's already being read by another, and the behavior can be different. Even a thread can present a state, but the current state hasn't been completed, resulting in a state that doesn't correspond to the actual completion.

One security measure to ensure integrity is by using locks to protect shared variables. This way, whether they are being read or modified, the result will guarantee that the current state will be intact, and inconclusive results will not occur.

An example where race conditions may occur are these places that share the variables:

def start_download(game_id: int, magnet: str, save_path: str):
    global downloading_game_id

def cancel_download():
    global downloading_game_id

def get_download_updates():
    while True:
        if downloading_game_id == 0:
            time.sleep(0.5)
            continue

The idea to ensure integrity and assert that the result is concluded and not in a state of completion is to use the RLock.

What it does is, while one thread is working, the other won't access the result before it's completed.

### DIFF
--- a/torrent-client/main.py
+++ b/torrent-client/main.py
@@ -16,6 +16,9 @@ write_fifo = Fifo(write_sock_path)
 torrent_handle = None
 downloading_game_id = 0
 
+# Global lock creation
+lock = threading.RLock()
+
 def get_eta(status):
     remaining_bytes = status.total_wanted - status.total_wanted_done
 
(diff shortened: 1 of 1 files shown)


thegrannychaseroperation:
Thanks a lot for this contribution, but we will be migrating to an Aria2 implementation instead of using Python.

PR #488, opened 2024-05-22
docs: Updated header titles for consistency

codesmith-emmy:


piradata (commented):
could you also add <a> tags on the headers you changed to be the same as the ones un the table of contents? that way when on creates a new readme based on the english one they dont have to worry about forgetting to change the linking name on the TOC (table of contents) to the headers

codesmith-emmy:
I have now added tags on the header titles to update the table of contents automatically

piradata:
LGTM

zamitto:
The lint verification is failing, could you test to see why?

thegrannychaseroperation:
No need to capitalize every single word.

Issue #228, opened 2024-05-07
Add a way to sort the library (alphabetically, by most played or download order)


hydralauncher:

vmsntos:
hello, I would like to start contributing to the project. Can I take on this task?

palexdev:
Before doing so, I would like to suggest you to move the library to a separate tab, with search options too
Having all the games on the nav bar is a bit inconvenient, especially as the list grows

Barraki:
https://github.com/hydralauncher/hydra/assets/33422772/d36986d1-f7de-46c1-9d1c-8160410a3b44

scrlkx:
I think having downloads always at the top it's a must.

PR #489, opened 2024-05-23
#228 feat: implementing sorting for library list in sidebar

Barraki:
[Task](https://github.com/orgs/hydralauncher/projects/1/views/1?pane=issue&itemId=62142229)

## Description

- Add `DropDownMenu` component
- Add logic for sorting library list in the sidebar
- Add localization for `en`, `ru`, `ua` lang for sorting list labels

Video:

https://github.com/hydralauncher/hydra/assets/33422772/d36986d1-f7de-46c1-9d1c-8160410a3b44


src/renderer/src/assets/sort-icon.svg
zamitto, line 1: no need to add this icon manually
there are suitable icons for sorting in the library that is already integrated in the project

https://primer.github.io/octicons/

src/renderer/src/hooks/use-event-listener.ts
zamitto, line 1: okay, this file sure looks like a copy paste from the internet. there seems to be no need for all these types and functions

Barraki: This hook from [usehooks-ts](https://usehooks-ts.com/react-hook/use-event-listener) library.

src/renderer/src/assets/sort-icon.svg
Barraki: Replace the icon from the library. https://primer.github.io/octicons/sort-desc-16

src/renderer/src/hooks/use-event-listener.ts
Barraki: - Refactored `use-on-click-outside` hooks and rewrote listener for detecting click-outside 
- Removed `use-event-listener` hook

PR #491, opened 2024-05-23
fixing the game installer in linux where it opens archive manager.

EMPTYVOID-DEV:
This PR addresses the issue where game installers were not being properly handled on Linux systems. The existing code was appending "setup.exe" to the game path, which resulted in an invalid path and caused the installer to be opened with the default application (e.g., Archive Manager) instead of using Lutris or Wine as intended.

The fix involves removing the unnecessary code that appends "setup.exe" to the game path, allowing the existing logic to handle opening the installer with Lutris or Wine correctly.

To reproduce the issue:

1. Open Hydra on a Linux distribution
2. Download any game
3. After the download is finished, click on "Install"

Before this change, the installer would fail to open correctly on Linux systems. After this change, the installer should open with Lutris or Wine as expected.

No other changes or regressions are introduced with this PR.


scrlkx:
Duplicated of #304

PR #518, opened 2024-05-25
Fixed incorrect translation of variable in italian translation

Giasko0:
Fixed variable that should not be translated


zamitto:
Would you like to also translate some missing strings in Italian?

image

Giasko0:
Sure. I will add the missing strings as soon as I can

Giasko0:
I added the missing translations, but i could not find game_details.settings and settings.real_debrid_api_token_description as they were not in the english locale file

zamitto (approved):
LGTM, thanks

About the variables you mentioned, I think that they are not being used anymore.

PR #522, opened 2024-05-26
Update README.md

retlAairtlA:
There is only one way to uninstall hydra on Linux/Ubuntu, to help people who are having difficulty I found these commands, they were useful for me

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ Build the Electron application by using this command:
 yarn make
 ```
 
+### Uninstalling Hydra on Linux/Ubuntu using terminal
+```bash
+sudo apt-get --purge remove hydralauncher
+```
+
+
 ## Contributors
 
 <a href="https://github.com/hydralauncher/hydra/graphs/contributors">


zamitto:
AFAIK, it is a pretty standard way to uninstall packages on Linux. I don't see why putting the command on README is needed. Did you see any Linux user who was having trouble uninstalling Hydra?

retlAairtlA:
> AFAIK, it is a pretty standard way to uninstall packages on Linux. I don't see why putting the command on README is needed. Did you see any Linux user who was having trouble uninstalling Hydra?

First it is the first time testing the pull Request, usually when I see apps in Github always shows how to install and uninstall using the terminal, in case it only shows how to make your build (which in turn needs to be adapted to other distributions Like Ubuntu) and yes I had problems to uninstall, the application itself of the error when trying to use Ubuntu's uninstalling propy, the only way to uninstall is via command in the terminal

katsu-chan:
It's better to add "Consult your distribution wiki for uninstall instructions". That would be an obvious thing to a mildly experienced user though.
It's not like you can just add instructions for 60+ package managers in readme.md. Still you can't cover all of them.

Also, @retlAairtlA, you should abstain from manually compiling/building packages until you get more experience of using linux-based oses. After all it is what are package managers for

retlAairtlA:
okee

retlAairtlA:
oke

PR #529, opened 2024-05-28
feat: adjust auto update and tray icon for macos

zamitto:
- Move sub header for auto update into own component
- Adds logic to handle auto update for MacOS
- Add tray for MacOS
- Remove duplicated types for Auto Update

Tray on MacOS


src/main/services/aria2c.ts
```diff
@@ -1,20 +1,26 @@
...
+    "--file-allocation=none",
+    "--allow-overwrite=true",
+  ]);
+
+  cp.stdout?.on("data", async (data) => {
+    const msg = Buffer.from(data).toString("utf-8");
+    logger.info(msg);
+  });
+
+  return cp;
```
zamitto, line 25: these changes will be reverted after some more testing

thegrannychaseroperation (commented):
LGTM

PR #530, opened 2024-05-28
Typo correction

Papela:
Using "**reiniciar**" (_restart_) instead of "**reinstalar**" (_reinstall_).


zamitto:
What do you think @Lianela ?

Lianela:
It works and can be added, "Haga clic" feels more generic instead of "Presiona" (means the same so it's just more about feeling more friendly) but "reiniciar" is corrected, I think it should be merged @zamitto

zamitto (approved):

PR #544, opened 2024-05-30
Feature/better repack modal 2

ChristoferMendes:
As the main feature was reverted, this PR aims to merge the feature definitively

Known issues: When a torrent throws an error after 3 attempts, it logs inside the electron server terminal. It is not breaking anything, but it can be unpleasant to debug other things in the future. I am still trying to find a way to suppress this error message, any help is welcome :)

```
Error occurred in handler for 'getMagnetHealth': Error: Failed to get seeders and peers after 3 retries.
    at Object.getSeedersAndPeers (file:///home/dev/workspace/projects/opensource/hydra/out/main/index-DrJ3egtA.js:342:43)
    at async WebContents.<anonymous> (node:electron/js2c/browser_init:2:77963)
```


zamitto:
Ok, two important things

1. As of #562, Hydra will get the download options from an external source, then we will probably pass the responsability of adding tags to the provider of the JSON
2. The seeds/peers feature would still be really useful, but I don't know if it is the library or what, but they are not loading most of the time.

image
image
image
image

zamitto:
Closing this PR, at least for now, as it is not working as we expected

PR #574, opened 2024-06-05
Feature/delete installation folder dont set game as removed

zamitto:
- Removes settings game as `removed` after removing game installer
- Sets downloadPath and folderName as null after removing game installer
- Filters games that don't have downloadPath from downloads page
- On Game Details page for games that are in the library:
  - Shows `Install` button only if game has progress === 1 and has `downloadPath`
  - Shows `Open download options` button if game has progress === 1 and does not have `downloadPath`
  - Shows button to Add/Remove from library only if progress !== 1 (i.e. game was added to library manually)


thegrannychaseroperation (approved):
LGTM

PR #634, opened 2024-06-22
Make other executable extension selectable in game-details filepicker

katsu-chan:
Executables with extensions different from ```.exe``` currently aren't selectable in the file-picker. This PR adds following extensions to file-picker filter: ```bat```, ```sh``` and ```bash```.


src/renderer/src/context/game-details/game-details.context.tsx
```diff
@@ -141,7 +141,7 @@ export function GameDetailsContextProvider({
         filters: [
           {
             name: "Game executable",
-            extensions: ["exe"],
+            extensions: ["exe", "bash", "sh", "bat"], //TODO: check if empty extension can be added, so it's possible to select linux binaries without extinsion. Add if can be added.
```
scrlkx, line 144: AFAIK empty is not a valid option but could have something like this.

```typescript
{ name: 'All Files', extensions: ['*'] }
```

zamitto:
There is one big problem in selecting other files. Hydra won't be able to get the executable name to observe its process and count the playtime

katsu-chan:
Would explicitly forbidding running games as background jobs in the documentation be a good solution?
Also, windows executables should be able to do something like linux bg jobs.
Launchers, for example, frequently give user an option to close themselves for performance boost when a game starts

src/renderer/src/context/game-details/game-details.context.tsx
katsu-chan: It would be so wonderful if it was possible to filter files by their execution permissions. I think this solution is good enough though

PR #806, opened 2024-07-09
Improvements in ES translation

Lianela:


src/locales/es/translation.json
```diff
@@ -69,12 +69,12 @@
     "minutes": "minutos",
     "amount_hours": "{{amount}} horas",
     "amount_minutes": "{{amount}} minutos",
-    "accuracy": "{{accuracy}}% precisión",
+    "accuracy": "Completar al {{accuracy}}%",
```
zamitto, line 72: I'm curious about this one. `accuracy` is how precise is the time estimate for this category on HLTB

zamitto: This change seems to alter the meaning of the text

Lianela: Yup, I change it to make it a bit more understandable, precision could be interpreted in spanish as different ways (one example, precision of skills needed to completed in that time) , but it can be changed back to "precsión"

zamitto (approved):

PR #823, opened 2024-07-12
Support building for arch based distributions

damnkrat:
AUR hydra package is very outdated. Maybe support in main branch will revive support of arch build.

Current AUR version is 1.2.4-2. It even doesn't start because of auto updater

So my plan is to make a new AUR package that takes bin package from GitHub releases directly. ??? Easy support


damnkrat:
Please rerun workflows. It failed due to 500 on download

.github/workflows/build.yml
```diff
@@ -33,6 +33,12 @@ jobs:
       - name: Build with cx_Freeze
         run: python torrent-client/setup.py build
 
+      - name: Install bsdtar
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libarchive-tools
```
zamitto, line 40: You could move these commands to the Linux block below, so it all stays together

zamitto, line 40: Electron builder documentation says to install `bdstar` package. Any reason why you need the package you added?

https://www.electron.build/multi-platform-build.html#linux

PS: I'm not a regular Linux user and I have zero experience with Arch

electron-builder.yml
```diff
@@ -1,4 +1,4 @@
-appId: site.hydralauncher.hydra
+appId: site. hydralauncher.hydra
```
zamitto, line 1: this can be reverted

.github/workflows/build.yml
damnkrat: Because bsdtar is a part of libarchive-tools. [There is no separate package for it on ubuntu](https://packages.ubuntu.com/search?keywords=bsdtar&searchon=names&suite=noble&section=all).

Maybe docs is a bit outdated

zamitto (approved):

.github/workflows/build.yml
vinycalves: > Because bsdtar is a part of libarchive-tools. [There is no separate package for it on ubuntu](https://packages.ubuntu.com/search?keywords=bsdtar&searchon=names&suite=noble&section=all).
> 
> Maybe docs is a bit outdated

This link is not working anymore. But I believe it is not possible to use the **apt** within arch, because arch does not use apt but pacman as default.

damnkrat: изображение
It is still working for me. But anyway, there is no separate package for Ubuntu.

On Arch, it seems like it is preinstalled in some package, but it doesn't matter because all Linux packages are build on Ubuntu-latest

damnkrat:
@zamitto Gentle ping

zamitto:
@damnkrat this changes will probably not be needed.
After you opened this PR and talked to us we searched a bit and now the AUR (https://aur.archlinux.org/packages/hydra-launcher-bin) should be updated with next releases correctly

damnkrat:
> this changes will probably not be needed

I think they are. Currently, AUR package unpack deb and install it content. This can result in problems like unable to use current package due to hydra update mechanism like it was in first series.

Current version still has some kind of it. Not inability to use launcher, but if I change version in PKGBULID to older one it recommends me to update with deb. If I cancel, it just restarts then recommends again. 
изображение

At least, it is not intuitive to user

damnkrat:
@zamitto As of [update schedule](https://t.me/hydralauncher/55/254159), please merge my request or fix/remove inapp update for aur

PR #835, opened 2024-07-15
Delete acc of the girl named IRINA_here

Mirsadd4:
Why not


JackEnx:
I have no idea what you want, what is the meaning of this pull?

Issue #667, opened 2024-06-24
[REQUEST] Add support an external torrent clients (Transmission Daemon, qBittorrent and etc.)


Maks1mS:
### Is your feature request related to a problem? Please describe.

In my daily life, I use the Transmission Daemon on my NAS and would like to be able to use external torrent clients as well. One of the main benefits for me would be the ability to download games even when my PC is turned off.

### Describe the solution you'd like

I would like to be able to use external torrent clients for downloading. It would be nice to start with Transmission. I also know that qBittorrent has an API.

### Additional context

**Transmission** 

Transmission's RPC specification: https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md
One of the libraries that implements Transmission's RPC: https://www.npmjs.com/package/transmission-promise

**qBittorrent**

qBittorrent WebUI API:  https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)

bpavuk:
you can install the game as usual and then add it to the Hydra library

JackEnx:
It seems like something very interesting, can you download files via torrent directly to the NAS? I don't think it's possible at the moment because it involves a certain level of research and it seems like something that wouldn't be used soo much (implemented by the core team, nothing prevents someone implementing it), perhaps in the future

Maks1mS:
> It seems like something very interesting, can you download files via torrent directly to the NAS?

Yes, I do download files directly to my NAS and then access them using SMB. In my case, the folder is mounted as a regular file system, so it looks like that it functions similarly to an external client on my local machine with a local file system. However, this is just one option. Other users may prefer to access the files using SMB or HTTP, for example.

> it seems like something that wouldn't be used soo much

Of course, I suspected it. But without creating this issue, we really don't know it. :)

> (implemented by the core team, nothing prevents someone implementing it)

I tried to implement this, but I abandoned the idea for the following reasons:
1. There are two download options in one class. I was afraid to break it and there was no way to check the Real Debrid
2. Started migration from aria2c to libtorrent with code refactoring.
3. In addition to implementation, this requires support. It may be discontinued in the future due to low usage. Also, if there was a plugin system, maybe it would be a good plugin that core team would not be responsible for.

Maybe in the future, I will return to this.

Masong19hippows:
[...]

PR #945, opened 2024-09-12
Feature/external torrent client

Masong19hippows:
Adding support for external torrent client support as in #667. I only did it with qbittorrent, but i made it so that it is easily expandable. This did also require me to make UI changes to the settings page. I put all of my changes in the "advanced" section. This is where you can select to use a different client. 

The UI for changing clients with the dropdown is ugly. The current styling isn't applied to the HTML DataList element for some reason. I tried to make it better, but was unable to do so. This might be something that someone more experienced with webdev can help with. 

I definitely want to add more feedback to the user when adding an external client. Something like a test to see if the client is up or not and handling when its not. I just wanted to grab feedback on if a change like this was even welcome before I polished anything. 

All of this was tested using my qbittorrent setup running on ubuntu server 20.04. Libtorrent version 2.0.11 and qbittorrent version 4.6.5.


Masong19hippows:
Forgot to add ss of what the changes in UI look like. I also kept the logic for the built in client the same. I don't understand how the translations for each language are inputted - if you have a team to input them or if its up to the author of PR. I only speak/write English, but I can use google translate to the best of my ability if that's something you require. 

image

image

Masong19hippows:
If this change is accepted, I feel like #946 would be a good option to add to the structure as well. All it would require is ORing the client object with a new TorrentBox Object/Entity, and then build the python api object out for it. I think one thing that needs discussed (if this is aproved) is the ability to download the files once they are done from the external client. AFAIK, there is no inherit way to do this via qbittorrent, but there is with torrentbox.

One potential way to solve this is with a checkmark box that's automatically enabled whenever you are adding a client, and this checkmark would control whether files are downloaded or not, from the client, once they are done Then there could be a corresponding object that's controlled/created for the download method. For TorrentBox, this could be an integration with their existing download solution, and for qbittorrent, this could be an implementation of the sftp protocol to use ssh to download the files. All of this combined with a checkmark box on wether to delete the torrent from the torrent once its completed, should solve all edge cases with downloading files.

Wamy-Dev:
Very nice commit, happy to see something like this, although in its current state, there is no simple way to move the file(s) from the external client (even locally hosted) to the client which has Hydra installed. 

You mentioned that TorBox does this automatically, which is a benefit, and mentioned that SFTP is an option. 

My suggestion is that there should be an easy to follow guide or even a pre-built qbittorrent docker container already setup with everything you need to download from the external client. You could put qbittorent + webdav host into a simple container which has a completion script ready and set up. This in my opinion would be the simplest way as a qbittorrent instance is already in a docker container.

If docker containers arent an option, maybe an rclone instance hosted with webdav that a user can set up. Many seed boxes already support this. It would require much more setup as well as a very thorough guide but would work.

In the end I don't honestly see the benefit other than not having to use a VPN on the host system, especially if you already have a home server. It would only be one-time setup and I am sure many users will appreciate it, so in the end I like this idea. Good work, hopefully an author will give more insight than me.

Masong19hippows:
> Very nice commit, happy to see something like this, although in its current state, there is no simple way to move the file(s) from the external client (even locally hosted) to the client which has Hydra installed. 

Yeah, I had a few ways I wanted to try and do this, but I felt that it was best left to try and discuss. I honestly don't even know if a change like this will be accepted. 

> You mentioned that TorBox does this automatically, which is a benefit, and mentioned that SFTP is an option.

I was thinking we could maybe do a list of all available download methods per client. Like, if we created a torbox entity/object, we could have a variable inside of the object labeled something like "download method" where it could be an enum of all available download methods for that client. For torbox, it would just be the built in implementation, but for qbittorrent, there could be a list of either sftp, ftp, http, etc. Then as the API is built out to allow for more integrations, this list would grow and so will the amount of implementations per client. 

> My suggestion is that there should be an easy to follow guide or even a pre-built qbittorrent docker container already setup with everything you need to download from the external client. You could put qbittorent + webdav host into a simple container which has a completion script ready and set up. This in my opinion would be the simplest way as a qbittorrent instance is already in a docker container.

This is a very good idea on how it could work for the end user. However, I feel like we shouldn't limit the possibilities here. A big reason why I did this was so that people with more complex setups could integrate this into their setup. It feels a bit backwards to take this and then limit the possibilities down to a container that the end user has to use. I feel like that as long as it's marketed in the UI for more advanced setups, the end user can do whatever they want with the client, and hydra would step back at that point.  Ex: qbittorrent running on a remote seedbox, webdav server running on a raspberry pi, etc. 

> In the end I don't honestly see the benefit other than not having to use a VPN on the host system, especially if you already have a home server. It would only be one-time setup and I am sure many users will appreciate it, so in the end I like this idea. Good work, hopefully an author will give more insight than me.

Another solution might be to have a remote paths setting similar to the arr services. Where you just provide a GUI for mounting remote filesystems in a directory the application can reach. I feel like this limits the possibilities of services that have ddl capabilities, but just another idea.

dogemanttv:
pls merge

zamitto:
@Masong19hippows could you update your branch fixing the conflicts?
Also, I think the Advanced tab should be the last one. And I think a more "critical" warning would be good, to scare curious but unexperienced users. 
Not saying we will merge this now. I'd need the rest of the team to agree on this now.
But if they agree I can help with the UI adjustments and other suggestions

src/main/events/user-preferences/update-user-client-preferencesz.ts
zamitto, line 1: also, typo on name file: "preferencesz.ts"

thegrannychaseroperation:
Closing this due to how different the downloading logic has become and no updates. If you still have interest in collaboration, please open a new pull request.

PR #985, opened 2024-09-18
fix: migration to fix missing columns

zamitto:
- Adds migration to add columns that didn't exist before 2.0.0 (if needed)
- Adds script to remove "%appdata%/hydra" and "%appdata%/hydralauncher" on uninstall
- Adds User-Agent header on HydraApi.ts with Hydra version


thegrannychaseroperation (approved):
LGTM

PR #1089, opened 2024-10-22
Added Estonian translation

Zormein:
Added Estonian ReadMe and translation.json.


zamitto:
You need to also add your translation to the src/locales/index.ts file.
You need to use the correct code for the Estonian language

zamitto:
The ISO 639-1 code for Estonian is et

docs/README.ee.md
zamitto, line 1: Could you rename this file to match the ISO code for Estonian as well?

zamitto (approved):

PR #1123, opened 2024-10-24
Update achievements.css.ts

FabricioFacco:
Bordas redondas nas conquistas


zamitto:
Hmmm, no, this makes the achievements page move away from the UI patterns Hydra has

PR #1275, opened 2024-12-04
fix: set window.userDetails after login

zamitto:


src/renderer/src/main.tsx
```diff
@@ -34,6 +35,8 @@ const Achievements = React.lazy(
   () => import("./pages/achievements/achievements")
 );
 
+console.log = logger.log;
```
Hachi-R, line 38: should this go to main or just for debugging?

Hachi-R (approved):
LGTM :)

PR #1282, opened 2024-12-09
Fix/intercepting cookies

zamitto:
- Intercepts cookies so they work even on file:// protocols
- Smal UI adjustment on Clear game paths

image

image


thegrannychaseroperation (approved):
LGTM

PR #1287, opened 2024-12-11
fix: workaround for broken or missing tray icon menu on linux

JarEXE:

**Fill in the PR content:**

Hydra's tray icon has been broken for quite some time on Linux. When right clicking, either a blank menu is displayed, or nothing at all. I have experienced this issue on various desktop environments such as Cinnamon, Gnome, and XFCE. This is probably more of an issue with Electron itself. 

This is a workaround that is meant to set the context menu on initial app load, so that Linux users will at least be able to do something with the tray icon. For example, they can now safely "Quit" the app from the tray menu if they have the "Don't hide Hydra when closing" behavior option unchecked. Without this workaround, Linux users either have to keep that option checked to remove the "hide" feature or kill the app from the CLI or their "process manager" if they left it unchecked. 

**Explaining why this is likely an Electron issue:** 
According to the Electron docs - https://www.electronjs.org/docs/latest/api/tray 

The `right-click` instance event is only supported on macOS and Windows. This event listener is being set in Hydra along with the regular `click` event, but it does absolutely nothing on Linux. The event never gets registered, therefore `showContextMenu` never runs. 

The `popUpContextMenu` method is only supported on macOS and Windows. This is what Hydra is using in the `showContextMenu` function. It does absolutely nothing on Linux.

### DIFF
--- a/src/main/services/window-manager.ts
+++ b/src/main/services/window-manager.ts
@@ -195,7 +195,7 @@ export class WindowManager {
     this.mainWindow?.focus();
   }
 
-  public static createSystemTray(language: string) {
+  public static async createSystemTray(language: string) {
     let tray: Tray;
 
     if (process.platform === "darwin") {
@@ -263,6 +263,7 @@ export class WindowManager {
         },
       ]);
 
+      tray.setContextMenu(contextMenu);
       return contextMenu;
     };
 
@@ -274,6 +275,8 @@ export class WindowManager {
     tray.setToolTip("Hydra");
 
     if (process.platform !== "darwin") {
+      await updateSystemTray();
+
       tray.addListener("click", () => {
         if (this.mainWindow) {
           if (


JarEXE:
Before workaround
hydra_actual


After workaround
hydra_tray_expected-2024-12-11_13 16 15

zamitto:
This probably even remove the need to set to "right-click" as well for other platforms as well. Thanks a lot

zamitto (approved):

JarEXE:
> Thanks a lot

@zamitto  No problem at all 🙂


> This probably even remove the need to set to "right-click" for other platforms. 

I would be a bit cautious with removing "right-click" for the other platforms. The "right-click" is calling `showContextMenu` which is in turn calling `updateSystemTray.` 

I notice that `updateSystemTray` has a nice feature where it adjusts the "most recently played" games in the context menu before it is displayed.

This is nice for macOS and Windows users, they can at least have their most recently played games updated in the menu every time they right click. 

But for the Linux users, they will just have to be satisfied with the workaround until someone else knows a better way to update the "most recently played" games every time or until Electron has better Linux support.

PR #1289, opened 2024-12-13
fix: handle achievement exception

zamitto:
Prevents unhandled exception that caused Hydra to not search for new achievements in real time


JackEnx (approved):
LGTM

Issue #914, opened 2024-08-25
[REQUEST] Add starting game with parameters


mealet:
### Is your feature request related to a problem? Please describe.

Recently I've downloaded **Dishonored** from your launcher and tried to translate it to Russian Language.
But there's a problem: game must be launched with specific argument and launcher cannot do it.

### Describe the solution you'd like

Just add `Launch Options` field to settings and append them after start command.

### Describe alternatives you've considered

Allow editing execute command for manually adding launch args

### Additional context

_Nothing here_

thisisignitedoreo:
+1
running games as an argument to the other command (e.g. proxychains or prime-run for laptops with optimus on linux) would also be great to see.

PR #1355, opened 2024-12-27
feat: add custom launch options to game

dvsouto:

- [X] I have read and understood the [Contributor Guidelines](https://github.com/hydralauncher/hydra?tab=readme-ov-file#ways-you-can-contribute).

**Fill in the PR content:**
This feature addresses the issues [#914](https://github.com/hydralauncher/hydra/issues/914) and [#1247](https://github.com/hydralauncher/hydra/issues/1247). 

A field has been added to the settings of each game to allow the addition of launch options (similar to Steam), enabling players to define parameters such as language and other game-specific configuration settings.

Print screen


src/main/events/library/open-game.ts
```diff
@@ -2,18 +2,28 @@ import { gameRepository } from "@main/repository";
...
-  await gameRepository.update({ id: gameId }, { executablePath: parsedPath });
+  await gameRepository.update({ id: gameId }, { executablePath: parsedPath, launchOptions });
 
-  shell.openPath(parsedPath);
+  if (process.platform === "linux" || process.platform === "darwin") {
+    shell.openPath(parsedPath);
+  }
+
+  if (process.platform === "win32") {
+    spawn(parsedPath, parsedParams, { shell: false });
```
zamitto, line 25: spawn is bad because if Hydra is closed, the game process would be closed as well

zamitto: at least as the way it is currently coded in this file

zamitto:
```suggestion
    spawn(parsedPath, parsedParams, { shell: false, detached: true });
```
this should be enough

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
```diff
@@ -226,6 +244,33 @@ export function GameOptionsModal({
...
+                  <>
+                    {game.launchOptions && (
+                      <Button
+                        onClick={handleClearLaunchOptions}
+                        theme="outline"
+                      >
+                        {t("clear")}
+                      </Button>
+                    )}
+                  </>
```
zamitto, line 268: Here there's no need to add a react fragment, as there only one element. this is enough:
```suggestion
                  game.launchOptions && (
                    <Button onClick={handleClearLaunchOptions} theme="outline">
                      {t("clear")}
                    </Button>
                  )
```

src/main/events/library/open-game.ts
```diff
@@ -2,18 +2,31 @@ import { gameRepository } from "@main/repository";
...
+  );
 
-  shell.openPath(parsedPath);
+  if (process.platform === "linux" || process.platform === "darwin") {
+    shell.openPath(parsedPath);
+  }
+
+  if (process.platform === "win32") {
+    spawn(parsedPath, parsedParams, { shell: false, detached: true });
+  }
```
zamitto, line 29: this feature would probably be useful for other SO as well.
but we leave it like this until we test in Linux/Mac

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
```diff
@@ -116,9 +117,26 @@ export function GameOptionsModal({
...
+
+    window.electron.updateLaunchOptions(game.id, value).then(updateGame);
+  };
+
+  const handleClearLaunchOptions = async () => {
+    setLaunchOptions("");
+
+    window.electron.updateLaunchOptions(game.id, null).then(updateGame);
+  };
+
```
zamitto, line 133: maybe a debounce would be good? 
if you spam this text input you can see that the cpu use raises a bit because it is calling 2 queries in sequence every input change

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
dvsouto: Fragment removed

src/main/events/library/open-game.ts
dvsouto: I chose to use this if statement and keep Linux/macOS as it was because I noticed that on Linux, Hydra only supports games via Wine, and on macOS, there is no Hydra release. I did a small test on Ubuntu 22.04 LTS and noticed that the `spawn` behavior is different from `shell.openPath`, but I couldn't configure Wine to test it

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
dvsouto: It really affects performance, I implemented a 1sec debounce like you suggested

zamitto (approved):

PR #1378, opened 2025-01-03
Feat: Datanodes Hoster

Shisuiicaro:

**Fill in the PR content:**

- Added Datanodes Hoster for direct downloads


src/main/services/hosters/datanodes.ts
```diff
@@ -0,0 +1,51 @@
+import axios, { AxiosResponse } from "axios";
+
+export class DatanodesApi {
+  private static session = axios.create({});
```
thegrannychaseroperation, line 4: Could you please make this property `readonly`?

src/main/services/hosters/datanodes.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  public static async getDownloadUrl(downloadUrl: string): Promise<string> {
+    const parsedUrl = new URL(downloadUrl);
+    const pathSegments = parsedUrl.pathname.split("/");
+
+    const fileCode = decodeURIComponent(pathSegments[1]);
+    const fileName = decodeURIComponent(pathSegments[pathSegments.length - 1]);
+
+    const headers: Record<string, string> = {
```
thegrannychaseroperation, line 13: I think it would be more efficient if you just passed the whole config object directly into the function call on line `39`. This way, you wouldn't have to write any types + you would inherit the types directly from `axios`.

src/main/services/hosters/datanodes.ts
```diff
@@ -0,0 +1,51 @@
...
+      adblock_detected: "",
+    });
+
+    const config = {
+      headers,
+      maxRedirects: 0,
+      validateStatus: (status: number) => status === 302 || status < 400,
+    };
+
+    const response: AxiosResponse = await DatanodesApi.session.post(
```
thegrannychaseroperation, line 39: Use `this` instead of `DatanodesApi`.

thegrannychaseroperation (changes requested):
Looks great overall. Just a few changes and it's good to go.

thegrannychaseroperation (approved):

PR #1393, opened 2025-01-11
feat: button to go up when search result is less than 10

KelvinDiasMoreira:

**Fill in the PR content:**

image


KelvinDiasMoreira:
@zamitto 
@thegrannychaseroperation

KelvinDiasMoreira:
@zamitto , @thegrannychaseroperation

davimarcilio:
nice feature

zamitto:
i don't really see why it is useful.
i mean, if the results are not enough to make the screen scroll, the click of the button does nothing

KelvinDiasMoreira:
> i don't really see why it is useful. i mean, if the results are not enough to make the screen scroll, the click of the button does nothing

I thought about it, but if the user wants to add another filter, there's no reason to scroll up. If the screen scrolls up based on the result and he wants to add another filter that's down there, he'll have to scroll down again. The button is just to say that something was found, and if he wants to, he can just click it. Currently, if I select a filter and I'm down there, I don't know if anything was found, so I have to scroll up.

PR #1488, opened 2025-03-04
simplify vector design of hydra icon (add svg file and reduce vertice…

piradata:
Hello!

Just thought about contributing again, and as i am learning how to use inkscape again that was a nice practice.

Just say if it has anything worth changing :)

Added also the original SVG file that used to make the draw importing from bitmap, so that future modifications are easier.

image


piradata:
@zamitto what do you think?

zamitto:
@piradata could you add screenshots of the Windows Tray and bar so we can compare the final result?

piradata:
There you go :) @zamitto 

image
image
image
image
image

piradata:
i think the best part tho would be the installation screen, was there that i noticed a improvement could be made

This one:

image

could not generate the install.exe to test tho, build keeps breaking here

piradata:
sadly, but ok, was a nice play with inkscape

PR #1708, opened 2025-05-19
fix: ensure achievement notification preview has same styles as actual notification [HYD-828]

zamitto:
- also adds option to disable friend starting game notification


thegrannychaseroperation (approved):
LGTM

PR #1710, opened 2025-05-19
Update Turkish

mikropsoft:
I have updated the translation according to the latest update; the previous contributor made significant changes, but I find my version more accurate. For easier comparison and separation of changes, I also left blank spaces between sections. We can also get his feedback if needed.


zamitto:
i think it was @spydea-tr helped in the Turkish translation before too

spydea-tr:
I think my translation is more accurate, well done, but it would have been better if you translated the parts that have been added with new updates and have not been translated yet, rather than trying to translate the parts that have no problems yet.

mikropsoft:
> I think my translation is more accurate, well done, but it would have been better if you translated the parts that have been added with new updates and have not been translated yet, rather than trying to translate the parts that have no problems yet. For example, starting words in a sentence with capital letters does not comply with the original text and spelling rules. You can reach me via Discord (@spydea) if you want.

Hey there, just a heads up, I don’t use Discord. Also, all the strings from the latest update are already added. I noticed in the translation changes that some parts need to be flipped to match Turkish conventions. For example, in "accuracy": "{{accuracy}}% doğruluk", the percentage in Turkish is usually on the left, so it should be "%{{accuracy}} doğruluk". I fixed a few other spots like that as well, adjusting their positions. Actually, I had updated all the translations from scratch before you made your changes, but your direct edits kind of disrupted my work a bit. Still, we can work on this together if you’d like. You’re right about the capitalization, and if you’re okay with it, we can merge this PR, and then you can open a new one to handle the capitalization fixes. I used uppercase for the headings because I thought they looked better that way. For example, in "my_library": "Kütüphane", “my library” could be translated as “kütüphanem” in Turkish, but I went with a style closer to English. We can adjust it however you prefer.

mikropsoft:
Thanks.

zamitto:
thanks guys, i'll merge it later today

mikropsoft:
> thanks guys, i'll merge it later today

By the way, you might want to consider Crowdin for translations. This allows us to translate more easily with each update.

zamitto:
> > thanks guys, i'll merge it later today
> 
> By the way, you might want to consider Crowdin for translations. This allows us to translate more easily with each update.

yeah, we will probably take a better look into it when possible

PR #1800, opened 2025-10-01
All-debrid support (fixed conflicts)

ItsYeBoi20:
This is fixing pr-1452, which had various conflicts. Hopefully this allows for easier merging.

**Fill in the PR content:**
- Added AllDebrid Support


chubbygrannychaser:
Thanks for updating it. Will give it a try today.

ItsYeBoi20:
> Thanks for updating it. Will give it a try today.

Thanks! Just as additional info, to correctly start the downloads via AllDebrid, you need to add "__init__.cp39-win_amd64.pyd", "libcrypto-1_1-x64.dll" and "libssl-1_1-x64.dll" to the "resources\hydra-python-rpc\" folder (where the newly compiled "hydra-python-rpc.exe" is)

These files are stored in "hydra-python-rpc\lib\libtorrent\" by default, but for some reason "hydra-python-rpc.exe" expects these dependencies next to the .exe itself. I'm saying this just in case you run into the same issue as me. Without these files in the root directory, pressing "download now" does nothing.

LazyClicks:
omg finally

LazyClicks:
@thegrannychaseroperation when will this pr changes be included in the main build?
thought next build after merge it will be but 3 minor versions in and its still not there

PR #1870, opened 2025-11-21
feat: Add genre translation system and support for other languagues

quirxsama:

**Fill in the PR content:**
Before this, I noticed that the Genre and Tag sections were not visible on the Catalog page for languages not supported in the application—in my example, this was Turkish—and I changed this. Now:

- Translations can be created from locales even for languages not supported by the API
- If there is no translation for a language in locales, we check the API; if it is not there either, we provide English
- New translation keys have been added for both English and Turkish.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hydralauncher",
-  "version": "3.7.5",
+  "version": "3.7.6",
   "description": "Hydra",
   "main": "./out/main/index.js",
   "author": "Los Broxas",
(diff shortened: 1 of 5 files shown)


chubbygrannychaser:
Hey @quirxsama, thanks for your contribution!
I don't think this will suffice in the long-term because having to translate dynamic data like this on i18n level is simply too complicated and might cause bugs.
If you want (let's say) us to add a specific language to our API regarding support of dynamic content, please use our feedback channel where we can discuss this further.
https://feedback.hydralauncher.gg/

PR #1960, opened 2026-01-24
Feat: Changing order of queued downloads and adding games directly to queue

Moyasee:
… translations

**Fill in the PR content:**
- Added new buttons to context menu of queued download - Move Up, Move Down
- changed behavior of repacks-modal if there's an active download, now the button "Download now" will be changing to "Add to queue" and games would be adding to queue without interfering current active download.
- Fixed color pinking for download chart in downloads page, now if the color is too similar to the background color of chart - bars will be using white color as a fallback


chubbygrannychaser:
CI green. Solid refactors and queue actions. A couple of things to double-check:

1) Queue order semantics: queued list is ordered by timestamp DESC. In updateDownloadQueuePosition, “move up” decrements index — that may move the item toward newer timestamps, which could be the opposite of expected (older items first). Worth confirming intended ordering.

2) cancelDownload now only cancels if it’s the active download. Make sure queued items still clear correctly via cancel-game-download and UI state updates.

Also: PR has no label (feature/fix/etc) — please add one.

Moyasee:
> CI green. Solid refactors and queue actions. A couple of things to double-check:
> 
>     1. Queue order semantics: queued list is ordered by timestamp DESC. In updateDownloadQueuePosition, “move up” decrements index — that may move the item toward newer timestamps, which could be the opposite of expected (older items first). Worth confirming intended ordering.
> 
>     2. cancelDownload now only cancels if it’s the active download. Make sure queued items still clear correctly via cancel-game-download and UI state updates.
> 
> 
> Also: PR has no label (feature/fix/etc) — please add one.

Verified both:
Queue order  - "Move up" correctly swaps timestamps so the item appears higher and downloads sooner. Both frontend and processNextQueuedDownload use timestamp DESC, so index 0 = first to download.
Cancel queued items  - cancelDownload only stops the active download when downloadKey === downloadingGameId. Queued items just get marked "removed" and cleaned up via deleteGameFolder.

PR #1985, opened 2026-02-05
feat: implement retry mechanism and stall detection in JsHttpDownloader

Moyasee:

**Fill in the PR content:**


src/main/services/download/js-http-downloader.ts
```diff
@@ -22,6 +22,25 @@ export interface JsHttpDownloaderOptions {
   headers?: Record<string, string>;
 }
 
+const MAX_RETRY_ATTEMPTS = 10;
+const INITIAL_RETRY_DELAY_MS = 1000;
+const MAX_RETRY_DELAY_MS = 15000;
+const STALL_TIMEOUT_MS = 8000;
```
thegrannychaseroperation: stall detection: interval is 5s and timeout is 8s, so worst-case detection can be ~13s (depending on alignment). not necessarily wrong, but if you want “~8s stall”, consider checking more frequently (e.g., 1s/2s) or bump the timeout to match the sampling interval.

src/main/services/download/js-http-downloader.ts
```diff
@@ -45,35 +71,161 @@ export class JsHttpDownloader {
...
-    try {
-      await this.executeDownload(
-        url,
-        requestHeaders,
-        filePath,
-        startByte,
+  private async startDownloadWithRetry(): Promise<void> {
+    if (!this.currentOptions) return;
+
+    while (this.retryCount <= MAX_RETRY_ATTEMPTS && !this.isPaused) {
```
thegrannychaseroperation: possible off-by-one: `while (this.retryCount <= MAX_RETRY_ATTEMPTS)` will allow 11 total attempts if `retryCount` counts retries (0..10). If you intend “max 10 retries”, prefer `< MAX_RETRY_ATTEMPTS` in the loop condition and keep the same meaning between `retryCount` and the constant.

PR #2117, opened 2026-03-25
Feat/lbx 650

japa2k:

**Fill in the PR content:**


src/renderer/src/pages/settings/settings-context-downloads.tsx
```diff
@@ -53,9 +53,10 @@ export function SettingsContextDownloads() {
     useNativeHttpDownloader: true,
     seedAfterDownloadComplete: false,
     showDownloadSpeedInMegabytes: false,
-    extractFilesByDefault: true,
+    extractFilesByDefault: false,
```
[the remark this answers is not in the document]
japa2k: That change wasn't intended for this PR. It was left over from a local behavior test I was running. I'll revert it.

src/renderer/src/pages/settings/settings-context-downloads.tsx
```diff
@@ -77,6 +78,8 @@ export function SettingsContextDownloads() {
               userPreferences.showDownloadSpeedInMegabytes ?? false
             )
           : "",
+      deleteArchiveFilesAfterExtractionByDefault:
```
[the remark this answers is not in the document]
japa2k: Here are two screenshots from DevTools showing the userPreferences object in LevelDB with `deleteArchiveFilesAfterExtractionByDefault`

Screenshot 2026-03-30 112234
Screenshot 2026-03-30 112310

The key is being persisted and read back correctly. No data loss for existing users since this is a new key

src/main/events/torrenting/start-game-download.ts
```diff
@@ -61,16 +62,16 @@ const startGameDownload = async (
...
     fileIndices,
     selectedFilesSize,
     fileSize: selectedFilesSize ?? null,
   };
 
   try {
-    await DownloadManager.startDownload(download).then(() => {
-      return downloadsSublevel.put(gameKey, download);
-    });
+    await downloadsSublevel.put(gameKey, download);
```
chubbygrannychaser: Persisting the download before `DownloadManager.startDownload()` regresses the failure path. If `startDownload()` throws while resolving the URL / validating credentials, we now leave a queued+active record in LevelDB even though nothing actually started. The previous order only wrote after a successful start, so please either restore that ordering or clean up the inserted record in the `catch` block.

src/renderer/src/pages/game-details/modals/download-settings-modal.tsx
```diff
@@ -765,6 +781,9 @@ export function DownloadSettingsModal({
     if (!visible) {
       setShowTorrentStepModal(false);
     }
+    if (userPreferences?.extractFilesByDefault) {
```
chubbygrannychaser: This re-open/reset effect only ever forces the toggle to `true`. If the user preference is `false`, closing and reopening the modal keeps whatever stale value was left from the previous session instead of resetting back to the preference. Please set the state from `userPreferences?.extractFilesByDefault ?? true` on open, rather than gating on truthiness.

chubbygrannychaser (changes requested):
Found two regressions that should be fixed before merge; see inline comments.

src/renderer/src/pages/game-details/modals/download-settings-modal.tsx
```diff
@@ -465,6 +472,15 @@ export function DownloadSettingsModal({
     setSelectedDownloader(getDefaultDownloader(availableDownloaders));
   }, [getDefaultDownloader, userPreferences?.downloadsPath, downloadOptions]);
 
+  useEffect(() => {
```
chubbygrannychaser, line 479: This modal rehydrates `deleteArchiveFilesAfterExtraction` on open, but `automaticExtractionEnabled` is still only initialized once from props. If the user toggles extraction, closes the modal, and reopens it, this checkbox can keep stale local state instead of reflecting `userPreferences?.extractFilesByDefault ?? true`. Please reset that state on open too, the same way you do for the new delete-archive toggle.

src/main/services/game-files-manager.ts
```diff
@@ -186,10 +186,28 @@ export class GameFilesManager {
...
+          valueEncoding: "json",
+        }),
+      ]);
+
+      const shouldDelete =
+        download?.automaticallyDeleteArchiveFiles ??
+        userPreferences?.deleteArchiveFilesAfterExtractionByDefault ??
+        false;
+
+      if (shouldDelete) {
```
chubbygrannychaser, line 202: The automatic-delete path bypasses the existing `deleteArchive` flow and calls `fs.unlink()` directly. That loses the extra bookkeeping in `src/main/events/library/delete-archive.ts` (notably clearing `installerSizeInBytes`, plus centralized logging/error handling), so auto-deletion can leave stale installer metadata behind. Could this reuse the same deletion path instead of unlinking files inline?

src/main/services/game-files-manager.ts
```diff
@@ -186,10 +186,31 @@ export class GameFilesManager {
...
+        for (const archivePath of archivePaths) {
+          WindowManager.mainWindow?.webContents.send(
+            "deleteArchive",
```
chubbygrannychaser: This auto-delete path is sending a renderer event named `deleteArchive`, but there is no `ipcRenderer.on("deleteArchive", ...)`/preload subscription for that channel. The existing `deleteArchive` implementation is an `ipcRenderer.invoke()` -> `registerEvent("deleteArchive", ...)` request in the opposite direction, so this branch becomes a no-op and archives never get deleted automatically. Please route this through the existing main-side deletion handler instead (or add a real renderer listener, though that seems less direct).

chubbygrannychaser (changes requested):
Found one blocking issue: the new auto-delete branch in `GameFilesManager` emits a `deleteArchive` renderer event, but Hydra only exposes `deleteArchive` as a renderer->main invoke handler. As written, the preference can be enabled without actually deleting the archive files. See inline comment.

chubbygrannychaser (approved):
LGTM

PR #2148, opened 2026-04-12
Feature/lbx 668

japa2k:

**Fill in the PR content:**


src/renderer/src/pages/catalogue/release-year-section.tsx
```diff
@@ -0,0 +1,149 @@
...
+                  rgba(255,255,255,0.12) 0%,
+                  rgba(255,255,255,0.12) ${gtePercent}%,
+                  ${color} ${gtePercent}%,
+                  ${color} ${ltePercent}%,
+                  rgba(255,255,255,0.12) ${ltePercent}%,
+                  rgba(255,255,255,0.12) 100%
+                )`,
+              }}
+            />
+            <input
```
chubbygrannychaser, line 126: These two range inputs need accessible names (`aria-label`/`aria-labelledby`, or visible labels). As written, screen readers will encounter two unlabeled sliders, so users cannot tell which thumb controls the start year vs. the end year.

PR #2161, opened 2026-04-21
Feat/navigation big picture

Hachi-R:
big thanks to eco verde


src/big-picture/src/main.tsx
```diff
@@ -0,0 +1,23 @@
...
+import Catalogue from "./pages/catalogue/catalogue";
+import Downloads from "./pages/downloads/downloads";
+import LibraryPage from "./pages/library/page";
+import Settings from "./pages/settings/settings";
+
+ReactDOM.createRoot(document.getElementById("root")!).render(
+  <React.StrictMode>
+    <BrowserRouter>
+      <Routes>
+        <Route path="/" element={<App />}>
```
chubbygrannychaser, line 20: There is no index/default child route here, so opening Big Picture at `/big-picture` leaves the `<Outlet />` empty. Since the sidebar exposes a `Home` entry pointing at the base path, it would be good to either add an index route or redirect the base route to one of the real pages.

src/big-picture/src/app.tsx
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(() => {
+    if (!IS_DESKTOP) {
+      document.documentElement.style.colorScheme = "dark";
+    }
+  }, []);
+
+  return (
+    <Fragment>
+      <NavigationStateBridge />
+      <NavigationDiagnostics />
```
chubbygrannychaser: Should this diagnostics surface ship by default? Right now every Big Picture session gets an always-available debug panel/button. If this is mainly for development/gamepad tuning, I would gate it behind a dev flag or an explicit debug setting before merging.

src/big-picture/src/helpers/gamepad-layout.ts
```diff
@@ -0,0 +1,346 @@
...
+const STANDARD_GAMEPAD_LAYOUT = GAMEPAD_LAYOUTS.find(
+  (layout) => layout.name === "Standard Gamepad"
+)!;
+
+const LINUX_STANDARD_GAMEPAD_LAYOUT = GAMEPAD_LAYOUTS.find(
+  (layout) => layout.name === "Linux Standard Gamepad"
+)!;
+
+export const getGamepadLayout = (gamepad: globalThis.Gamepad) => {
+  console.log("getGamepadLayout", gamepad.id);
```
chubbygrannychaser: Looks like a leftover debug log. Since `getGamepadLayout()` runs in the core input path, this will spam the console for every detected controller in normal usage. I’d remove it (or route it through a gated debug logger) before merging.

PR #2170, opened 2026-04-24
Add game transfer with progress bar, cancel, and drive picker

yassine808:
**When submitting this pull request, I confirm the following (please check the boxes):**

- [x] I have read the [Hydra documentation](https://docs.hydralauncher.gg/getting-started.html).
- [x] I have checked that there are no duplicate pull requests related to this request.
- [x] I have considered, and confirm that this submission is valuable to others.
- [x] I accept that this submission may not be used and the pull request may be closed at the discretion of the maintainers.

**Fill in the PR content:**

##  What this does
Added game file transfer feature allowing users to move installed games between drives/folders directly from the game options modal, with progress tracking and cancel support.

##  Features
-  **Drive selector** — pick a drive (auto-creates `Hydra Games` folder) or enter a custom path
-  **Space indicator** — shows free space, used space, and game size with color-coded usage bars
-  **Space guard** — blocks transfer if destination doesn't have enough free space
-  **Live progress** — real-time bar with speed (MB/s) and ETA (seconds → minutes → hours → days)
-  **Cancel with confirmation** — stops all parallel copies instantly and cleans partial files
-  **Multiple transfers** — move several games at once, each tracked independently
-  **8 parallel streams** — optimized cross-drive copying

###  New files
| File | What it does |
|------|-------------|
|  `transfer-game-files.ts` | Backend copy engine, cancel, space check |
|  `general-section.scss` | Dark-themed styles with green accents |

###  Modified files
| File | What changed |
|------|-------------|
|  `game-options-modal.tsx` | Added IPC listeners, drive fetching, transfer UI |
|  `general-section.tsx` | Added drive picker, progress bar, cancel modal |
|  `hero-panel-actions.tsx` | Added "Transferring 45%" button during moves |
|  `game-details.context.tsx` | Added shared transfer state across components |

##  Screenshots

Transfer game button

<p align="center"><strong>Transfer button</strong> — appears below executable path when game is installed</p>

clicking transfer

<p align="center"><strong>Drive selector</strong> — choose a drive or enter a custom path</p>

<p align="center">image

<p align="center"><strong>Error toast</strong> — alerts when destination already has the game</p>

moving files

<p align="center"><strong>Progress window</strong> — real-time speed, ETA, and cancel option</p>

<p align="center">image

<p align="center"><strong>Success msg</strong> — Says the game has been moved </p>

image

<p align="center"><strong>Hero panel button</strong> — displays transfer progress on the game page</p>

PR #2179, opened 2026-04-25
feat: transfer game feature (updated #2170)

Moyasee:

**Fill in the PR content:**
Improvements / fixes for #2170


src/main/events/library/transfer-game-files.ts
```diff
@@ -0,0 +1,390 @@
...
+    ...game,
+    executablePath: newExePath,
+    installedSizeInBytes,
+  });
+
+  const download = await downloadsSublevel.get(gameKey).catch(() => null);
+  if (download) {
+    await downloadsSublevel.put(gameKey, {
+      ...download,
+      downloadPath: path.dirname(newExePath),
```
chubbygrannychaser: This rewrites the persisted install root from the executable path. That works only when the executable lives directly under the game root, but many installs keep the exe in a nested subdirectory (`Game/Binaries/...`). Other parts of Hydra reconstruct the install path as `path.join(download.downloadPath, download.folderName)`, so after a transfer this can point at the wrong directory (for example `<target>/Binaries/Binaries`). Could we preserve the actual transferred root here instead of deriving it from `newExePath`?

src/main/events/library/transfer-game-files.ts
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    const relExe = path.relative(gameRoot, game.executablePath);
+    const newExePath = path.join(targetRoot, relExe);
+    const gameKey = levelKeys.game(shop, objectId);
+
+    try {
+      await updateDatabaseAfterTransfer(game, gameKey, newExePath, gameSize);
+    } catch {
+      await cleanupOnError(id, targetRoot);
```
chubbygrannychaser: At this point `moveGame()` has already removed the source directory. If the later DB update fails, `cleanupOnError()` deletes `targetRoot` too, so a transient LevelDB failure can leave the user with neither the old install nor the moved copy. We should avoid deleting the destination after a successful move unless we can roll back safely.

src/main/events/library/transfer-game-files.ts
```diff
@@ -0,0 +1,390 @@
...
+
+    // Validate game root
+    const rootValidation = await validateGameRoot(game);
+    if (!rootValidation.valid) {
+      activeTransfers.delete(id);
+      return { ok: false, error: rootValidation.error };
+    }
+    const gameRoot: string = rootValidation.gameRoot;
+
+    const folderName = path.basename(gameRoot);
```
chubbygrannychaser, line 329: I do not see a guard that rejects an already-existing `targetRoot`. With the current flow we create the destination parent, copy into `path.join(destParent, folderName)`, and later remove the source. If that folder already exists, this can merge into or overwrite existing data, and an error/cancel path may then recursively delete that pre-existing destination folder. Can we fail early unless `targetRoot` does not exist (or is confirmed empty)?

src/renderer/src/pages/game-details/hero/hero-panel-actions.tsx
```diff
@@ -213,6 +216,22 @@ export function HeroPanelActions() {
...
 
   const gameActionButton = () => {
+    if (isTransferring) {
+      const percent = Math.round(transferProgress * 100);
+      return (
+        <Button
+          theme="outline"
+          className="hero-panel-actions__action"
+          onClick={() => {
+            setGameOptionsInitialCategory("general");
```
chubbygrannychaser: While a transfer is running, this button opens the modal on the `general` category, but the transfer controls/progress UI live under the new `locations` category. That means the CTA for an active transfer lands on the wrong tab, so the user does not immediately see the transfer status / cancel controls they clicked for.

chubbygrannychaser (changes requested):
Found a few blocking issues in the new transfer flow; see inline comments.

Main blockers:
- transfer rewrites `downloadPath` / `folderName` from the executable path instead of the moved install root
- a DB update failure after a successful move can delete both source and destination
- destination collisions are not rejected before copying

Also left one UX comment about the transfer CTA opening the wrong tab while a transfer is in progress.

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
```diff
@@ -248,46 +269,57 @@ export function GameOptionsModal({
...
+  };
+
+  const handleStartTransfer = async (destPath: string) => {
+    const result = await window.electron.transferGameFiles(
+      game.shop,
+      game.objectId,
+      destPath
+    );
+    if (!result.ok) {
+      showErrorToast(result.error || `Transfer failed for ${game.title}`);
```
chubbygrannychaser: These transfer toasts are still hardcoded / raw backend strings (`Transfer cancelled`, `not_enough_space`, etc.). Can we map the transfer outcomes to translated UI messages before surfacing them here?

src/renderer/src/context/game-details/game-details.context.tsx
```diff
@@ -114,6 +119,68 @@ export function GameDetailsContextProvider({
...
+        updateGame();
+      }
+    };
+
+    const onTransferCancelled = (
+      _: unknown,
+      shop: string,
+      objectId: string
+    ) => {
+      console.log(
```
chubbygrannychaser: Could we remove these `console.log` statements before merge? The project rules prefer the logger over console output, and this looks like temporary transfer-debug scaffolding left in the renderer path.

chubbygrannychaser (commented):
Re-reviewed the updated head. The earlier blocking transfer-flow issues I flagged are now addressed:
- install root metadata now uses the moved target root
- destination collisions are rejected
- DB update failure no longer deletes the moved copy
- the transfer CTA now opens the locations tab

I left two small follow-up inline comments that look non-blocking (hardcoded transfer messages / debug console logs). From my side, I do not see a remaining blocker in the latest head.

yassine808:
@chubbygrannychaser  
Hey! I'm the author of #2170 which this PR was based on — I wrote the majority of the commits (30+) that made it into #2179. I don't appear in the contributors graph or in the v3.9.6 release notes though, likely due to how the branch/merge was handled.

Would you be able to amend the merge commit to add:

Co-authored-by: yassine808 <ysmagri@gmail.com>

And if possible, mention me in the v3.9.6 release notes since the transfer feature originated from my PR.

Thanks in advance !!

PR #2192, opened 2026-04-29
Bp boot hotfix

Hachi-R:
bitch dont kill my vibe


chubbygrannychaser (changes requested):
Thanks for the boot-flow rework. I found two blocking issues in the current version:

1. Cold-starting directly into Big Picture removes the always-created hidden main window, but several existing update and event paths still only send to WindowManager.mainWindow. With a BP-only cold start, those sends can become no-ops.
2. Unknown or future deep links can now suppress normal window creation without a fallback visible window, so the app can appear to do nothing.

Because both are behavioral regressions, I am requesting changes.

PR #2193, opened 2026-04-29
Feature/lbx 703 big picture make up 

Hachi-R:
my legs hurt


src/main/services/download-orchestrator.ts
```diff
@@ -0,0 +1,586 @@
...
+      await this.activateDownload(download);
+      const nextDownloads = await this.getAllDownloads();
+      await setDownloadLayoutQueues(nextDownloads, queueIds, pausedIds);
+      WindowManager.sendDownloadsUpdated();
+      return true;
+    }
+
+    if (targetArea === "queue") {
+      if (isHero && isActiveLikeDownload(download)) {
+        await this.pauseDownload(download, {
```
chubbygrannychaser: If the active hero download is moved to the queue, this branch pauses it with `reason: "paused"` and `startNextQueued: false`, which leaves the persisted item in the paused state. After that we only insert its id into `queueOrder`, so `setDownloadLayoutQueues()` will normalize it as paused rather than queued. I think this makes "send to queue" behave like "pause" for the active hero item.

src/big-picture/src/pages/library/page.tsx
```diff
@@ -1,30 +1,228 @@
...
+      (game: LibraryGame) => {
+        openDownloadModalFromContextMenu(game);
+      },
+      [openDownloadModalFromContextMenu]
+    )
+  );
+
+  const logOptionsPlaceholder = useCallback(
+    (game: LibraryGame, source: "library-hero" | "library-context-menu") => {
+      console.log("big-picture options placeholder", {
```
chubbygrannychaser: These handlers are currently wired into real context-menu actions, but they only log placeholders. That means `Options`, `Uninstall`, and `Remove from Library` are exposed as functional actions in the new UI while doing nothing at runtime. I think we should either hook these up before merge or hide/disable the actions for now.

src/big-picture/src/components/common/checkbox/index.tsx
```diff
@@ -40,23 +45,32 @@ export const Checkbox = ({ label, ...props }: Readonly<CheckboxProps>) => {
...
+        >
+          {isChecked && (
+            <CheckIcon className="checkbox__input__icon" size={14} />
+          )}
+        </button>
+      </FocusItem>
 
       {label && (
-        <label className="checkbox__label" id={`${id}-label`} htmlFor={id}>
+        <label className="checkbox__label" id={`${id}-label`}>
```
chubbygrannychaser: Dropping `htmlFor={id}` here looks like a regression: in the non-`block` variant, clicking the visible label no longer toggles the checkbox because the label is no longer associated with the control.

src/big-picture/src/components/common/user-disk-item/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+        id={focusId}
+        navigationOverrides={focusNavigationOverrides}
+        navigationState={focusNavigationState}
+        asChild
+      >
+        <button
+          type="button"
+          className={rootClassName}
+          onClick={onClick}
+          onFocus={onClick}
```
chubbygrannychaser: Calling `onClick` from `onFocus` means just moving focus with keyboard/gamepad changes the selected install directory before the user confirms the choice. That feels pretty risky for controller UX — would it be better to keep focus and selection separate, and only update on explicit activation?

src/big-picture/src/hooks/use-game-details.hook.ts
```diff
@@ -37,11 +37,7 @@ export function useGameDetails(objectId: string, shop: GameShop) {
     setIsLoading(true);
 
     const [shopDetailsResult, statsResult, assets] = await Promise.all([
-      globalThis.window.electron.getGameShopDetails(
-        objectId,
-        shop,
-        navigator.language
-      ),
+      globalThis.window.electron.getGameShopDetails(objectId, shop, "en"),
```
chubbygrannychaser: This hardcodes shop details to English. Was that intentional? It looks like a localization regression compared with the rest of the app, since big-picture game details will ignore the user's language here.

chubbygrannychaser (approved):
Approved per maintainer instruction for head c486876b36184736fda58e35ddcbe86f37ab3973.

PR #2208, opened 2026-05-04
chore: update to electron 40

zamitto:


chubbygrannychaser (changes requested):
Thanks for the Electron 40 update. I found a blocking regression in the current version:

- Big Picture copy actions now hard-require window.electron.clipboard.writeText(...) in places that previously used navigator.clipboard.writeText(...). In the browser or non-desktop Big Picture flow, window.electron is not available, so those copy actions will fail.

Please keep an Electron path for desktop, but preserve a browser-safe fallback when window.electron is unavailable.

PR #2217, opened 2026-05-05
Update Polish translation

juliazero:
Update Polish translation using Weblate for hydralauncher/hydra

Current translation status:

Weblate translation status

Converted as draft until # 2222 merge.


chubbygrannychaser (changes requested):
Thanks for the translation update. I found a blocking regression in the current version:

- The PR removes the _zero plural forms for several existing count-based Polish strings. Current call sites do pass zero counts, so this changes zero-state UI copy from dedicated "Brak ..." messages to generic pluralized forms like "0 opcji pobierania".

Please keep the zero-count variants for the existing count-based keys before merge.

chubbygrannychaser (approved):
LGTM. The previous pluralization blocker looks fixed on the current head — the dedicated Polish zero forms are preserved and the new plural keys are consistent with Polish plural categories.

Issue #2223, opened 2026-05-07
feature req: change default folder in which new wine prefixes are created


Elxyxs:

PeMeireles:
I might look into it

PR #2224, opened 2026-05-08
feat: change default wine prefix folder

PeMeireles:

**Fill in the PR content:**

This PR solves  [#2223](https://github.com/hydralauncher/hydra/issues/2223) by adding a setting to change the default folder Hydra uses for new Wine prefixes.
The new option is available in Settings > Compatibility and lets users choose a different base directory for Hydra-managed Wine prefixes. Hydra still creates one prefix per game, but now uses the configured folder instead of only the built-in default.
To avoid breaking existing installs, changing the default does not move already installed Windows games to the new prefix path. Existing games keep their previous resolved prefix, while the new default only applies to future default-managed prefixes.
image
image


chubbygrannychaser (approved):
LGTM. I reviewed the changed Wine-prefix preference flow and did not find a concrete code-level issue in the current diff.

PR #2225, opened 2026-05-09
feat: Steam account integration (link account, sync achievements & library)

LED-N:
## What this PR does

Adds full Steam account integration to Hydra, allowing users to:

- **Link their Steam account** via Steam OpenID 2.0 (ownership verified, not just SteamID lookup)
- **Import achievements** for individual Steam games from the game's achievement page
- **Batch sync all achievements** across the entire Hydra library, with smart rate-limit retry (no fixed delay — only pauses on HTTP 429)
- **Sync Steam library** — imports all owned Steam games with correct playtime (takes the highest value between Steam and Hydra) and icons from the Steam CDN
- **Library source filter** — filter your library by All / Steam imported / Hydra only
- **Non-blocking UX** — progress modals are closeable while sync continues in background, with a cancel button
- **Steam display name + avatar** shown in settings instead of raw SteamID64
- **Auto-sync on startup** — optional checkbox in settings
- **EN + FR translations** included

## Technical notes

- Rate limiting: requests fire as fast as possible; on HTTP 429, exponential backoff (5s → 10s → 15s, up to 3 retries) instead of a fixed per-request delay.
- Playtime sync uses the absolute `/playtime` endpoint so the profile page reflects the correct value.
- A Steam Web API key is required per user (free at steamcommunity.com/dev/apikey) — no shared key is possible due to Valve ToS and per-key rate limits.

## Notes on Steam authentication

Steam login uses the OpenID 2.0 flow with a temporary local HTTP server
as the callback endpoint (`openid.realm = http://127.0.0.1:<random_port>`).
This is the standard approach for desktop applications — Valve explicitly
supports `127.0.0.1` as a valid realm. As a result, Steam's login page
displays "Sign in through 127.0.0.1" to the user, which is expected
behavior for native apps.

If the team would prefer to show a branded domain (e.g. `hydralauncher.gg`),
this would require routing the OpenID callback through a backend endpoint
and returning the SteamID to the app via a custom protocol link
(`hydra://steam-auth?steamid=...`). Happy to explore this if preferred,
but kept the self-contained local server approach for simplicity.

## Checklist (from PR template)

- [x] I have read the Hydra documentation
- [x] I have checked that there are no duplicate pull requests related to this request
- [x] I have considered, and confirm that this submission is valuable to others
- [x] I accept that this submission may not be used and the pull request may be closed at the discretion of the maintainers

### Settings — Steam integration (Not yet linked)
Settings Steam Not yet linked

### Settings — Steam integration (Linking)
Settings Steam Linking

### Settings — Steam integration (Linked)
Settings Steam Linked

### Library source filter
Library filter

### Sync progress modal
Sync progress

### Sync from Steam button (achievements)
Sync from Steam


moi952:
Hi, I tried it, it's nice.

However, I thought the integration also allowed you to launch the game installation (via Steam of course).
Do you think that could be added?

Thanks

Hachi-R:
i'll be closing this for now because i don't think this is something we should maintain directly on the launcher side.

if we decide to support this kind of integration, it should probably be implemented on the api side instead, so we can handle it in a more centralized and safer way.

PR #2244, opened 2026-05-19
feat(library): support auto-launching setup.exe for Repacks

TeALO36:
Adds support for repacks:
- Implements automatic launch of setup.exe for repacked games when appropriate.

### DIFF
--- a/python_rpc/main.py
+++ b/python_rpc/main.py
@@ -5,9 +5,20 @@ import re
 import sys
 import tempfile
 import threading
-import time
 import urllib.parse
 from typing import Any, Optional
+import sys
+import os
+
+if sys.platform == "win32":
+    dll_dir = os.path.join(sys.base_prefix, "DLLs")
+    if os.path.isdir(dll_dir):
+        try:
+            os.add_dll_directory(dll_dir)
+        except Exception:
+            pass
+
+print(f"USING PYTHON: {sys.executable}", file=sys.stderr)
 
 import libtorrent as lt
 
@@ -340,10 +351,12 @@ def status():
         downloader = downloads.get(downloading_game_id)
 
     if not downloader:
+        logger.debug("status(): downloader not found for %s", downloading_game_id)
         return None
 
     status_payload = downloader.get_download_status()
     if not status_payload:
+        logger.debug("status(): status_payload is None for %s", downloading_game_id)
         return None
 
     return status_payload
@@ -437,7 +450,7 @@ def action(data: Optional[dict] = None):
     if not action_name:
         raise RpcError("invalid_action")
 
-    requires_game_id = {"start", "pause", "cancel", "resume_seeding", "pause_seeding"}
+    requires_game_id = {"start", "pause", "cancel", "resume_seeding", "pause_seeding", "force_recheck"}
     if action_name in requires_game_id and not game_id:
         raise RpcError("invalid_game_id")
 
@@ -511,6 +524,37 @@ def action(data: Optional[dict] = None):
 
             for downloader in active_downloaders:
                 apply_download_limit(downloader)
+        elif action_name == "force_recheck":
+            logger.info("force_recheck received for %s with data: %s", game_id, data)
+            downloading_game_id = game_id
+            with downloads_lock:
+                downloader = downloads.get(game_id)
+
+            if not downloader and "url" in data and "save_path" in data:
+                logger.info("force_recheck: creating new downloader (paused)")
+                # Use paused flag so the torrent doesn't start downloading immediately
+                start_torrent_download(
+                    game_id,
+                    data["url"],
+                    data["save_path"],
+                    flags=lt.torrent_flags.paused,
+                )
+                with downloads_lock:
+                    downloader = downloads.get(game_id)
+
+            if downloader:
+                logger.info("force_recheck: pausing then executing force_recheck on downloader")
+                downloading_game_id = game_id
+                # Pause the torrent first so force_recheck doesn't trigger a download
+                with downloads_lock:
+                    handle = getattr(downloader, "torrent_handle", None)
+                    if handle and handle.is_valid():
+                        handle.pause()
+                        handle.unset_flags(lt.torrent_flags.auto_managed)
+                downloader.force_recheck()
+                logger.info("force_recheck: force_recheck() called successfully for %s", game_id)
+            else:
+                logger.warning("force_recheck: failed to get or create downloader")
         else:
             raise RpcError("invalid_action")
     except RpcError:
--- a/src/main/events/library/verify-game-integrity.ts
+++ b/src/main/events/library/verify-game-integrity.ts
@@ -1,0 +1,106 @@
+import { registerEvent } from "../register-event";
+import { downloadsSublevel, gamesSublevel, levelKeys } from "@main/level";
+import { DownloadManager } from "@main/services/download/download-manager";
+import { Downloader } from "@shared";
+import fs from "node:fs";
+
+import { logger } from "@main/services";
+import type { Download, GameShop } from "@types";
+
+const verifyGameIntegrity = async (
+  _event: Electron.IpcMainInvokeEvent,
+  shop: GameShop,
+  objectId: string,
+  fallbackUri?: string,
+  fallbackDownloadPath?: string
+) => {
+  const gameKey = levelKeys.game(shop, objectId);
+  logger.info(`[VerifyIntegrity] Called for ${gameKey}`);
+
+  let download = await downloadsSublevel.get(gameKey).catch(() => null);
+  const game = await gamesSublevel.get(gameKey).catch(() => null);
+
+  if (!game) {
+    logger.error(`[VerifyIntegrity] Game not found in DB for ${gameKey}`);
+    throw new Error("Jeu introuvable dans la base de données.");
+  }
+
+  // If there's no download entry but we have fallback info (uri + path), create a minimal one
+  if (!download) {
+    const uri = fallbackUri;
+    const downloadPath = fallbackDownloadPath;
+
+    if (uri && downloadPath) {
+      logger.info(
+        `[VerifyIntegrity] No download entry found, creating from fallback uri for ${gameKey}`
+      );
+
+      const newDownload: Download = {
+        shop: shop as any,
+        objectId,
+        uri,
+        downloadPath,
+        folderName: null,
+        progress: 0,
+        downloader: Downloader.Torrent,
+        bytesDownloaded: 0,
+        fileSize: null,
+        shouldSeed: false,
+        status: "paused",
+        queued: false,
+        pinnedToHero: false,
+        timestamp: Date.now(),
+        extracting: false,
+        automaticallyExtract: false,
+        automaticallyDeleteArchiveFiles: false,
+      };
+
+      await downloadsSublevel.put(gameKey, newDownload);
+      download = newDownload;
+      logger.info(`[VerifyIntegrity] Created download entry for ${gameKey}`);
+    } else {
+      // No fallback: check if the executable at least exists
+      logger.error(
+        `[VerifyIntegrity] Download not found in DB for ${gameKey} and no fallback provided`
+      );
+
+      if (game.executablePath && fs.existsSync(game.executablePath)) {
+        logger.info(
+          `[VerifyIntegrity] Executable exists, cannot run torrent recheck without uri`
+        );
+        throw new Error(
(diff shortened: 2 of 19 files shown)


python_rpc/main.py
```diff
@@ -5,9 +5,20 @@
 import sys
 import tempfile
 import threading
-import time
 import urllib.parse
 from typing import Any, Optional
+import sys
+import os
```
Moyasee, line 11: `import time` was removed from this import block, but `time.time()` is still used at lines 188, 205, 419, and 431. This throws `NameError: name 'time' is not defined` on every `action()` call and breaks downloading. Please re-add `import time`.

python_rpc/main.py
```diff
@@ -5,9 +5,20 @@
 import sys
 import tempfile
 import threading
-import time
 import urllib.parse
 from typing import Any, Optional
+import sys
```
Moyasee, line 10: Duplicate `import sys` (already imported on line 5). Please remove this one, and `import os` can move up into the main import block.

python_rpc/main.py
```diff
@@ -5,9 +5,20 @@
...
+import os
+
+if sys.platform == "win32":
```
Moyasee, line 13: What is this `add_dll_directory` block for? It silently swallows every exception, which hides real load failures. If it is needed for libtorrent on a specific setup, please add a comment explaining why; otherwise please remove it.

python_rpc/main.py
```diff
@@ -5,9 +5,20 @@
...
+    dll_dir = os.path.join(sys.base_prefix, "DLLs")
+    if os.path.isdir(dll_dir):
+        try:
+            os.add_dll_directory(dll_dir)
+        except Exception:
+            pass
+
+print(f"USING PYTHON: {sys.executable}", file=sys.stderr)
```
Moyasee, line 21: Leftover debug `print(...)`. Please remove.

src/main/events/misc/add-windows-defender-exclusion.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+};
+
+const addWindowsDefenderExclusion = async (
+  _event: Electron.IpcMainInvokeEvent,
+  exclusionPath: string
+) => {
+  if (!exclusionPath?.trim()) throw new Error("Invalid path.");
+  return runDefenderCommand(
+    `Add-MpPreference -ExclusionPath '${exclusionPath}'`
```
Moyasee, line 27: This interpolates the path into an elevated PowerShell command inside single quotes. Windows folder names can legally contain an apostrophe (for example `C:\Users\O'Brien\...`), which breaks the quoting, and since this runs elevated via `sudo-prompt`, that is a command injection risk. Please escape the path (double up `'` to `''`) or pass it as an argument / via `-EncodedCommand` instead of raw string interpolation. Same applies to lines 37 and 48.

src/main/services/game-files-manager.ts
```diff
@@ -304,6 +321,24 @@ export class GameFilesManager {
         WindowManager.sendToAppWindows("on-library-batch-complete");
 
         await this.createDesktopShortcutForGame(game.title);
+
+        if (isSetup) {
+          logger.info(
+            `[GameFilesManager] Setup detected, launching installer: ${foundExePath}`
+          );
+          try {
+            spawn(foundExePath, [], {
```
Moyasee, line 330: This auto-launches a detected setup executable from an extracted repack with no user confirmation. Silently running a downloaded installer is a security and UX concern; it should be user initiated. Also, on non-Windows this will try to `spawn` a `.exe` directly. Can we gate this behind an explicit user action?

src/main/services/game-files-manager.ts
```diff
@@ -634,6 +669,119 @@ export class GameFilesManager {
...
+      return 50;
+    if (normalizedExeName === "game" || normalizedExeName.includes("game"))
+      return 10;
+    return 0;
+  }
+
+  private async autoDetectExecutable(
+    folderPath: string,
+    gameTitle: string
+  ): Promise<{ exePath: string; isSetup: boolean } | null> {
```
Moyasee, line 690: The scoring and fallback heuristic (substring match, then a `game` keyword, then shortest filename) can pick the wrong executable fairly easily. Worth testing against a few real repacks, and logging the chosen candidate so mistakes are debuggable.

src/main/services/download/download-manager.ts
```diff
@@ -576,6 +576,9 @@ export class DownloadManager {
       if (shouldPauseSeedingForExtraction) {
         await this.cancelDownload(gameId);
 
+        // Wait for libtorrent to fully release file handles
+        await new Promise((resolve) => setTimeout(resolve, 2000));
```
Moyasee, line 580: A fixed `setTimeout(2000)` to wait for libtorrent to release file handles is racy: too short on slow disks, wasted time otherwise. Can we wait on an actual signal (session abort confirmed, handle released) instead of a magic delay? This block is also duplicated at line 591; worth extracting.

src/main/events/library/verify-game-integrity.ts
```diff
@@ -0,0 +1,106 @@
...
+) => {
+  const gameKey = levelKeys.game(shop, objectId);
+  logger.info(`[VerifyIntegrity] Called for ${gameKey}`);
+
+  let download = await downloadsSublevel.get(gameKey).catch(() => null);
+  const game = await gamesSublevel.get(gameKey).catch(() => null);
+
+  if (!game) {
+    logger.error(`[VerifyIntegrity] Game not found in DB for ${gameKey}`);
+    throw new Error("Jeu introuvable dans la base de données.");
```
Moyasee, line 25: Thrown error messages here are hardcoded French. Backend errors should be English, and ideally surfaced through i18n keys on the renderer. Same for the other `throw new Error(...)` strings in this file (lines 72, 76, 97, 103).

src/main/events/library/verify-game-integrity.ts
```diff
@@ -0,0 +1,106 @@
...
+    const uri = fallbackUri;
+    const downloadPath = fallbackDownloadPath;
+
+    if (uri && downloadPath) {
+      logger.info(
+        `[VerifyIntegrity] No download entry found, creating from fallback uri for ${gameKey}`
+      );
+
+      const newDownload: Download = {
+        shop: shop as any,
```
Moyasee, line 39: `shop: shop as any`: `shop` is already a `GameShop`, so the cast should not be needed. If the `Download` type does not accept it, let us fix the type rather than cast.

src/main/events/library/verify-game-integrity.ts
```diff
@@ -0,0 +1,106 @@
...
+        objectId,
+        uri,
+        downloadPath,
+        folderName: null,
+        progress: 0,
+        downloader: Downloader.Torrent,
```
Moyasee, line 45: This always sets `downloader: Downloader.Torrent` when fabricating a download entry. If repair runs on a direct (non-torrent) game with no existing download entry, this creates a bogus torrent entry and runs `force_recheck` on it. The synthesized downloader should reflect the actual source.

Moyasee (commented):
Thanks for the big effort here! This PR bundles around 5 independent features (executable auto-detect, setup auto-launch, Windows Defender exclusions, the extraction race fix, and the verify-integrity / force_recheck system). Could you split these into separate PRs? It would make each one reviewable and mergeable on its own.

Two blockers regardless of splitting:

1. `python_rpc/main.py` removed `import time`, but `time.time()` is still called at lines 188, 205, 419, and 431, so every `action()` call will raise `NameError: name 'time' is not defined` and break downloading. Please re-add `import time`.
2. User-facing strings are hardcoded French (thrown errors and i18n fallbacks). Hydra is English-first, so these need English text and locale keys. Several keys are also missing from `src/locales/en/translation.json`: `repair`, `verify_integrity`, `verify_integrity_success`, `verify_integrity_error`, `no_download_path`, `no_download_path_selected`, `add_windows_defender_exclusion`, plus the remove / update / success variants.

Inline notes on the rest.

src/renderer/src/hooks/use-windows-defender-exclusion.ts
```diff
@@ -0,0 +1,117 @@
...
+  const isWindows = window.electron.platform === "win32";
+
+  const handleAddExclusion = useCallback(async () => {
+    try {
+      await window.electron.addWindowsDefenderExclusion(downloadsPath);
+      onPreferenceChange({ hasWindowsDefenderExclusion: true });
+      showSuccessToast(
+        t(
+          "windows_defender_exclusion_success",
+          "Exclusion Windows Defender ajoutée avec succès !"
```
Moyasee, line 34: All toast and label fallback strings in this hook are French (lines 34, 42, 55, 63, 94, 110, 114). They need English defaults, and the keys (`add_windows_defender_exclusion`, `windows_defender_exclusion_success`, and so on) added to `src/locales/en/translation.json`; they do not currently exist.

src/renderer/src/pages/game-details/hero/hero-panel-actions.tsx
```diff
@@ -102,6 +103,44 @@ export function HeroPanelActions() {
...
 
+  const handleVerifyIntegrity = async () => {
+    if (!game) return;
+    try {
+      const uri = game.download?.uri || repacks[0]?.uris[0];
+      const downloadPath = game.download?.downloadPath;
+
+      if (!uri)
+        throw new Error(
+          t("no_repacks_found", "Aucune source trouvée pour ce jeu.")
```
Moyasee, line 114: French fallback strings and missing locale keys (`verify_integrity_success`, `verify_integrity_error`) here, also at lines 120, 133, 139. Please use English defaults and add the keys to the en locale.

src/renderer/src/pages/game-details/hero/hero-panel-actions.tsx
```diff
@@ -260,6 +299,21 @@ export function HeroPanelActions() {
       );
     }
 
+    // Show a "Repair" button when the game has a download folder but no executable yet
+    if (game?.download?.downloadPath) {
```
Moyasee, line 303: This returns the Repair button whenever `download.downloadPath` exists. Please confirm it does not shadow the normal play / install button for games that are already installed, and check the ordering of this return against the existing action logic. Also, this `handleVerifyIntegrity` is duplicated almost verbatim in `game-options-modal.tsx`; consider extracting a shared hook.

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
```diff
@@ -121,6 +124,48 @@ export function GameOptionsModal({
...
+  const handleVerifyIntegrity = async () => {
+    try {
+      setVerifyingIntegrity(true);
+      // Pass uri + downloadPath as fallback so the handler can create a DB entry if missing
+      const uri = game.download?.uri || repacks[0]?.uris[0];
+      const downloadPath = customDownloadPath || game.download?.downloadPath;
+
+      if (!uri)
+        throw new Error(
+          t("no_repacks_found", "Aucune source trouvée pour ce jeu.")
```
Moyasee, line 138: Same French fallbacks and missing keys as the hero panel (lines 138, 144, 157, 163). This handler is also copy-pasted from `hero-panel-actions.tsx`; please dedupe.

src/renderer/src/pages/game-details/modals/game-options-modal/general-section.tsx
```diff
@@ -214,6 +222,89 @@ export function GeneralSettingsSection({
...
 
+          {game.download?.downloadPath || customDownloadPath ? (
+            <>
+              <h4
+                className="game-options-modal__header-description"
+                style={{ marginBottom: 8, marginTop: 16 }}
+              >
+                {t(
+                  "download_path",
+                  "Dossier de téléchargement (utilisé pour la vérification)"
```
Moyasee, line 233: These two ternary branches (roughly lines 225 to 305) are nearly identical; the only real difference is the value shown and whether the Browse button renders. Please collapse into a single block. The fallback text here is also French, and the keys `no_download_path_selected` (line 291) and `verify_integrity` (line 354) are missing from the en locale.

Hachi-R:
i appreciate the effort here, but i don't think we should merge this as-is.

this pr currently mixes several independent features: setup auto-launch, executable auto-detection, windows defender exclusions, extraction race handling, verify integrity / force_recheck, and cleanup on game removal. this makes it hard to review and also increases the risk of regressions.

because of that, i'll close this pr for now. the setup.exe support can be revisited later as a smaller focused pr.

PR #2246, opened 2026-05-20
Feature/lbx 708

Hachi-R:

**Fill in the PR content:**


chubbygrannychaser (changes requested):
## Summary
This is a substantial Big Picture/settings/game-page expansion, and a lot of the structure looks promising, but I found several regressions that I think should be fixed before merge.

## Blocking issues
- Download-directory normalization can silently drop the currently selected custom default and reset the effective downloads path on a later settings save.
- Big Picture settings don't implement the new replacement flow when the saved custom directory list is full.
- Returning from the sidebar to the description breaks for long descriptions because the 50% visibility threshold can become impossible to satisfy.
- Gamepad repeat ownership is not restored when a temporary owner on the same lane is released.
- The download modal can overwrite in-progress user selections when `userPreferences` resolves/updates while it's open.
- The new Big Picture settings/game UI introduces a broad i18n regression with many hardcoded English strings and no locale updates in this PR.

I left inline comments on the main spots.

src/shared/download-directories.ts
```diff
@@ -0,0 +1,491 @@
...
+  return normalizedRecords
+    .filter((record) => {
+      if (seenPaths.has(record.path)) {
+        return false;
+      }
+
+      seenPaths.add(record.path);
+      return true;
+    })
+    .slice(0, MAX_OPTIONAL_DOWNLOAD_DIRECTORIES)
```
chubbygrannychaser: Because normalization slices to the newest 4 entries before validating `downloadsPath`, an older-but-still-selected custom default can be dropped here. On the next `updateUserPreferences()` call that makes `persistedDefaultPath` fall back to the OS default, so an unrelated settings save can silently reset the user's download location. I think the selected default needs to be preserved during trimming (or re-inserted before resolving the final default).

src/big-picture/src/pages/game/game.tsx
```diff
@@ -100,6 +504,435 @@ export default function Game() {
...
+      isMainContentFocusId(lastMainContentFocusId) &&
+      lastMainContentFocusId !== GAME_DESCRIPTION_BODY_ID
+    ) {
+      return getItemFocusTarget(lastMainContentFocusId);
+    }
+
+    if (
+      lastMainContentFocusId === GAME_DESCRIPTION_BODY_ID &&
+      isMainContentFocusId(lastMainContentFocusId) &&
+      isDescriptionVisibleEnoughForReturn()
```
chubbygrannychaser, line 598: This return path looks fragile for long descriptions. `isDescriptionVisibleEnoughForReturn()` requires at least 50% of the description block to be visible, which becomes impossible once the description is taller than ~2 viewport heights. In that case moving right into the sidebar and then left won't return the user to the description they were reading.

src/big-picture/src/services/gamepad.service.ts
```diff
@@ -322,44 +344,174 @@ export class GamepadService {
...
+
+    if (!owners) return;
+
+    const currentOwner = owners.get(lane);
+
+    if (!currentOwner || !this.isSameInputDescriptor(currentOwner, input)) {
+      return;
+    }
+
+    owners.delete(lane);
```
chubbygrannychaser: I think repeat ownership can get stuck here. If one held horizontal/vertical input owns repeat, a second input on the same lane claims ownership, and then the second input is released, we delete the current owner but never restore the first still-held input. The user then has to release and press again to resume repeating.

src/big-picture/src/components/modals/download-game/index.tsx
```diff
@@ -725,12 +711,13 @@ function DownloadGameOptions({
 
       if (cancelled) return;
 
-      setUserPreferences(preferences);
-      setSelectedDownloadPath(initialDownloadPath);
+      setSelectedDownloadPath(resolvedDirectories.defaultPath);
```
chubbygrannychaser: This effect now depends on `userPreferences`, so it will re-run whenever preferences finish loading or update while the modal is open. That means an in-progress user choice (selected directory / extraction toggles) can get overwritten by the latest persisted defaults after the modal is already visible. It would be safer to only initialize this state on open (or only while the user hasn't interacted yet).

src/big-picture/src/pages/settings/settings.tsx
```diff
@@ -1,29 +1,256 @@
...
+  CONTENT_ITEM_FOCUS_IDS,
+  DOWNLOADS_BEHAVIOR_ITEM_FOCUS_IDS,
+  DOWNLOAD_DIRECTORIES_DEFAULT_SELECT_ID,
+  getIntegrationProviderCheckboxFocusId,
+  NOTIFICATIONS_LIBRARY_ITEM_FOCUS_IDS,
+} from "./settings-navigation";
+
+import "./page.scss";
+
+const ALL_SETTINGS_TABS = [
```
chubbygrannychaser, line 29: This whole tab list is hardcoded in English, and I don't see any corresponding locale changes in this PR. The same pattern shows up across the new Big Picture settings/game UI (`Language Selector`, `Search language`, `Compatibility Tier`, `Steam Deck`, etc.), so this looks like a broad i18n regression rather than a one-off string.

chubbygrannychaser (commented):
Follow-up re-review on the latest head (`578187a`).

Good news: a few of the earlier blockers look fixed now:
- the selected default download directory is preserved during normalization
- the download modal no longer appears to clobber in-progress user choices when preferences refresh
- the gamepad repeat-owner handoff logic looks corrected

The blockers I still see are:
- **Big Picture full-list directory management still lacks the replace flow** when the saved custom directory list is full (`src/big-picture/src/pages/settings/download-directories-section.tsx`)
- **The broad i18n regression still stands** — the touched Big Picture UI continues to add hardcoded English strings without locale wiring (for example in game hero/achievements/download modal files)

Because the head changed and some issues were fixed, I’m leaving this as a follow-up comment rather than repeating the original full review. Once those remaining blockers are addressed, I’m happy to check again.

chubbygrannychaser (approved):
LGTM

PR #2262, opened 2026-05-26
feat:add TorBox web download support

Vidanox:

**Fill in the PR content:**

Allow TorBox to handle direct file hoster links by adding web download methods to TorBoxClient and corresponding types. Update getDownloadersForUri to return TorBox for supported hosts and support multiple downloaders per URI.

Torbox supported filehosts:
- Vikingfile
- Pixeldrain
- Gofile
- 1fichier
- mediafire


Moyasee:
Thanks for this! The multi-downloader refactor in `getDownloadersForUri` is clean and the consumers handle it fine. Verified against TorBox's official API though, and there are a few things to fix before it'll actually work end-to-end:

1. Possible string vs number ID mismatch (please verify)

 it looks like `createwebdownload` returns `webdownload_id` as a string, while webdl/mylist returns each item's id as a number. If that's accurate for the real responses, this strict-equality check in `getWebDownloadInfo` would never match:

```
  return webDownloads.find((item) => item.id === id) ?? null;
  //                          number === string → always false
```

  That would make `getWebDownloadInfo` always return null, which makes the file-resolution block effectively dead - files stays empty, `zipLink` is always false, no `file_id` is ever sent, and the name always falls back to the generic download name. Multi-file downloads (gofile/mediafire folders) would break, and single-file ones would only work if TorBox defaults to the lone file when no `file_id` is given.

  Could you confirm with a real `createwebdownload` + `mylist` response whether the IDs come back as the same type? If they do differ, a safe fix is to coerce:
  `return webDownloads.find((item) => String(item.id) === String(id)) ?? null;`
  (and type `webdownload_id` as string — the web_id/id fallbacks don't appear to be real fields in the response).

  2. Mediafire host won't match

  `torboxHosts` has `"https://mediafire.com"`, but real mediafire URIs (and the existing Mediafire matcher right above) use `"https://www.mediafire.com"`. `startsWith` fails, so TorBox is never offered for mediafire. Change it to `"https://www.mediafire.com"`.

  3. Gofile missing
  
  The description lists Gofile as supported, but `https://gofile.io` isn't in `torboxHosts`, so TorBox never gets offered for gofile links. Either add it or drop gofile from the description.

  4. Minor: vikingfile trailing slash

  `"https://vikingfile.com/"` has a trailing slash unlike the other entries — drop it for consistency (a bare `https://vikingfile.com` URI won't match otherwise).

  One more question: for non-cached hosts, is the web download guaranteed ready by the time `requestdl` is called? Might be worth checking download_state/cached before requesting the link so we don't request too early.

Moyasee:
Thanks for the updates, the ID type changes and the readiness polling look good. One thing though: I think the merge from `main` got tangled and broke `getDownloadersForUri`, wanted to flag it before it goes further.

`main` was using the old style where each host does an early `return [Downloader.X]`, but in this PR you refactored it to collect everything into the `downloaders` array with `.push()`. When the branches merged, the two styles got mixed and one leftover line slipped through:

```js
  if (
    uri.startsWith("https://buzzheavier.com") ||
    uri.startsWith("https://bzzhr.co") ||
    uri.startsWith("https://fuckingfast.net")
  ) {
    downloaders.push(Downloader.Buzzheavier);
  }
    return [Downloader.Mediafire];   // leftover from main's early-return style
  if (uri.startsWith("https://fuckingfast.co")) {
    downloaders.push(Downloader.FuckingFast);
  }
```

That `return [Downloader.Mediafire];` isn't inside any `if`, so it runs for every URI that reaches it and returns early. The effect is that almost everything — magnet links, TorBox/RealDebrid hosts, vikingfile, rootz, fuckingfast — ends up returning `[Mediafire]`, and everything below that line never runs. That's also what SonarQube is flagging with the unreachable code warning (S1763, L189).

The fix is just deleting that one line. Everything below it is already correct, so once it's gone the `fuckingfast.co` check becomes reachable again and the dedup plus magnet fallback at the bottom work as intended.

Might be worth a quick once-over of the rest of this file too, in case the merge left anything else half-resolved.

Hachi-R:
Closing this because the branch history is no longer compatible with the current `main`. GitHub reports this PR as `DIRTY`/`CONFLICTING` and shows 4,776 commits across 1,117 files, so the intended change can no longer be reviewed safely. Feel free to reopen this PR after rebuilding the branch or cherry-picking the intended commits onto the current `main`.

PR #2268, opened 2026-05-28
Feature/lbx 726

Hachi-R:
i love painting my nails 💅


src/big-picture/src/pages/catalogue/catalogue.tsx
```diff
@@ -1,521 +1,154 @@
...
+  );
 
-  const restoreChips = () => {
-    setChips([
-      { label: "Library", color: "#8aeb13" },
-      { label: "Cloud", color: "#67e8f9" },
-      { label: "Beta", color: "#f3c611" },
-    ]);
-  };
+  if (!catalogueData) return null;
```
chubbygrannychaser: This makes the whole catalogue render a blank page until all metadata exists, and it also hides the grid's own loading/error state when metadata fails. `catalogueData` stays undefined if any of the metadata requests errors, so users get no error UI and no search results even though `search.isError`/`search.error` are populated. Please render the page/grid shell independently of metadata and show a localized loading/error/degraded-filter state instead of returning `null`.

src/big-picture/src/pages/catalogue/use-catalogue-data.ts
```diff
@@ -0,0 +1,464 @@
...
+
+const externalResourcesInstance = axios.create({
+  baseURL: import.meta.env.RENDERER_VITE_EXTERNAL_RESOURCES_URL,
+});
+
+function parseParam<T>(value: string | null): T | undefined {
+  if (!value) return undefined;
+
+  try {
+    return JSON.parse(value) as T;
```
chubbygrannychaser: Please validate parsed query params before casting them to the expected type. For example `/catalogue?tags=1` parses to the number `1`, then `header.tsx` later calls `tags?.map(...)` and crashes because this is not an array. Array filters should only accept arrays (and filter/coerce item types); otherwise fall back to `[]`.

src/big-picture/src/pages/catalogue/use-catalogue-data.ts
```diff
@@ -0,0 +1,464 @@
...
+
+    const loadMetadata = async () => {
+      try {
+        const [
+          genresResponse,
+          tagsResponse,
+          developersResponse,
+          publishersResponse,
+          rawDownloadSources,
+        ] = await Promise.all([
```
chubbygrannychaser: One optional metadata failure currently rejects the entire batch and leaves `catalogueData` undefined, which makes the catalogue unusable because the page returns `null`. Please load these independently or use `Promise.allSettled` so the search results can still render while only the failed filter list is disabled/empty. Local download sources also should not be coupled to remote Steam metadata availability.

src/big-picture/src/pages/catalogue/header.tsx
```diff
@@ -0,0 +1,259 @@
...
+      className="catalogue-header"
+    >
+      <div className="catalogue-header__summary">
+        <div className="catalogue-header__search-term">
+          {hasActiveFilters ? (
+            <Typography
+              variant="label"
+              className="catalogue-header__search-term-text"
+            >
+              Showing search results for {title && <q>{title}</q>}
```
chubbygrannychaser, line 175: This PR adds a lot of new Big Picture UI copy as hardcoded English strings (catalogue header/filter labels, pagination text, settings labels, toast messages, virtual keyboard labels, etc.). Project rules require user-facing strings to go through i18next. Please move the new copy into the locale files and use translation keys/interpolation instead of literals/template text.

src/big-picture/src/components/common/toast/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+            <p className="big-picture-toast__description">{message}</p>
+          ) : null}
+        </div>
+
+        {action ? (
+          <Button
+            variant="secondary"
+            size="medium"
+            className="big-picture-toast__action"
+            focusable={false}
```
chubbygrannychaser, line 184: This action button is interactive but intentionally removed from both spatial focus and tab order, so keyboard/controller users cannot reach the action that sighted users can click. Either integrate toast actions into the Big Picture focus/navigation model or do not expose action controls in these toasts. The close button below has the same issue with `tabIndex={-1}`.

src/big-picture/src/layout/header/index.tsx
```diff
@@ -50,28 +171,114 @@ function Header() {
...
+      const isVirtualKeyboardClick = Boolean(
+        target?.closest(".virtual-keyboard")
+      );
+
+      if (isVirtualKeyboardClick && isSearchVirtualKeyboardTarget) {
+        return;
+      }
+
+      if (isSearchOpen && !searchRef.current?.contains(e.target as Node)) {
+        dismissSearch();
```
chubbygrannychaser: When clicking outside the search while the virtual keyboard is open, this closes the search UI but does not close the virtual keyboard. `handleFocusOut` in the keyboard only clears `suppressedTargetRef`, so the keyboard can remain open targeting a blurred/hidden input until the interval notices the target is gone. Please dismiss the keyboard here too (for example `dismissSearch({ closeKeyboard: true })` when the search input is the active virtual-keyboard target).

chubbygrannychaser (changes requested):
Found a few issues that should be addressed before merging. See inline comments.

chubbygrannychaser:
Thanks for the update — I re-checked the latest head. The catalogue blank-page/error handling, malformed URL param crash, metadata partial-failure handling, and search/virtual-keyboard dismissal issues look addressed now.

I still think this needs changes before merge because the earlier product-readiness blockers remain:

- New Big Picture user-facing copy is still hardcoded in English instead of going through i18next, e.g. catalogue header/filter text, sort labels, toast copy, settings labels, virtual keyboard labels/hints, etc.
- Toast action/close controls are still interactive but removed from keyboard/spatial focus (`focusable={false}` / `tabIndex={-1}`), so keyboard/controller users cannot reach them.

CI is green on the latest commit.

PR #2271, opened 2026-05-29
fix: use absolute path for ldconfig detection

### LINKED ISSUE #2270: GameMode detection fails because `ldconfig` is not in PATH
I found a false negative in the GameMode detection logic when running the AppImage on Debian 13.

The current implementation uses:

```ts
spawnSync("ldconfig", ["-p"])
```

This assumes `ldconfig` is available in the process `PATH`.

However, when launching through AppImage, `/usr/sbin` is not present in `PATH`, even though GameMode is correctly installed and working.

Example:

```bash
which ldconfig
# /usr/sbin/ldconfig
```

## Suggested fix

Use an absolute path for `ldconfig`, such as:

```bash
/usr/sbin/ldconfig
```

MarlonLuan:

**Fill in the PR content:**

closes #2270


Moyasee (changes requested):
This probably fixes the SteamOS/Linux PATH issue if ldconfig is located at `/usr/sbin/ldconfig`, but it may break or still fail on distros where ldconfig is under /sbin. Safer to check multiple known paths or resolve it dynamically.

I would update it to be something like this just to be extra sure

```
const ldconfigPaths = [
  "/usr/sbin/ldconfig",
  "/sbin/ldconfig",
  "ldconfig",
];
```

Moyasee (approved):
LGTM, thank you

chubbygrannychaser (approved):
LGTM

PR #2296, opened 2026-06-08
feat(catalogue): show 5 page buttons for multi-page jumps

Moyasee:

**Fill in the PR content:**
Widen the catalogue pagination window from 3 to 5 visible pages so a single click can advance up to 3 pages forward without typing a page number. Replace the isLastThree special-case with showLeadingJump/ showTrailingJump derived from actual hidden-page existence, which also prevents the last-page button from rendering twice at the wider width. Applies to both the desktop and big-picture catalogues.
image


chubbygrannychaser (approved):
LGTM

Issue #2261, opened 2026-05-26
[BUG] Hydra Launcher Still Captures Controller Input While Another App or Xbox Fullscreen Mode Has Focus


moi952:
When using Hydra Launcher in Big Picture mode, controller input is still captured by Hydra even after another application gains focus.

For example, if I launch a game, my controller continues to control Hydra in the background. This can cause unintended behavior such as launching the same game multiple times while already playing.

The same issue happens when using the Xbox fullscreen/navigation experience on Windows 11. If the Xbox interface or another application is focused, controller navigation inputs are still intercepted by Hydra instead of being sent exclusively to the active application.

https://github.com/user-attachments/assets/f245bb09-dca1-4c39-b60b-25913ba7d656

Moyasee:
Hey, should've been fixed in the most recent udpate. Feel free to create another issue if you still experience problems with that.

PR #2302, opened 2026-06-09
Feature/lbx 767

Hachi-R:
392 lines 

SIIIIIX SEVEEEEE


Hachi-R:
may fix #2261

Moyasee:
@greptile review pls

chubbygrannychaser (changes requested):
Thanks for the Big Picture work. I found a few blocking localization/accessibility issues that should be fixed before this lands:

- Several new Big Picture surfaces hardcode user-facing English strings instead of using i18next. Examples include the disc selection modal (`src/big-picture/src/components/modals/disc-selection/index.tsx`: title/description/region labels/actions), launch error toasts and confirmation copy (`src/big-picture/src/pages/game/game.tsx`), notification actions/header/empty state (`src/big-picture/src/layout/sidebar/notifications-dropdown.tsx`), the profile page labels/loading/empty states (`src/big-picture/src/pages/profile/profile.tsx`), and emulation settings toasts/actions. The repo rules require all user-facing UI copy, placeholders, aria labels, and toast text to be translated.
- The notifications dropdown can open with no controller/gamepad focus target when the list is empty: `initialFocusId` becomes `undefined`, while the visible close button is a raw button outside the spatial focus model and the empty state is not focusable. Big Picture modals need an initial reachable target for controller-first navigation.
- Notification rows currently render a focusable `div role="button"` containing another interactive menu `<button>`. That nested-interactive structure is fragile for keyboard/screen-reader semantics; split the row/menu into sibling controls or make the container non-interactive.

CI note: GitHub build/lint/text-integrity checks are green, but SonarCloud is currently failing on this PR.

chubbygrannychaser (approved):
LGTM

PR #2305, opened 2026-06-10
fix: update Content Security Policy and add registerSchemesAsPrivileged

zamitto:

**Fill in the PR content:**

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -46,6 +46,20 @@ i18n.init({
 
 const PROTOCOL = "hydralauncher";
 
+// Register the custom schemes as privileged so the renderer can fetch them
+// (supportFetchAPI) and use the results on a canvas without tainting it
+// (corsEnabled). Must run before the app is ready.
+protocol.registerSchemesAsPrivileged([
+  {
+    scheme: "local",
+    privileges: { supportFetchAPI: true, corsEnabled: true, stream: true },
+  },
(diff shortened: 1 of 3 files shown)


src/main/index.ts
```diff
@@ -46,6 +46,20 @@ i18n.init({
...
+// Register the custom schemes as privileged so the renderer can fetch them
+// (supportFetchAPI) and use the results on a canvas without tainting it
+// (corsEnabled). Must run before the app is ready.
+protocol.registerSchemesAsPrivileged([
+  {
+    scheme: "local",
+    privileges: { supportFetchAPI: true, corsEnabled: true, stream: true },
+  },
+  {
+    scheme: "gradient",
```
chubbygrannychaser, line 58: Since this PR registers `gradient` with `supportFetchAPI`/CORS, the renderer CSP also needs to allow `gradient:` (in both `src/renderer/index.html` and `src/big-picture/index.html`, like `local:`). The current `default-src` only adds `blob:`; the `*` source does not cover custom schemes, so `fetch('gradient:...')` or image loads for this scheme will still be blocked by CSP.

chubbygrannychaser (changes requested):
Found one blocking issue: the newly privileged `gradient:` scheme still needs to be allowed by the renderer CSP before it can be loaded/fetched from the renderer.

PR #2312, opened 2026-06-12
refactor: migrate friend request update logic to direct API call via …

Hachi-R:
yobwoc robas


chubbygrannychaser (approved):
LGTM. Code looks good; note that the branch currently has merge conflicts with main that need to be resolved before it can merge.

PR #2313, opened 2026-06-12
refactor: single reply button on parent review in vote line

Moyasee:
Move the reply action so it always sits inline with the upvote/downvote buttons on the parent review and no longer jumps down to the hide-replies row once replies load. Drop the duplicate reply button from the thread actions and remove the reply button from child replies.

**Fill in the PR content:**


chubbygrannychaser (approved):
LGTM

PR #2323, opened 2026-06-13
[fix] sidebar-section__header: improve interaction area

Wiwok:

**Fill in the PR content:**
This fixes the sidebar header button interaction area, allowing categories to be expanded and collapsed by clicking the entire header, not just the text inside it.


src/renderer/src/pages/game-details/sidebar-section/sidebar-section.tsx
```diff
@@ -30,32 +29,32 @@ export function SidebarSection({
...
             className={`sidebar-section__chevron ${
               isOpen ? "sidebar-section__chevron--open" : ""
             }`}
           />
           <span>{title}</span>
-        </button>
+        </span>
 
         {subtitle && subtitleHref && (
           <a
```
chubbygrannychaser, line 50: This `<a>` is now nested inside the outer `<button>`, which is invalid HTML because interactive content cannot be a descendant of a button. `stopPropagation()` prevents the mouse click from toggling, but it does not fix the invalid DOM/accessibility semantics and can still cause browser/React nesting warnings or odd keyboard behavior when `subtitleHref` is used. Please keep the link as a sibling of the toggle button (or make only the non-link header area a button) instead of nesting it inside the button.

chubbygrannychaser (changes requested):
Found one component regression in the latest update: the subtitle link is now rendered inside a native button, which creates invalid nested interactive markup. The wider header click target is good, but the link needs to remain outside the button/toggle control.

src/renderer/src/pages/game-details/sidebar-section/sidebar-section.tsx
Wiwok: Fixed it!

I couldn't find any place where `subtitle` is currently used, so I wasn't entirely sure how it was expected to look. I've kept the behavior/styling the same as before for now.

chubbygrannychaser (approved):
LGTM

PR #2332, opened 2026-06-15
feat: support more linux executable extensions in file pickers

Moyasee:

**Fill in the PR content:**
Add Linux-specific filters to the custom game modal and the game-details executable picker. On Linux, offer AppImage, sh, x86_64, x86, run and bin, plus an All files fallback so extensionless ELF binaries and odd-case names stay selectable. Windows and macOS filters are unchanged.


src/renderer/src/context/game-details/game-details.context.tsx
```diff
@@ -414,16 +414,27 @@ export function GameDetailsContextProvider({
   const selectGameExecutable = async () => {
     const downloadsPath = await getDownloadsPath();
 
+    const filters =
+      window.electron.platform === "linux"
+        ? [
+            {
+              name: "Game executable",
```
chubbygrannychaser: This label is user-facing and is shown in the native file picker, but it is hardcoded in English. Please add/use a translation key here (and for the non-Linux executable filter below) instead of `"Game executable"`, matching the repo i18n rule.

chubbygrannychaser (changes requested):
Found one i18n issue in the file picker labels.

Moyasee:
@greptile review pls

Moyasee:
@greptile review again

Moyasee:
@greptile review again

src/main/events/library/close-game.ts
```diff
@@ -28,17 +30,45 @@ const closeGame = async (
...
+  const targetPaths =
+    game.executablePath && !isWindowsBatchFile(game.executablePath)
+      ? [game.executablePath, ...trackingPaths]
+      : trackingPaths;
+
+  const gameProcesses = processes.filter((runningProcess) => {
+    const matchesTargetPath = targetPaths.some((targetPath) => {
+      if (process.platform === "linux") {
+        return (
+          runningProcess.name === targetPath.split("/").at(-1) ||
```
chubbygrannychaser: This basename-only match is now used inside `filter()` and also applies to every tracking path, so closing one game can kill every running process with the same filename. That is risky for common Linux launchers like `run.sh`, `start.sh`, `Game.AppImage`, or helper tracking executables that may exist in unrelated directories. Please require path-qualified evidence here (`exe === targetPath`, `APPIMAGE === targetPath`, or a cwd/dirname match with the same kind of guard used by `processReferencesExecutable`) before adding the process to the kill list.

src/main/helpers/launch-game.ts
```diff
@@ -44,13 +66,41 @@ const launchNatively = (
...
+        },
+      }
+    );
+
+    processRef.unref();
+
+    return processRef.pid ?? null;
   }
 
   const processRef = spawn(
```
chubbygrannychaser, line 110: With this PR Linux native launches now go through `spawn()` for `.sh`, `.run`, `.bin`, AppImages, wrapper commands, etc., but the returned ChildProcess has no `error` listener. If the selected file has a bad shebang, missing interpreter, wrong architecture, or cannot be executed, Node emits an unhandled `error` event in the main process, which can crash Hydra instead of showing a launch failure. Please attach an `error` listener before `unref()` (ideally resolve from `spawn`/`error` and only store the pid after `spawn`).

chubbygrannychaser (changes requested):
Found two remaining Linux launch/close regressions that should be tightened before merge.

chubbygrannychaser (approved):
LGTM. The previous close-game basename matching and spawn error-listener issues look fixed, SonarCloud quality gate is green now, and local node/web typechecks pass.

PR #2333, opened 2026-06-15
feat: support .cso disc images and fix PS1 disc serial extraction

Moyasee:

**Fill in the PR content:**
Add a CISO v1 reader (cso-reader + cso-decode) and route .cso through SKU extraction in extract-disc-sku, so compressed PS2 images get matched to metadata on import. Malformed or unsupported (v2/ZSO) files fail cleanly and land in the unmatched list.

Fix the disc serial regex that only matched the PS2 3-2 digit split; widen it to also handle the PS1 2-3 split (e.g. SLUS_01.569 -> SLUS-01569) so PS1 CHD, bin, iso and cso images match. Move the scan into a sku-scan helper shared by the chd and cso paths.


chubbygrannychaser (approved):
LGTM

PR #2344, opened 2026-06-16
fix(download): wait for file handles release in python_rpc and download manager

TeALO36:
Addresses extraction race conditions:
- Fixed python_rpc/main.py import issue (re-added import time).
- Replaced magic delay/setTimeout in download-manager.ts with robust EBUSY retry/polling logic to wait for file handles to be released.


Hachi-R (changes requested):
thanks for the pr.

the file handle retry before extraction makes sense to me, but the force_recheck part seems unrelated to this fix and does not seem to be used by this flow.

could you please remove the force_recheck changes from this pr and keep this focused only on the extraction/file handle issue?

also, please move the retry count/delay to named constants and extract the retry logic to a helper function, since sonar is reporting the download completion function as too complex after these changes.

Moyasee:
Can you please fix the merge conflicts, address the findings from Greptile, and fix the Sonar findings? Greptile is at 3/5 with 3 open comments and Sonar reports 1 new issue.

Hachi-R:
Closing this because the branch history is no longer compatible with the current `main`. GitHub reports this PR as `DIRTY`/`CONFLICTING` and shows 4,774 commits across 1,117 files, so the intended change can no longer be reviewed safely. Feel free to reopen this PR after rebuilding the branch or cherry-picking the intended commits onto the current `main`.

PR #2345, opened 2026-06-16
feat(library): implement game integrity verification hook

TeALO36:
Implements the game integrity verification hook:
- Created use-verify-game-integrity.ts hook.
- Removed hardcoded French user-facing strings and migrated them to proper i18n English-first locale keys.
- Updated src/locales/en/translation.json with missing keys.


Hachi-R:
thanks for the pr.

the feature is presented as verify integrity / repair, but currently it only seems to do a real recheck for torrents. for other downloaders, it only checks if the executable exists, which is not really an integrity verification.

i think this needs a more complete implementation before exposing it in the library, so i'll close this pr for now.

PR #2346, opened 2026-06-16
refactor(auto-detect-executable): simplify scoring logic and handle setups

TeALO36:
Improves executable auto-detection:
- Simplified the scoring logic for game executables (removed length-based heuristic).
- Explicitly leaves setup.exe unbound to avoid incorrect auto-bindings.


Hachi-R:
thanks for the pr.

i get the idea, but i don't think we should merge this right now.

scanning the game folder and guessing the executable can easily bind the wrong file, especially for games with multiple binaries.

for now, relying on the known executables list feels safer, so i'll close this pr.

PR #2347, opened 2026-06-16
feat(settings): add Windows Defender exclusion toggle with secure PowerShell execution

TeALO36:
Implements secure Windows Defender exclusion adding/removing:
- Encodes PowerShell commands in Base64 to prevent argument injection vulnerabilities.
- Added UI toggle in Settings, along with the English translations in src/locales/en/translation.json.

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -858,7 +858,15 @@
     "downloads": "Downloads",
     "create_shortcuts_on_download": "Create desktop and Start Menu shortcuts when a game finishes downloading",
     "default_proton_version": "Default Proton version",
-    "default_proton_version_description": "Select the Proton version used by Auto mode when a game has no custom override"
+    "default_proton_version_description": "Select the Proton version used by Auto mode when a game has no custom override",
+    "windows_defender_exclusion_success": "Windows Defender exclusion added successfully!",
+    "windows_defender_exclusion_error": "Error adding Windows Defender exclusion.",
+    "windows_defender_exclusion_removed_success": "Windows Defender exclusion removed successfully!",
+    "windows_defender_exclusion_removed_error": "Error removing Windows Defender exclusion.",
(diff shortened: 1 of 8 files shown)


chubbygrannychaser:
Hi @TeALO36, thank you for taking the time to put this PR together — the effort is genuinely appreciated!

Unfortunately, we're going to have to decline this one. Adding a Windows Defender exclusion toggle is a sensitive security feature that could be misused or perceived negatively by users and security tools alike. It's the kind of change that puts Hydra in a difficult position, so we'd prefer to keep it out of the project for now.

No hard feelings at all — we hope you'll continue contributing in other ways. Thanks again!

TeALO36:
Hi @chubbygrannychaser,

Thank you for the explanation! I completely understand and respect your decision. Security and keeping Hydra's reputation clean are absolutely paramount, so avoiding features that might trigger antivirus flags or look suspicious makes total sense.

I just wanted to clarify our original intention and raise a UX concern to get your thoughts. The goal of this PR was not to apply exclusions automatically or by default to all users, but rather to provide an optional, simple toggle for users who know the risks. 

From my own experience, I've seen many friends instantly delete/uninstall Hydra after a false positive because they think the launcher itself is malicious or broken. I often have to manually explain to people how to add exclusions in Windows Defender so their games actually launch without Defender silently deleting half of the game files. Providing a manual button in the settings is meant to help these users solve this issue easily.

Given that an automated toggle inside the app is out of the question for security reasons, do you think it would make sense to address this passively? For example:
- Displaying a simple warning or a link to a guide/documentation page explaining how to manually add an exclusion in Windows Defender when a game fails to launch or launch setup.
- Showing a helpful tooltip/troubleshooting section in the settings.

I'd love to hear how you think we should handle this type of user confusion, or if you prefer to keep this out of the launcher entirely. Thanks again for your time!

Issue #2330, opened 2026-06-15
Option to Disable on-screen keyboard


alexmagicpotions:
When using a Steam Deck, there is already an OS level on-screen keyboard that makes it impossible to enter text in Hydra due to its own onscreen keyboard. There should be a setting in Hydra to disable the built-in OSK so that it doesn't conflict with any external OSK the user may already have.

PR #2363, opened 2026-06-17
Feature/lbx 791

Hachi-R:
💃

        🦽🏌️‍♂️


Hachi-R:
closes #2330

src/big-picture/src/pages/game/game.tsx
```diff
@@ -366,35 +378,36 @@ export default function Game() {
...
     protonDBData,
     achievements,
+    preferredAssets,
     openGame,
     closeGame,
     toggleFavorite,
     updateGame,
+    refreshGameDetails,
   } = useGameDetails(objectId!, shop!);
+  const [automaticCloudSync, setAutomaticCloudSync] = useState(
```
chubbygrannychaser: `game` is usually still `null` on the first render while `useGameDetails` loads, so this state initializes to `false` and never syncs when the real game arrives (or when navigating to another game). The Cloud tab can therefore show automatic cloud sync as disabled even for a game that already has it enabled, and saving the modal can persist the stale value. Please sync this state from `game?.automaticCloudSync` in an effect, or reuse the modal state hook that already tracks game changes.

src/big-picture/src/components/pages/game/game-settings-modal/customization-tab.tsx
```diff
@@ -0,0 +1,288 @@
...
+  onClearAsset: (assetType: AssetTab) => Promise<void>;
+}
+
+function getAssetPreviewState(game: LibraryGame): AssetPreviewState {
+  const preferredAssets = resolvePreferredGameAssets(game, null);
+
+  return {
+    icon: {
+      src: preferredAssets.iconSrc,
+      hasCustom: Boolean(game.customIconUrl),
```
chubbygrannychaser: For custom games, the selected artwork is stored in `iconUrl`, `logoImageUrl`, and `libraryHeroImageUrl` (see the Big Picture handlers that call `updateCustomGame`). This check only looks at the non-custom `custom*Url` fields, so custom-game artwork is rendered as if it were not custom and the clear controls never become available. Please branch on `game.shop === "custom"` and use the custom-game fields for both `hasCustom` and the fallback preview.

src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.tsx
```diff
@@ -0,0 +1,344 @@
...
+      () => {
+        showSuccessToast("Cloud backup complete");
+        setCreatingBackup(false);
+        void loadArtifacts();
+        void loadBackupPreview();
+      }
+    );
+
+    const removeDownloadCompleteListener =
+      window.electron.onBackupDownloadComplete(game.objectId, game.shop, () => {
```
chubbygrannychaser: The main process sends `false` on restore failure for `on-backup-download-complete`, but the preload listener drops that boolean and this handler always shows `Cloud save restored`. In Big Picture, a failed restore therefore looks successful and refreshes the UI as if it completed. Please pass the success flag through `onBackupDownloadComplete` and show an error/keep the failed state when it is false.

src/big-picture/src/components/modals/download-game/index.tsx
```diff
@@ -265,72 +782,111 @@ export function DownloadGameModal({
...
           </motion.div>
-        </AnimatePresence>
+
+          {pendingStep !== null ? (
+            <div
+              ref={pendingStepMeasureRef}
+              className={`download-game-modal__step-measure download-game-modal__step download-game-modal__step--${pendingStepTransitionKey}`}
+              aria-hidden="true"
+            >
+              {renderStepContent(pendingStep)}
```
chubbygrannychaser: This hidden measurement copy mounts the full step content, including live `FocusItem`s and children that use `stealFocusOnAppear`. `aria-hidden` only hides it from accessibility APIs; it does not prevent the custom navigation service from registering/focusing those invisible controls during the transition. That can move controller focus into offscreen content while changing steps. Please make measurement renders inert to the navigation system (for example pass a measurement/navigation-disabled flag to suppress FocusItem registration/stealFocus) or measure a non-interactive shell instead of mounting the active controls.

chubbygrannychaser (changes requested):
Found a few Big Picture settings/download regressions that should be fixed before merge. Also note that SonarCloud is currently failing on this head.

src/big-picture/src/components/common/focus-item/index.tsx
```diff
@@ -80,6 +87,8 @@ export function FocusItem({
   }
 
   useEffect(() => {
+    if (isMeasurement) return;
```
chubbygrannychaser, line 90: This only skips `registerNavigationNode`; the measurement copy still runs the later effects that call `requestFocusWhenAvailable` for `stealFocusOnAppear` items and `registerItemActions` for the invisible controls. Because the measured tree uses the same focus IDs as the real step content, an offscreen measurement render can still queue/clear focus for those IDs or register duplicate actions while changing steps. Please make measurement `FocusItem`s fully inert to the navigation system (e.g. also skip the update/steal-focus/action-registration effects, or derive an `effectiveNavigationState` of `hidden` plus avoid action registration) rather than only skipping node registration.

chubbygrannychaser (dismissed):
Thanks for the fixes; the cloud sync state, custom artwork, and restore-result forwarding issues look addressed, and local web/node typechecks pass. One Big Picture measurement focus issue remains: the measurement render no longer registers navigation nodes, but it can still run focus-request and action-registration effects.

src/big-picture/src/hooks/queries/use-game-download-options.hook.ts
```diff
@@ -78,7 +164,20 @@ export function useGameDownloadOptions(
     return () => {
       cancelled = true;
     };
-  }, [game.objectId, game.shop, shouldLoadDownloadOptions]);
+  }, [
+    downloadSourcesDependencyKey,
+    game.downloadSources,
```
chubbygrannychaser: `game.downloadSources` is still in the effect dependencies alongside `downloadSourcesDependencyKey`, which is already the stable content key derived from the same array. If `downloadSources` is recreated with identical contents while assets/details load, this effect runs again, clears the option state, flips the loading flags, and repeats the fetch unnecessarily. Please depend on the stable key only so equivalent source lists do not reset/reload the modal.

chubbygrannychaser (changes requested):
Thanks for the updates. The previous measurement-render focus/action issue looks addressed now, and both local typechecks pass. I’m still holding approval because there is one remaining hook dependency issue and SonarCloud is failing the quality gate on new-code duplication (12.1%, limit <= 3%).

chubbygrannychaser (changes requested):
Thanks, the previous hook dependency issue is fixed in this head, and the earlier measurement/focus, cloud sync, custom artwork, and restore-result issues still look addressed. Local `typecheck:node` and `typecheck:web` both pass.

I’m still holding approval because SonarCloud is failing the quality gate for this head. Please address the new-code duplication failure and ping me again once the quality gate is green.

chubbygrannychaser (changes requested):
Thanks for the updates. The previous hook dependency issue is fixed in this head, and the earlier measurement/focus, cloud sync, custom artwork, and restore-result issues still look addressed. Local `typecheck:node` and `typecheck:web` both pass.

I’m still holding approval because SonarCloud is failing the quality gate for this head: new-code duplication is 5.9% with a required maximum of 3%. Please address that quality-gate failure and ping me again once SonarCloud is green.

chubbygrannychaser (changes requested):
Thanks for the updates. I re-checked the latest head: the previous hook dependency issue is still fixed, and the earlier measurement/focus, cloud sync, custom artwork, and restore-result issues still look addressed. Local `typecheck:node` and `typecheck:web` both pass.

I’m still holding approval because SonarCloud is failing the quality gate for this head: new-code duplication is 7.3% with a required maximum of 3%. Please reduce the duplication and ping me again once SonarCloud is green.

src/big-picture/src/pages/settings/big-picture.tsx
```diff
@@ -0,0 +1,450 @@
...
+        userPreferences.bigPictureVirtualKeyboardEnabled ?? true,
+      bigPictureDiagnosticsEnabled:
+        userPreferences.bigPictureDiagnosticsEnabled ?? false,
+      bigPictureDiagnosticsPosition:
+        userPreferences.bigPictureDiagnosticsPosition ?? "bottom-center",
+    });
+  }, [userPreferences]);
+
+  const updateUserPreferences = async (values: Partial<BigPictureForm>) => {
+    const nextForm = { ...form, ...values };
```
chubbygrannychaser: `updateUserPreferences` merges with the `form` value captured by whichever memoized `onChange` callback is being invoked. Because each item array only depends on its own field, those callbacks can be stale for the other fields. For example, toggle sounds off, then toggle `launchInBigPicture`: the startup callback can still merge with the older form where sounds were `true`, so the local UI state reverts the sounds checkbox even though the persisted preference was updated. Please avoid merging against a stale closure here (for example use a functional `setForm((current) => ({ ...current, ...values }))`, or include a stable callback with the right dependencies).

chubbygrannychaser (changes requested):
Thanks for the Sonar fix — the quality gate is green now, and local `typecheck:node` / `typecheck:web` both pass. I found one Big Picture settings state regression that should be fixed before merge.

src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.tsx
```diff
@@ -0,0 +1,364 @@
...
+  }, [game.objectId, game.shop]);
+
+  useEffect(() => {
+    loadArtifacts().catch(() => {});
+    loadBackupPreview().catch(() => {});
+  }, [loadArtifacts, loadBackupPreview]);
+
+  useEffect(() => {
+    const removeUploadCompleteListener =
+      globalThis.window.electron.onUploadComplete(
```
chubbygrannychaser, line 412: Big Picture waits for these upload/download completion callbacks to clear `creatingBackup` / `restoringArtifactId`, but the main-process cloud-save code still sends `on-upload-complete-*`, `on-backup-download-progress-*`, and `on-backup-download-complete-*` only to `WindowManager.mainWindow`. When the action is triggered from the Big Picture window, this listener never fires, so backup/restore can stay stuck in a loading state and the user never sees failure/success feedback. Please broadcast these cloud events to app windows (or otherwise target Big Picture too) before relying on them here.

src/big-picture/src/components/common/tabs/index.tsx
```diff
@@ -256,9 +295,90 @@ export function Tabs<TValue extends string = string>({
...
+
+      return i >= 0 && i < resolvedItems.length ? i : -1;
+    },
+    [resolvedItems]
+  );
+
+  useEffect(() => {
+    if (variant === "settings") return;
+
+    const removeLeftBumper = onButtonPressed(
```
chubbygrannychaser, line 393: These bumper handlers are registered globally for every non-settings `Tabs` instance and only check whether the gamepad event is active. They do not check whether this tab list is focused/visible or in the active navigation layer. Since page tabs stay mounted underneath modals, pressing LB/RB inside a modal can also switch background tabs (and multiple mounted tabsets can react to the same press). Please gate the handler to the active/focused tab region/layer, or move bumper handling to the focused tab list.

chubbygrannychaser (changes requested):
Found a couple more Big Picture regressions that should be fixed before merging.

src/big-picture/src/pages/settings/big-picture.tsx
```diff
@@ -0,0 +1,471 @@
...
+    },
+    [updateUserPreferences]
+  );
+
+  const startupItems = useMemo<BigPictureItem[]>(() => {
+    return [
+      {
+        id: "launch-in-big-picture",
+        focusId: BIG_PICTURE_ITEM_FOCUS_IDS.launchInBigPicture,
+        label: "Launch Hydra in Big Picture",
```
chubbygrannychaser: This new settings section is rendering hardcoded English strings instead of using the Big Picture i18n namespace. The translation keys for these labels/section copy were added in `src/big-picture/src/locales/*/translation.json`, but this component never calls `useTranslation`, so non-English Big Picture users will still see English for labels like `Launch Hydra in Big Picture`, `Enable virtual keyboard`, `Enable Big Picture sounds`, `Enable diagnostics`, and `Diagnostics position`. Please route these labels and generated position labels through `t(...)`.

chubbygrannychaser (changes requested):
Thanks for the updates — the two regressions from my previous review look fixed now, and local typecheck/eslint/tests pass. I found one remaining project-rule issue before this is merge-ready.

chubbygrannychaser (approved):
Thanks for the updates. The previous Big Picture settings i18n blocker is fixed, and I don't see any remaining blocking issues in this head. Local checks passed: typecheck, ESLint on changed JS/TS files (warnings only), and npm test.

PR #2364, opened 2026-06-17
feat: emulator install helper and PS1/PS2/PS3

Moyasee:

**Fill in the PR content:**
SIX SEVEEEEEEEEEEEEEEE 

Add a guided install step to the emulator setup wizard and BIOS checks for DuckStation/PCSX2, mirroring the existing RPCS3 firmware flow.

- Resolve and download the latest installer per OS/arch from GitHub releases (DuckStation rolling, PCSX2 stable + pre-release, RPCS3 win .7z auto-extracts and sets the path, Linux AppImage reveals). AUR/Flatpak shown as links.
- Add official-website, install-from-GitHub (with a visit-release link) and a Work Wonders guide card to the download step.
- Detect PS1/PS2 BIOS on the wizard and block launch when missing, surfacing a clear toast instead of letting the emulator crash silently.
- Brand icons for GitHub/AUR/Flatpak, single-column compact cards, loading state, and a fix for the horizontal scrollbar at minimal modal height.
- New i18n keys across en/pt-BR/pt-PT/ru/es.


Moyasee:
@greptile review this again

src/main/services/emulators/emulator-install-sources.ts
```diff
@@ -0,0 +1,350 @@
...
+};
+
+const githubEntries = (
+  binary: EmulatorBinary,
+  os: InstallOs,
+  arch: InstallArch
+): EmulatorSourceEntry[] => {
+  if (binary === "duckstation") return duckstationEntries(os, arch);
+  if (binary === "pcsx2") return pcsx2Entries(os, arch);
+  return rpcs3Entries(os);
```
chubbygrannychaser: `binary` comes from IPC, so the TypeScript union does not protect this boundary at runtime. Any unexpected string that is not `duckstation` or `pcsx2` falls through to the RPCS3 entries here; on Windows `downloadAndInstallEmulator` then uses the original `binary` in `path.join(managedEmulatorsDir(), binary)` before extracting the archive. A renderer call such as `installEmulator("../../...", "rpcs3-install")` can therefore move extraction outside the managed emulator directory. Please validate `binary` against the known emulator list at the event/service boundary and reject unknown values before resolving options or building paths.

src/main/services/emulators/emulator-install-sources.ts
```diff
@@ -0,0 +1,350 @@
...
+      id: "duckstation-aur",
+      binary: "duckstation",
+      linkKind: "aur",
+      url: "https://aur.archlinux.org/packages/duckstation-git",
+    },
+  ];
+};
+
+const pcsx2AssetPattern = (os: InstallOs, arch: InstallArch): RegExp => {
+  if (os !== "win32") return /linux-appimage-x64-Qt\.AppImage$/i;
```
chubbygrannychaser, line 89: On Linux this ignores `arch` and always looks for the x64 AppImage. RPCS3 has the same issue because `rpcs3Entries` does not receive `arch` and always offers the `_linux64.AppImage`. On arm64 Linux the setup modal will present/download incompatible x64 builds. Please gate these GitHub install options by architecture: return an arm64 asset only when one exists, otherwise expose only the AUR/Flatpak/release-page link instead of an installable x64 binary.

chubbygrannychaser (changes requested):
Found a couple of install-helper issues around IPC input validation and Linux arm64 options.

chubbygrannychaser (approved):
Thanks for the updates. The IPC binary validation and Linux arm64 install-option gating issues I flagged are addressed, and local node/web typechecks pass.

PR #2370, opened 2026-06-18
fix: restore collection creation in library and reposition scan cancel button

Moyasee:
## What

Two small fixes bundled together.

### Restore collection creation on the library page
The sidebar performance rewrite (#2366) removed the only entry point that opened the create collection modal outside the game context menu, so there was no longer any way to create a collection from the library. The modal and the underlying `createCollection` flow were both untouched, just orphaned.

This adds a `+` button with a tooltip to the collections header on the library page that opens the existing `CreateCollectionModal`. Same auth gate as before: signed-out users get the sign-in window, signed-in users get the modal. On success the new collection is dispatched to the store and shows up in the chips immediately.

### Reposition the scan cancel button
In the emulator setup flow, while a ROM folder is being scanned, the cancel scan button sat on the right side of the footer right before Continue. Moved it to the left side so it no longer crowds the primary action.

## Testing
- `yarn typecheck:web` passes
- `yarn format` clean

### DIFF
--- a/src/renderer/src/pages/library/library.scss
+++ b/src/renderer/src/pages/library/library.scss
@@ -198,6 +198,46 @@
     }
   }
 
+  &__collections-section {
+    display: flex;
+    flex-direction: column;
+    gap: calc(globals.$spacing-unit);
+    width: 100%;
+  }
+
+  &__collections-header {
+    display: flex;
+    align-items: center;
+    gap: calc(globals.$spacing-unit);
+  }
+
+  &__collections-title {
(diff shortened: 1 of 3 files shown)


chubbygrannychaser (approved):
LGTM. I checked the restored collection creation flow and the setup footer change; the current `endAction` call site is mutually exclusive with Back/Cancel, translations are present, SonarCloud is green, and local `typecheck:node` / `typecheck:web` pass.

PR #2372, opened 2026-06-18
fix: emulator library, classics launch, memcard backups and cloud save connector

Moyasee:
Bundle of fixes around the emulator settings page reported by users.

## Stale games in the emulator library tab
Re-scanning a ROM folder only ever upserted the games it found, so a game whose files were moved or deleted kept its stale disc paths and stayed in the library tab no matter how many rescans. Now a scan reconciles afterwards and soft-deletes games whose disc files are no longer present in the scanned folders. It keys on disk presence rather than match success, so a transient SKU sniff miss cannot wipe a still-valid game.

## Classics launch with a moved or removed disc
Launch resolved a disc path but never checked the file still existed, so a moved or deleted ROM spawned the emulator into a broken state with no feedback. It now verifies the disc exists and surfaces NO_DISC otherwise. The coded launch errors were also thrown without any logging, which made user reports impossible to diagnose, so each one now logs its code and context.

## Concurrent memory card backups
Backup progress was a single active card plus a single progress value, so starting a second card's backup overwrote the first and the bar flickered between the two as their events interleaved. Progress is now keyed by card file path, so concurrent backups each render their own bar.

## Download sources modal spacing
With a single source above the divider the grid stretched its rows to fill the modal body, leaving a large gap before the install guide. Rows now pack to the top so spacing is consistent regardless of source count.

## Cloud save connector lines
The connector assumed a single row of cards, so once they wrapped to a second row the lower rows lost their lines. It now draws one bus per row with a central trunk linking them to the console, centers the cards under the console, and widens the row gap so each bus has room.

## Testing
- typecheck and lint pass

### DIFF
--- a/src/big-picture/src/pages/settings/emulation/cloud-saves-section.tsx
+++ b/src/big-picture/src/pages/settings/emulation/cloud-saves-section.tsx
@@ -13,13 +13,7 @@ import type {
   EmulatorConfig,
   MemcardRestoreTarget,
 } from "@types";
-import {
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useRef,
-  useState,
-} from "react";
+import { useCallback, useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 
 import {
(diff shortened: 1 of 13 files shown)


Moyasee:
@greptile review again

chubbygrannychaser (approved):
LGTM. I reviewed the emulator import/launch changes, per-card backup progress, shared cloud connector, and the renderer/big-picture call sites. Local typecheck:node, typecheck:web, and lint all pass (lint only reports existing warnings outside this PR), and GitHub checks are green.

PR #2376, opened 2026-06-19
Fix/add big picture missing locales and nav overrides 

Hachi-R:

**Fill in the PR content:**

### DIFF
--- a/src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.scss
+++ b/src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.scss
@@ -58,6 +58,7 @@
 
 .game-cloud-settings-tab__saves-shell {
   display: flex;
+  flex-direction: column;
   flex: 1;
   min-height: 0;
   min-width: 0;
@@ -71,14 +72,16 @@
   gap: calc(var(--spacing-unit) * 4);
   width: 100%;
   min-width: 0;
+  max-height: 320px;
+  overflow-y: auto;
 }
 
 .game-cloud-settings-tab__saves-viewport {
+  display: flex;
+  flex-direction: column;
   flex: 1;
   min-height: 0;
-  overflow-x: hidden;
-  overflow-y: auto;
-  scrollbar-width: none;
+  overflow: hidden;
 
   &::-webkit-scrollbar {
     display: none;
--- a/src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.scss
+++ b/src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.scss
@@ -173,7 +173,7 @@
   &__targets {
     list-style: none;
     margin: 0;
-    padding: 0;
+    padding: 4px;
     display: flex;
     flex-direction: column;
     gap: 8px;
--- a/src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.tsx
+++ b/src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.tsx
@@ -26,6 +26,11 @@ import {
 import { useBigPictureToast, useNavigation } from "../../../../hooks";
 import { useUserDetails } from "../../../../hooks/use-user-details.hook";
 import { SettingsSection } from "../../../../pages/settings/settings-section";
+import {
+  getEmulationCloudRestoreButtonNavigationOverrides,
+  getEmulationCloudRestoreTargetFocusId,
+  getEmulationCloudRestoreTargetNavigationOverrides,
+} from "../../../../pages/settings/settings-navigation";
 import { CloudSavesList } from "./cloud-saves-list";
 
 import "./cloud-tab.scss";
@@ -110,7 +115,7 @@ function EmulationRestoreModal({
     const frameId = globalThis.window.requestAnimationFrame(() => {
       setFocus(
         selectedTarget
-          ? `emu-saves-target-${selectedTarget}`
+          ? getEmulationCloudRestoreTargetFocusId(selectedTarget)
           : RESTORE_MODAL_PICK_BUTTON_ID
       );
     });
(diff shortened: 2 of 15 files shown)


chubbygrannychaser (dismissed):
Thanks for the Big Picture locale and navigation fixes. I’m holding approval for now because the current head still has a hook dependency regression in `LibraryGameContextMenu`: `onViewAchievements` is used by the memoized context menu items but is no longer listed in the dependency array, which ESLint also reports via `react-hooks/exhaustive-deps`. Please add it back so the menu cannot retain a stale achievements callback.

Also, the current SonarCloud quality gate is failing on new-code duplication (41.1%, required ≤ 3%). Local checks otherwise passed: typecheck, ESLint on changed files with warnings only, and npm test.

Moyasee:
Reviewed the latest head (`8a3a479b`). Status on the earlier feedback plus a few extra notes.

### Resolved since the reviews
- The `onViewAchievements` dependency regression that Greptile (P1) and @chubbygrannychaser flagged is fixed. The `useMemo` dep array in `game-context-menu.tsx` now lists both `onViewAchievements` and the new `t`, so there's no stale callback.
- Greptile's P2 about hardcoded hex in `renderer/.../emulation-save-modals.scss` is addressed. The focus-visible ring now uses `var(--primary)` and `color-mix(...)`, so it follows the theme. color-mix is fine for our Electron Chromium version.

### Still failing
- SonarCloud quality gate. It went from 41.1% down to 20.1% new-code duplication after the helper extraction, but it's still well over the 3% threshold so the gate stays red. The bulk of what's left is the restore-target markup, which is near-identical between `cloud-tab.tsx` (`EmulationRestoreModal`) and `cloud-saves-section.tsx` (`RestoreModal`): the `<FocusItem ... navigationOverrides={getEmulationCloudRestoreTargetNavigationOverrides({...})}>` block plus the two `<Button focusNavigationOverrides={getEmulationCloudRestoreButtonNavigationOverrides(selectedTarget)}>` instances. Pulling the nav logic into `settings-navigation.ts` helped, but the JSX wiring is still copy-pasted across both modals. A shared target-list/item component (or a small render helper) used by both would get the gate green.

### Other things worth a look
- `emulation/styles.scss`: this PR replaced `@include emulation-primary-focus-ring;` on the restore target with an inline `box-shadow` ring, and also added a second near-identical inline block at the top for `.emu-save-modal__target`. The same snippet now lives in two spots and steps outside the mixin pattern used in ~10 other places in this file. Worth extracting a sibling mixin (e.g. `emulation-target-focus-ring`) and `@include`-ing it in both. That also trims the Sonar duplication.
- Those two inline blocks overlap on the same element (`.emu-save-modal__target` `:focus-visible` and `[data-focus-visible="true"]`), so they're partly redundant.
- The renderer ring (`border-color` + soft `color-mix` glow) and the big-picture ring (`inset` + outer 1px hard ring) now render differently for the same restore target across the two layers. If they're meant to match, worth aligning them.
- `cloud-tab.tsx`: making the restore button focusId unique with `-${key}` is the right fix for the duplicate-ID bug. One thing to double-check: the cloud tab's content-entry target in `index.tsx` (`GAME_SETTINGS_TAB_FOCUS_IDS.hydra_cloud` = `GAME_CLOUD_SETTINGS_PRIMARY_CONTROL_ID`) is still the bare ID, which now only exists in the non-emulation branch. For emulation games no element carries the bare ID anymore, so entering the cloud tab leans on the nav layer's `preferRememberedFocus` fallback. Worth a quick gamepad check that focus still lands on the first restore button when opening the cloud tab on an emulation game.

Locale parity looks good. big-picture only ships en/es/fr/pt-BR/ru and `game_options` was added to all five.

Moyasee:
Re-reviewed the three new commits (head `0ba35774`). The big one is `4a1424d5`, which pulls the restore logic into a shared `EmulationCloudRestoreModal` so `cloud-tab.tsx` and `cloud-saves-section.tsx` both delegate to it instead of duplicating it.

### Main outcome
- SonarCloud quality gate now passes (0 new issues, duplication back under threshold). That was the blocker from the last round, so this clears it.

### Refactor looks correct
- Namespace is preserved: the shared component uses `useTranslation("settings")`, same as both originals, and the `cloud_restore_*` keys live in the shared `src/locales`. text-integrity and lint are green.
- Both call sites pass the full prop set (`save`/`platform`/`onClose`/`onRestored` plus the action IDs). The per-surface differences are kept: `cloud-saves-section` still passes `SETTINGS_TOAST_OPTIONS` and `modalClassName="emulation-settings__modal"`, and `cloud-tab` keeps its plain toasts.
- Nice side effect: `cloud-tab`'s picked-file label now goes through the shared `basename` helper instead of a `/`-only split, so Windows paths label correctly there too.

### Still open from the last review (non-blocking)
These weren't part of these commits, just flagging they're still there:
- `emulation/styles.scss` still has the inline `box-shadow` focus ring duplicated in two spots and steps outside the `emulation-primary-focus-ring` mixin used elsewhere in the file. A sibling mixin would tidy that.
- The renderer ring (`border` + soft glow) and the big-picture ring (`inset` + outer hard ring) still render differently for the same restore target.
- The cloud tab's entry-focus target (`GAME_CLOUD_SETTINGS_PRIMARY_CONTROL_ID`) is still the bare ID, which only exists in the non-emulation branch now. For emulation games entry leans on the nav layer's `preferRememberedFocus` fallback. Worth a quick gamepad check that focus still lands on the first restore button when opening the cloud tab on an emulation game.

### Minor
- The shared component sits in `pages/settings/emulation/` but `cloud-tab` reaches up four levels to import it. It matches the existing `settings-navigation` import so it's consistent, but a shared `components/` home might layer a bit cleaner. Optional.

Builds are still pending; otherwise this is in good shape.

Moyasee (approved):
Approving. The duplication blocker is resolved by the shared `EmulationCloudRestoreModal` extraction, the Sonar gate is green, and lint/text-integrity pass. The two earlier review concerns (`onViewAchievements` dependency and the new-code duplication) are both addressed, and the refactor reads as a faithful move with no behavior change at the call sites.

A few non-blocking follow-ups can be handled later: the inline `box-shadow` focus ring in `emulation/styles.scss` duplicates the snippet and could use a sibling mixin like `emulation-primary-focus-ring`, and it's worth a quick gamepad check that entering the cloud tab on an emulation game still focuses the first restore target now that the entry ID only exists in the non-emulation branch.

PR #2378, opened 2026-06-19
fix: detect emulator BIOS by content instead of file extension

Moyasee:

**Fill in the PR content:**

## What

BIOS detection in the emulator setup now validates files by their contents instead of their extension.

## Why

A user reported that his PS2 BIOS worked fine in PCSX2 but Hydra kept saying the BIOS was missing and refused to launch any game (`BIOS_NOT_CONFIGURED`). His BIOS folder contained a valid dump named `SCPH-90001_BIOS_V18_USA_230.ROM0`, with no `.bin`/`.rom` file at all.

The scan only accepted files ending in `.bin` or `.rom`. PCSX2 and DuckStation ignore the extension entirely and read the actual file contents, so anything dumped as `.ROM0` (or any other naming) gets picked up by the emulator but was rejected by us.

## How

- Removed the `.bin`/`.rom` allowlist.
- Validate by content the way the emulators do: the PS2 romdir signature (`RESET`/`ROMDIR` records) and the PS1 copyright string, gated behind the existing size range as a cheap prefilter.
- Skip only directories and `stat` each entry rather than trusting the `readdir` dirent type, so symlinked BIOS files are no longer dropped before the check runs.

## Testing

Verified against the reported BIOS set:
- the `.ROM0`-only folder is now detected
- a folder containing only a random 4MB file is correctly rejected (content check, not just size)
- the auxiliary `.MEC`/`.NVM`/`.DIFF`/`.INF` files are skipped
- existing `.bin` dumps and PS1 BIOS still detected

### DIFF
--- a/src/main/services/emulators/bios-detection.ts
+++ b/src/main/services/emulators/bios-detection.ts
@@ -103,7 +103,7 @@ export const resolvePs2BiosDirs = async (
   return Array.from(new Set(dirs)).filter((dir) => existsSync(dir));
 };
 
-const PS1_BIOS_SIGNATURE = Buffer.from("Sony Computer Entertainment Inc.");
+const PS1_BIOS_SIGNATURE = Buffer.from("Sony Computer Entertainment");
 const PS2_ROMDIR_RESET = Buffer.from("RESET\0");
 const PS2_ROMDIR_ROMVER = Buffer.from("ROMVER");
 const PS2_ROMDIR_ENTRY_BYTES = 16;
--- a/src/main/services/emulators/bios-detection.ts
+++ b/src/main/services/emulators/bios-detection.ts
@@ -105,32 +105,10 @@ export const resolvePs2BiosDirs = async (
 };
 
 const PS1_BIOS_SIGNATURE = Buffer.from("Sony Computer Entertainment");
-const PS2_ROMDIR_RESET = Buffer.from("RESET\0");
-const PS2_ROMDIR_ROMVER = Buffer.from("ROMVER");
-const PS2_ROMDIR_ENTRY_BYTES = 16;
-const PS2_ROMDIR_NAME_BYTES = 10;
+const PS2_RESET = Buffer.from("RESET");
+const PS2_ROMVER = Buffer.from("ROMVER");
 const PS2_HEADER_BYTES = 0x80000;
 const PS1_HEADER_BYTES = 0x10000;
-
-const ps2RomdirHasRomver = (head: Buffer): boolean => {
-  for (
-    let start = head.indexOf(PS2_ROMDIR_RESET);
-    start >= 0;
-    start = head.indexOf(PS2_ROMDIR_RESET, start + 1)
-  ) {
-    for (
-      let off = start;
-      off + PS2_ROMDIR_ENTRY_BYTES <= head.length;
-      off += PS2_ROMDIR_ENTRY_BYTES
-    ) {
-      const name = head.subarray(off, off + PS2_ROMDIR_NAME_BYTES);
-      const end = name.indexOf(0);
-      if (end <= 0) break;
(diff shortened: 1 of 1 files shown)


src/main/services/emulators/bios-detection.ts
```diff
@@ -104,10 +103,54 @@ export const resolvePs2BiosDirs = async (
   return Array.from(new Set(dirs)).filter((dir) => existsSync(dir));
 };
 
+const PS1_BIOS_SIGNATURE = Buffer.from("Sony Computer Entertainment Inc.");
```
Hachi-R: remove inc, ps1 bios files can have Sony Computer Entertainment Inc, Sony Computer Entertainment Europe or Sony Computer Entertainment America

`
const PS1_BIOS_SIGNATURE = Buffer.from("Sony Computer Entertainment");
`

with 

```
if (system === "ps1") {
  return head.includes(PS1_BIOS_SIGNATURE);
}
```

can cover all cases

src/main/services/emulators/bios-detection.ts
```diff
@@ -104,10 +103,54 @@ export const resolvePs2BiosDirs = async (
...
+const PS2_ROMDIR_RESET = Buffer.from("RESET\0");
+const PS2_ROMDIR_NEXT = Buffer.from("ROMDIR");
+const PS2_ROMDIR_ENTRY_BYTES = 16;
```
Hachi-R: change this lines 
```
const PS2_ROMDIR_RESET = Buffer.from("RESET\0");
const PS2_ROMDIR_NEXT = Buffer.from("ROMDIR");
```

for 
```
const PS2_RESET = Buffer.from("RESET");
const PS2_ROMVER = Buffer.from("ROMVER");
```

that is what PCSX2 uses to identify valid ps2 bios files

src/main/services/emulators/bios-detection.ts
```diff
@@ -104,10 +103,54 @@ export const resolvePs2BiosDirs = async (
...
+const PS2_HEADER_BYTES = 0x40000;
```
Hachi-R: use `const PS2_HEADER_BYTES = 0x80000;`

rpcs 2 looks for the first 512kb of the file 

```
for (u32 i = 0; i < 512 * 1024; i++)
{
    if (std::fread(&rd, sizeof(rd), 1, fp) != 1)
        return false;

    if (std::strncmp(rd.fileName, "RESET", sizeof(rd.fileName)) == 0)
        break;
}
```

src/main/services/emulators/bios-detection.ts
```diff
@@ -104,10 +103,54 @@ export const resolvePs2BiosDirs = async (
...
+    ) {
+      const next = i + PS2_ROMDIR_ENTRY_BYTES;
+      if (
+        head
+          .subarray(next, next + PS2_ROMDIR_NEXT.length)
+          .equals(PS2_ROMDIR_NEXT)
+      )
+        return true;
+    }
+    return false;
```
Hachi-R: remove 
```
for (
  let i = head.indexOf(PS2_ROMDIR_RESET);
  i >= 0;
  i = head.indexOf(PS2_ROMDIR_RESET, i + 1)
) {
  const next = i + PS2_ROMDIR_ENTRY_BYTES;
  if (
    head
      .subarray(next, next + PS2_ROMDIR_NEXT.length)
      .equals(PS2_ROMDIR_NEXT)
  )
    return true;
}
return false;
```

and use 

`return head.includes(PS2_RESET) && head.includes(PS2_ROMVER);`

pcsx2 checks for a romver entry to identify valid ps2 bios files, so checking for reset and romver is closer to its actual detection logic and doesnt depend on entry order

Moyasee:
@greptile review again

Hachi-R (changes requested):

Hachi-R:
@greptile

Hachi-R (approved):
LGTM

PR #2385, opened 2026-06-21
fix: update notification unread count locally on read and dismiss

Moyasee:

**Fill in the PR content:**

Closes LBX-808.

The sidebar notification badge stopped going down when you marked notifications as read. The API unread count is only fetched once at login (the old 5 minute poll was removed), and the `notificationsChanged` handler only recomputed the local count, so the cached API count stayed stale until the next websocket push.

This updates the cached API unread count locally instead of refetching:

- The `notificationsChanged` event now carries a small detail payload (`apiUnreadDelta` or `resetApiUnread`).
- Marking a single unread item as read, or dismissing one, decrements the cached count by one.
- Mark all as read and clear all reset the API portion to zero.
- Local notifications are unchanged since the sidebar already re-reads them from the local DB.

The websocket `on-sync-notification-count` push still overwrites the cached value with the authoritative server count, so any local drift heals on the next push. No extra call to `/profile/notifications/count` is made per action.

Testing:
- Mark a single API notification as read, badge drops by one.
- Mark all as read and clear all, badge API portion drops to zero.
- Dismiss an unread notification, badge drops by one.


Moyasee:
@greptile hey boy review again

Hachi-R (approved):
LGTM, thanks for fixing this.

I reviewed the current flow and didn't find any blocker. The local unread count update is nicely scoped: single notification actions decrement only when the item was actually unread, bulk actions reset the API portion, and the websocket count can still overwrite the local value later if anything drifts.

PR #2386, opened 2026-06-21
fix: restore .exe selection in Linux executable pickers

Moyasee:
Fixes LBX-813.

**Fill in the PR content:**

The recent change that added native Linux executable formats swapped the whole filter list instead of extending it, so `.exe`, `.lnk`, `.bat` and `.cmd` got dropped on Linux. Most games on Linux run through Wine/Proton and ship as `.exe`, so the file picker greyed them out, the executable path could never be set, and playtime tracking never kicked in.

Changes:
- Restore the Wine-runnable extensions next to the native Linux ones in the game executable and custom-game pickers.
- Make the tracking-executable picker and the Big Picture executable picker platform-aware, so native Linux binaries (`.sh`, `.AppImage`, `.x86_64`, etc.) can be selected there too instead of only `.exe`.
- Pull the per-OS game executable extension lists into `@shared` so the Linux list spreads the Windows one. Adding a Windows extension now carries over to Linux automatically and the arrays are no longer duplicated across the game details context, game options and Big Picture pickers.

All the game executable pickers now share the same source of truth: on Linux the filter lists both the Wine-runnable and native formats with an All files fallback, macOS lists `.app`, and Windows lists `.exe`/`.lnk`/`.bat`/`.cmd`. The custom-game picker keeps its own broader set since it also accepts installer formats (`.msi`, `.deb`, `.rpm`, `.dmg`).

Verified `yarn format`, `yarn lint` and `yarn typecheck` pass.

### DIFF
--- a/src/big-picture/src/components/pages/game/game-settings-modal/use-game-settings-modal-state.ts
+++ b/src/big-picture/src/components/pages/game/game-settings-modal/use-game-settings-modal-state.ts
@@ -3,6 +3,10 @@ import type { ChangeEvent } from "react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { platformToSystem } from "@renderer/helpers";
+import {
+  LINUX_GAME_EXECUTABLE_EXTENSIONS,
+  WINDOWS_GAME_EXECUTABLE_EXTENSIONS,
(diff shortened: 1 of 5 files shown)


Moyasee:
@greptile review again

Hachi-R (approved):
LGTM

PR #2389, opened 2026-06-22
Feature/lbx 817

Hachi-R:


src/locales/es/translation.json
```diff
@@ -1149,7 +1149,16 @@
     "setup_step_label_done": "Listo",
     "remove_emulator": "Quitar emulador",
     "remove_emulator_title": "¿Quitar {{name}} de Hydra?",
-    "remove_emulator_description": "Hydra olvidará la ruta de {{name}} y todas las carpetas de ROMs mapeadas. El emulador y tus archivos ROM permanecen en el disco. Puedes volver a ejecutar la configuración después."
+    "remove_emulator_description": "Hydra olvidará la ruta de {{name}} y todas las carpetas de ROMs mapeadas. El emulador y tus archivos ROM permanecen en el disco. Puedes volver a ejecutar la configuración después.",
+    "big_picture": "Big Picture",
+    "big_picture_startup": "Inicio",
```
Moyasee, line 1154: ES is missing `launch_hydra_in_big_picture`, which this panel uses for the Startup checkbox label (`t("launch_hydra_in_big_picture")`). Every other locale has it, so Spanish users will see the raw key. Could you add it here, something like:

```json
"launch_hydra_in_big_picture": "Iniciar Hydra en modo Big Picture (experimental)",
```

Moyasee (commented):
Nice clean mirror of the Big Picture in-mode settings into the desktop settings window. The new panel follows the existing `settings-context-*` convention (redux selector + local form + `updateUserPreferences`), the category id is registered in both the union and the `isSettingsCategoryId` guard, and the tab/routing wiring looks right. Removing the duplicate `remove` key in the settings namespace is safe since another one remains.

One thing to fix before merge, plus a couple of non-blocking notes below.

- The Spanish locale is the only one missing `launch_hydra_in_big_picture`. This panel is the first place that key is actually consumed in the main renderer, so the Startup checkbox will render the raw key string for ES users. The other seven locales already have it.

Non-blocking:
- Greptile flagged a missing `useMemo` on `diagnosticsPositionOptions`, but the current code already memoizes it, so nothing to do there.
- The in-mode `BigPictureSettingsSection` shows a description under each section heading (`settings_*_section_description`); this panel drops them. Fine if that is intentional for the desktop style, just calling it out for parity.

Moyasee (approved):
looks good to me

PR #2390, opened 2026-06-22
Feature/lbx 793

Hachi-R:


Hachi-R:
@greptile

src/main/events/misc/file-system.ts
```diff
@@ -0,0 +1,137 @@
...
+  isFile: boolean;
+}
+
+function assertTrustedSender(event: Electron.IpcMainInvokeEvent): void {
+  if (!event.senderFrame) {
+    throw new Error("Unauthorized IPC sender");
+  }
+
+  const url = event.senderFrame.url;
+  if (!url.startsWith("app://") && !url.startsWith("file://")) {
```
Moyasee: This allowlist does not match how the app loads the renderer. Nothing is served from `app://` (only `local` and `gradient` schemes are registered in src/main/index.ts). In dev the window loads `http://localhost` via `ELECTRON_RENDERER_URL`, and in production it loads the remote `https://release-...` launcher subdomain (see `WindowManager.loadWindowURL`). Only the `file://` fallback passes here, so `readDirectory` / `getPathInfo` / `listDrives` will throw "Unauthorized IPC sender" in dev and in remote-loaded prod builds, which breaks the picker.

Suggest allowing the origins the app actually uses (localhost in dev + the launcher subdomain), or validating the sender against a known window rather than the URL scheme. Heads up on the flip side: once the remote https origin is allowed, these handlers expose full filesystem enumeration to remotely hosted UI, so it is worth being deliberate about that.

src/big-picture/src/components/common/file-explorer-modal/use-file-explorer.ts
```diff
@@ -0,0 +1,296 @@
...
+import { type DirectoryEntry } from "../../../helpers";
+import {
+  getParentPath,
+  matchesFilters,
+  normalizeFilters,
+  type FileFilter,
+} from "./utils";
+
+const SKELETON_COUNT = 6;
+const PATH_INPUT_PLACEHOLDER = "Select a location";
```
Moyasee: These strings are hardcoded English (same for `DRIVES_LABEL`, the `EMPTY_*` titles, and the `ERROR_MESSAGES` map below). The surrounding tabs already use `t()`, so these will not be translated. Same goes for "Select this directory" in index.tsx and `title="Select Download Directory"` in download-directories-section.tsx. Could you move them into the locale files? New EN keys should also land in pt-BR/pt-PT/ru/es per the usual convention.

src/renderer/src/declaration.d.ts
```diff
@@ -932,8 +950,8 @@ declare global {
     cancelGameTransfer: (shop: GameShop, objectId: string) => Promise<void>;
 
     /* Event listeners for transfer progress */
-    on: (channel: string, listener: (...args: any[]) => void) => void;
-    off: (channel: string, listener: (...args: any[]) => void) => void;
+    on: (channel: string, listener: (...args) => void) => void;
```
Moyasee, line 955: This change looks unrelated to the picker and does not really help: with `noImplicitAny` off, `(...args)` is still implicitly `any[]`, and preload still declares these as `(...args: any[])`. Suggest reverting these two lines to keep the explicit type and avoid the drift.

Moyasee (changes requested):
Nice work, the picker is well structured and the icon mapping is thorough. One blocking issue with the IPC sender check, plus an i18n gap and a few nits.

Blocking:
- The `assertTrustedSender` allowlist (`app://`/`file://`) does not match how the renderer is actually loaded, so the picker IPC will throw in dev and in remote-loaded prod builds. Details inline on file-system.ts.

Should fix:
- The new modal hardcodes user-facing English instead of using `t()`. Details inline on use-file-explorer.ts.

Nits (non-blocking):
- file-system.ts readDirectory fires a `stat` per file via `Promise.all`; on huge folders that can spike open handles. Consider bounding concurrency.
- The `DirectoryEntry` shape is now declared three times (main file-system.ts, big-picture helpers/file-entry-icon.tsx, and `FileExplorerEntry` in declaration.d.ts). Would be nice to share one type.
- file-entry-icon.tsx: `"mka"` is listed twice in `audioExtensions` (lines 107 and 119). Harmless dupe.
- utils.ts getParentPath normalizes `\` to `/`, so on Windows currentPath ends up shown/stored with forward slashes. Cosmetic.
- use-file-explorer.ts `showSelectThisDir` returns the currentPath string rather than a boolean; works via truthiness but `Boolean(...)` reads cleaner.

src/big-picture/src/components/common/file-explorer-modal/use-file-explorer.ts
```diff
@@ -0,0 +1,271 @@
...
+  const { t } = useTranslation("big_picture");
+  const [currentPath, setCurrentPath] = useState<string>("");
+  const [entries, setEntries] = useState<DirectoryEntry[]>([]);
+  const [drives, setDrives] = useState<string[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const generatedId = useId();
+  const fileListRegionId = `file-explorer-region-${generatedId.replaceAll(":", "")}`;
+  const pathInputPlaceholder = t("file_explorer_path_placeholder");
+  const drivesLabel = t("file_explorer_drives");
```
Moyasee, line 79: bug: none of the `file_explorer_*` keys exist in any translation.json. The new strings landed in the `exact` section as English-sentence keys, but `t()` with the `big_picture` namespace resolves from `format`, so the modal renders raw keys like `file_explorer_drives` in every language. The `exact` entries are dead weight since that map only matches literal DOM text. fr also didn't get the new strings.

src/big-picture/src/components/common/file-explorer-modal/use-file-explorer.ts
```diff
@@ -0,0 +1,271 @@
...
+    };
+  }, [
+    visible,
+    currentPath,
+    fileExplorerErrorFallback,
+    fileExplorerErrorMessages,
+  ]);
+
+  useEffect(() => {
+    if (!visible || currentPath) {
```
Moyasee: bug: this clears `drives` whenever `currentPath` is set, but `handleBPress`/`goToParent`/`hasParent` only return to the drive list when `drives.length > 0`, which can now only be true while `currentPath` is already empty. Once the start path resolves the drive list is unreachable, and on Windows pressing B at a drive root closes the modal instead, so there's no way to switch drives. Keeping the loaded drives around and only clearing them on close would fix it.

src/big-picture/src/components/common/file-explorer-modal/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+export function getEntryMeta(entry: DirectoryEntry): string {
+  if (!entry.isFile) return "";
+  return formatBytes(entry.size);
+}
+
+export function normalizeFilters(filters?: FileFilter[]): Set<string> | null {
+  if (!filters || filters.length === 0) return null;
+
+  const allExtensions = filters.flatMap((f) => f.extensions);
+  const specificExtensions = allExtensions.filter((ext) => ext !== "*");
```
Moyasee: bug: flattening all filter groups drops the all-files group whenever another group has specific extensions, and files with no extension (`extension === ""`) never match the set. On Linux the exec picker passes `getGameExecutableFilters` plus all files, so extension-less native binaries are hidden with no way to pick them; the native dialog let you switch to All files. Same story for unusual ROM formats in the disc picker.

src/main/events/misc/file-system.ts
```diff
@@ -0,0 +1,210 @@
...
+  const result = await mapWithConcurrency(
+    entries,
+    FILE_STAT_CONCURRENCY,
+    async (entry): Promise<DirectoryEntry> => {
+      const fullPath = join(dirPath, entry.name);
+      const name = entry.name;
+      const isMacApp =
+        entry.isDirectory() && name.toLowerCase().endsWith(".app");
+      const isDirectory = entry.isDirectory() && !isMacApp;
+      const isFile = entry.isFile() || isMacApp;
```
Moyasee: bug: `Dirent.isFile()` and `isDirectory()` are both false for symlinks, so a symlinked directory comes back as `{isDirectory: false, isFile: false}`: invisible in directory pickers and treated as a file elsewhere. That's common on Linux/Steam Deck (symlinked Steam libraries, home dirs), and the old dialog followed symlinks. Needs a `stat` fallback when `entry.isSymbolicLink()`.

src/main/events/misc/file-system.ts
```diff
@@ -0,0 +1,210 @@
...
+  } catch {
+    return {
+      exists: false,
+      isDirectory: false,
+      isFile: false,
+    };
+  }
+};
+
+const listDrives = async (): Promise<string[]> => {
```
Moyasee: risk: the A-Z probe runs sequentially with no timeout, and `access()` on a stale mapped network drive or empty optical drive can stall for seconds per letter. `Promise.all` over the 26 probes avoids the worst case. Also `listDrives` skips the `assertTrustedSender` check its two siblings do; harmless data, but easier to keep the gate consistent.

src/main/events/misc/file-system.ts
```diff
@@ -0,0 +1,210 @@
+import { app } from "electron";
+import { access, readdir, stat } from "node:fs/promises";
+import { join } from "node:path";
+import { platform } from "node:os";
+import { registerEvent } from "../register-event";
+
+const FILE_STAT_CONCURRENCY = 32;
+
+interface DirectoryEntry {
```
Moyasee, line 9: nit: `DirectoryEntry` is defined here, again in `file-entry-icon.tsx`, and a third time as `FileExplorerEntry` in `declaration.d.ts` (same for `PathInfo`). Shared IPC payload types live in `@types` (e.g. `DiskUsage`), so one definition there would keep the copies from drifting.

src/big-picture/src/helpers/file-entry-icon.tsx
```diff
@@ -0,0 +1,579 @@
...
+  extension: string;
+  size: number;
+}
+
+const iconProps = {
+  size: 22,
+  weight: "fill",
+} as const;
+
+export const imageExtensions = new Set([
```
Moyasee: nit: all nine extension sets are exported but nothing outside this file imports them; only `getEntryIcon` is consumed. Dropping the exports makes it clear they're just the icon lookup table.

src/renderer/src/declaration.d.ts
```diff
@@ -934,8 +952,8 @@ declare global {
...
```
Moyasee, line 955: nit: this `on`/`off` change from `(...args: any[])` to `(...args)` looks accidental; it's unrelated to the PR and just trades an explicit `any[]` for an implicit one. Worth reverting.

Moyasee (commented):
The gamepad file explorer is a solid replacement for the native dialogs, but there are a few real issues to sort out before merge: the modal's translation keys don't resolve (raw `file_explorer_*` strings render in every language), the Windows drive list is unreachable, and filter/symlink handling regresses vs `showOpenDialog` on Linux. Details inline.

src/big-picture/src/components/common/file-explorer-modal/use-file-explorer.ts
```diff
@@ -0,0 +1,273 @@
...
+    fileExplorerErrorMessages,
+  ]);
+
+  useEffect(() => {
+    if (!visible) {
+      setDrives([]);
+      return;
+    }
+
+    if (currentPath || drives.length > 0) return;
```
Moyasee: bug: the drive list still ends up empty in most sessions. On first open this races: `loadDrives` only fires while `currentPath` is empty, and when `resolveStartPath` lands it changes the deps, so the cleanup sets `cancelled = true` and an in-flight `listDrives` result is thrown away. On every reopen it's deterministic: `currentPath` keeps the previous session's path (it's never reset on close, and this hook lives outside the Modal's `visible &&` unmount), so this guard always returns early and drives stay `[]`. Either way, B at a drive root falls through to `onClose()` again. Loading drives keyed on `visible` alone, without the `currentPath` guard and cancellation, avoids both.

src/big-picture/src/components/common/file-explorer-modal/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+  filters: NormalizedFilters | null,
+  directoryOnly: boolean
+): boolean {
+  if (directoryOnly && !entry.isDirectory) return false;
+  if (entry.isDirectory) return true;
+  if (!filters) return true;
+  if (filters.specificExtensions.size === 0) return filters.hasWildcard;
+  if (filters.specificExtensions.has(entry.extension)) return true;
+
+  return filters.hasWildcard && entry.extension === "";
```
Moyasee: q: wildcard now only admits extension-less files, so a ROM with an extension that's missing from `getEmulatorRomExtensions` stays hidden even though the disc picker passes an `all_files` group. With the native dialog the user could switch to All files and pick it. Intentional trade-off, or should the wildcard admit everything?

Moyasee (commented):
Thanks for the fixes. The i18n keys, symlink handling, extension-less files, drive scan and dead exports all check out now. One issue from the first pass survives in a different form (drive list still ends up unreachable, see inline), plus a question on the wildcard semantics. The `DirectoryEntry` triplication and the `(...args)` edit in `declaration.d.ts` from the first review are still there, but no big deal.

Moyasee (approved):
lgtm

src/big-picture/src/components/common/file-explorer-modal/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    >
+                      <FolderIcon size={22} weight="fill" />
+                      <span>{drive}</span>
+                    </button>
+                  </FocusItem>
+                ))}
+              </>
+            )}
+
+            {vm.filteredEntries.length === 0 &&
```
Moyasee, line 155: bug: on every open there's a window where this renders "This folder is empty" before anything has loaded. Initial state is `currentPath=""`, `isLoading=false`, and neither `resolveStartPath` nor the empty-path branch of the load effect sets `isLoading`, so until the start-path IPC resolves this condition is true. Initializing `isLoading` to true (or gating on `currentPath`) would cover it.

src/big-picture/src/components/common/file-explorer-modal/use-file-explorer.ts
```diff
@@ -0,0 +1,273 @@
...
+  }, [visible, initialPath]);
+
+  useEffect(() => {
+    if (!visible) {
+      setEntries([]);
+      setDrives([]);
+      return;
+    }
+
+    if (!currentPath) {
```
Moyasee, line 128: risk: this branch doesn't reset `isLoading`, and the in-flight load's `finally` is skipped when cancelled. So pressing B at a drive root while a load is in flight leaves `isLoading` latched true: the skeleton shows forever and the drive list (gated behind `!isLoading`) never appears. Today it's masked by the drives never loading, but it'll surface once that's fixed, so worth handling in the same change.

src/big-picture/src/components/pages/game/game-settings-modal/compatibility-tab.tsx
```diff
@@ -250,10 +251,8 @@ export function GameCompatibilitySettingsTab({
             <Button
               focusId={GAME_COMPATIBILITY_SETTINGS_WINE_SELECT_ID}
               variant="secondary"
-              icon={<FolderOpen size={16} />}
-              onClick={() => {
-                handleSelectWinePrefix().catch(() => {});
-              }}
+              icon={<FolderOpenIcon size={16} />}
+              onClick={handleSelectWinePrefix}
```
Moyasee: bug: the old call site wrapped this in `.catch(() => {})`; now a rejection from `getDefaultWinePrefixSelectionPath()` is an unhandled rejection and the picker silently never opens. Every sibling handler in this file still guards itself (`handleClearWinePrefix().catch(...)` etc.), so this one looks like an accidental drop. `handleWinePrefixPicked` has the same gap on the `onSelect` side, since the modal calls it fire-and-forget.

src/big-picture/src/components/common/file-explorer-modal/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+              !vm.showDriveList &&
+              !vm.isLoading && (
+                <EmptyState
+                  className="file-explorer__empty"
+                  icon={<FolderOpenIcon size={32} weight="fill" />}
+                  title={vm.emptyTitle}
+                />
+              )}
+
+            {vm.filteredEntries.map((entry) => (
```
Moyasee, line 165: risk: no cap or virtualization here, and `readDirectory` stats every file. Browsing into something like `/usr/bin` or `C:\Windows\System32` renders a few thousand `FocusItem` buttons into the gamepad focus system in one go, which the native dialog never had to survive. Worth trying on a big directory on target hardware before merge; a simple entry cap with a "too many files" notice may be enough.

Moyasee (commented):
Did one more deep pass over the state machine and the five call sites. Four more things, two small bugs and two risks. Also traced error-state recovery, B press/hold handling, reopen focus and the preload wiring; those all check out fine.

PR #2392, opened 2026-06-22
fix(big-picture): honour gamepad.mapping="standard" in layout resolution

### LINKED ISSUE #2382: Unable to navigate big picture settings with my controller bumpers.
Greetings, just now after I updated hydra to 4.0.2, I can no longer navigate the settings tabs with controller bumpers, XInput obviously works, but DirectInput doesnt really seem to do so.

I didn't really look into the source code, but SDL2/3 seems like a good choice for DirectInput handling (if hydra doesnt use it already)

Controller: 8BitDo Ultimate 2 Wireless
Mode: DirectInput

ozekvi:

- [X] I have read the [Hydra documentation](https://docs.hydralauncher.gg/getting-started.html).
- [X] I have checked that there are no duplicate pull requests related to this request.

**Fill in the PR content:**
getGamepadLayout() was doing id-pattern matching unconditionally, then
falling back to LINUX_XINPUT_MAPPINGS for any unrecognised device on Linux.
This breaks controllers where Chromium has already applied its own W3C
remapping (gamepad.mapping === "standard"): we were overwriting correct
standard indices with a hardware-specific layout

**Fix:** bail out early with STANDARD_GAMEPAD_LAYOUT when mapping === "standard".
Chromium's remapping is authoritative in that case - no id matching needed.

Also adds a raw-input layout for 8BitDo Ultimate 2 Wireless in DirectInput
mode (2dc8:3109, Linux), where mapping is "" and LB/RB appear at indices
6/7 rather than 4/5. Button indices derived from public HID descriptors;
if they're off on your hardware, the /diagnostics page shows raw indices

Fixes [#2382](https://github.com/hydralauncher/hydra/issues/2382)


ozekvi:
@greptile review again

Moyasee:
Nice cleanup, and I confirmed this isn't already handled on main. Two things before it goes in:

- Missing newline at end of file, which will trip `yarn format`. Can you re-run the formatter?
- The 8BitDo Ultimate 2 DirectInput indices are noted as derived from HID descriptors and not verified on real hardware. Since this targets #2382 directly, were you or the issue reporter able to confirm LB/RB actually land on 6/7 on a real unit? Would like to be sure before shipping a device-specific layout.

The rest checks out. The `makeAxisDpad` / `makeStickAxes` helpers preserve the existing indices for every layout.

Hachi-R:
Closing this because the branch history is no longer compatible with the current `main`. GitHub reports this PR as `DIRTY`/`CONFLICTING` and shows 5,106 commits across 1,373 files, so the intended change can no longer be reviewed safely. Feel free to reopen this PR after rebuilding the branch or cherry-picking the intended commits onto the current `main`.

PR #2395, opened 2026-06-23
fix: restart http download when server ignores range header on resume

### LINKED ISSUE #2388: Downloads adding gigs to game
When downloading a game, the download is about done, and then it sorta lags and then adds gigs to the download. The game went from 140 GB to 300 GB.

Moyasee:

**Fill in the PR content:**

Closes #2388

### Problem

When an HTTP download resumes after an interruption, we send a `Range` header asking the server to continue from where we stopped. We assumed the server always honors that. Some servers ignore it and respond `200` with the full file from byte 0 instead of `206`.

When that happened:
- `parseFileSize` computed `startByte + content-length`, inflating the total. A download sitting at ~140GB jumped to ~300GB (this is the report in #2388).
- The writer was in append mode, so the full file got glued onto the existing partial, producing an oversized, corrupt file on disk.

### Fix

In `executeDownload`, detect a `200` response on a resume (`startByte > 0`). That means the server ignored our range and is sending the whole file again. In that case:

- `parseFileSize` uses the raw `content-length` as the total instead of adding `startByte`, so the size is no longer inflated.
- We keep the existing partial on disk and discard the first `startByte` bytes off the response stream, then append only the remainder. The stream skips those bytes in `createReadableStream` via a new `bytesToSkip` parameter, so they never reach the write stream. This avoids both the corruption and re-writing the bytes we already have.
- During the skip phase we still mark activity so the stall watchdog does not trip while we are draining the leading bytes.

Genuine `206` resumes are untouched.

### Notes

- Tradeoff: when a server ignores the range, the skipped bytes are still re-transferred over the network (we just drop them instead of writing them), so resume does not save bandwidth in that case. It does avoid the corrupt file and avoids rewriting the existing partial. This only triggers on servers that ignore range requests.
- Considered clamping the measured size against the catalog size in the download manager as a backstop, but skipped it: the catalog size is often approximate and clamping would risk breaking legitimate downloads. The root cause is fixed at the source, so the backstop is not needed.
- HTTP status codes used in this path were extracted into named constants (`HTTP_OK`, `HTTP_PARTIAL_CONTENT`, `HTTP_RANGE_NOT_SATISFIABLE`, `HTTP_CLIENT_ERROR_THRESHOLD`).

### Testing

- `yarn typecheck` passes
- `yarn test` passes (4/4)
- `yarn format` clean

### DIFF
--- a/src/main/services/download/js-http-downloader.ts
+++ b/src/main/services/download/js-http-downloader.ts
@@ -353,7 +353,8 @@ export class JsHttpDownloader {
 
     const contentLength = response.headers.get("content-length");
     if (contentLength) {
-      this.fileSize = startByte + Number.parseInt(contentLength, 10);
+      const length = Number.parseInt(contentLength, 10);
+      this.fileSize = response.status === HTTP_OK ? length : startByte + length;
     }
   }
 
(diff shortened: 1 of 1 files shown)


Moyasee:
@greptile review

Moyasee:
@greptile review

Moyasee:
@greptile review

PR #2398, opened 2026-06-23
fix: recover DDL downloads that freeze on hosters ignoring HTTP Range

Moyasee:
Closes LBX-723.

## What was happening

Downloads from DDL hosters (most noticeably Datanodes and Buzzheavier) would freeze partway through for some users. Pausing and resuming temporarily got them going again, but they never actually finished.

I pulled a set of user logs for a ~39 GB Buzzheavier download and the cause was clear: those hosters serve tokenized CDN links that contain no filename, so the real name only arrives in the `Content-Disposition` header. Every time the connection stalled, the resume logic looked for the partial file under the wrong name, didn't find it, and silently restarted the whole download from byte 0. On a large file over a flaky connection it could never converge.

A few related weaknesses made it worse: the server often answered a resumed request with a full `200` body (ignoring our `Range` header), which the old code appended onto the partial file and corrupted it; undici `terminated` socket drops were treated as fatal instead of retryable; the retry budget was cumulative for the whole download; and once it gave up it sat looking "active" forever instead of reporting an error.

## What this changes

- Remember the `Content-Disposition` filename and reuse it across retries so the partial file is found and the download truly resumes.
- If a resumed request comes back as a full `200` (server ignored `Range`), restart the file cleanly instead of appending and corrupting it.
- Treat `terminated` / `UND_ERR_*` socket drops as retryable.
- Reset the retry budget once data is flowing again (after ~4 MB of fresh progress), so occasional stalls don't accumulate to a permanent give-up.
- Detect stalls by outstanding reads instead of time-since-last-byte, and tighten the timing (stall timeout 8s → 6s, check interval 2s → 1s, max backoff 15s → 8s) so genuine stalls recover faster without falsely aborting healthy in-flight reads.
- When a download genuinely can't recover, mark it as errored and move the queue on instead of leaving it frozen.
- Surface a stalled state in the UI: the download progress now carries `isStalled` / `retryingInMs`, and the downloads view shows "Download stalled, reconnecting…" (with a retry countdown when one is pending).

The decision logic was extracted into small pure helpers (`classifyRetryOutcome`, `shouldResetRetryBudget`, `shouldRestartFromIgnoredRange`, `resolveResumeFilename`, `isRetryableDownloadError`) to keep the retry/resume branches readable.

## How to test

See the QA steps on the Linear issue (LBX-723).


Moyasee:
@greptile review

Moyasee:
@greptile review

Moyasee:
@greptile review

Moyasee:
@greptile review

Moyasee:
@greptile review please

Moyasee:
@greptile review

PR #2399, opened 2026-06-23
fix: harden JS HTTP downloader resume, retry and stall handling

### LINKED ISSUE #2388: Downloads adding gigs to game

Moyasee:

**Fill in the PR content:**

Reworks the native HTTP downloader so direct downloads stop stalling, freezing and growing past the real file size on some hosters. Closes the two reported issues and a handful of related edge cases found along the way.

Fixes #2388 (downloads adding gigs) and LBX-723 (DDL hosters freezing).

### Size doubling (#2388)
The downloader resumed by stat-ing the partial and sending `Range`, but never checked the server honored it. When a host ignores `Range` and replies `200` with the full body, the old code appended the whole file on top of the partial (140GB became ~300GB) and reported `startByte + contentLength` as the total.

- On a `200` resume, stream-skip the bytes already on disk and append only the remainder, keeping the partial intact instead of duplicating it.
- Report `contentLength` (not `startByte + contentLength`) as the total for a `200`.
- Align to the server's `Content-Range` start on a `206` so an early/overlapping partial cannot duplicate bytes, and restart cleanly if it would leave a gap.
- Refuse to append a body shorter than the existing partial.

### Freezing / stalling (LBX-723)
`undici` surfaces a dropped socket as `TypeError: terminated` with the real code on `err.cause`. The old retry check only looked at `err.message`/`err.code`, so a recoverable disconnect was treated as fatal and the download stopped (pause + resume was the only workaround).

- Treat `terminated`, `err.cause.code` and the `UND_ERR_*` family as retryable.
- Reset the retry budget once a meaningful amount of new data has flowed in an attempt, with an absolute cap so a genuinely dead link still gives up.
- Detect stalls by how long a single read has been blocked (not time since the last chunk) and raise the timeout, so a slow-but-alive transfer is not aborted. A download that truly cannot progress now raises an error and lets the queue advance instead of sitting "active" at 0 B/s.

### Transient HTTP errors
`408/429/500/502/503/504` were fatal, so a rate-limited link (common on Datanodes) failed to even start. They now back off and retry in both the preflight check and the downloader; permanent 4xx (403/404/410) still fail fast.

### Other hardening
- Send `Accept-Encoding: identity` and bail out of size-based resume under transfer compression.
- Cancel the response reader on teardown to avoid leaking sockets / file descriptors.
- Remember the `Content-Disposition` filename across retries so a resume targets the right partial.
- Clamp reported progress, fix AllDebrid batch accounting, route batch failures through the error handler, and add a generation token so a cancel during preparation cannot spawn a second writer on the same file.

### Testing
- Pure decision logic extracted into a helpers module with unit tests (`yarn test`, 41 cases).
- Verified end-to-end against a local server reproducing every observed behavior (200-on-resume, mid-stream `terminated`, 206 overlap, short body, 429, pause/resume, throttle, chunked) with byte-for-byte (SHA-256) checks.
- Verified against live pixeldrain and Datanodes downloads: partial download, pause, and 206 resume are byte-identical to the server with no corruption at the boundary.


Moyasee:
Addressed both points in 58432feb:
- `isRetryableDownloadError` now early-returns `false` when an error is explicitly tagged `retryable: false`, so the message-fragment fallback can't override a permanent failure (added a unit test for it).
- Pulled the preflight backoff delay out into `PREFLIGHT_RETRY_BASE_DELAY_MS`.

@greptile review

Moyasee:
Investigated koddy's tester logs (reported stall).

Root cause: it was not a mid-stream stall. He repeatedly hit GoFile, which was failing for him (official 403, alternate CDN `gofilecdn.eu.cc` returning 400/429). On the 429s, the new transient-status retry kept retrying for ~2.5 minutes (1s -> 15s backoff x10) while the download still showed as active, so it looked frozen. His other downloads (Celeste, The Forest on other hosts) completed cleanly, no actual stall.

Fix in 1d705bf2: transient HTTP statuses (429/503/etc.) now get a short dedicated retry budget separate from network/stall retries, so a hard rate-limited or unavailable link fails within a few seconds with a clear "rate-limiting or temporarily unavailable" message and the queue moves on, instead of masquerading as a stalled download. Also honors `Retry-After` (seconds or HTTP-date, capped). Added unit + integration coverage (persistent-429 now fails bounded/fast; 429-then-200 still recovers; 403 still fails fast).

Follow-up worth considering separately (not in this PR, it is in the GoFile resolver, not the downloader): when the alternate GoFile CDN returns 4xx/429, fall back to the official GoFile source instead of dead-ending on the bypass CDN.

Doobies:
> Investigated koddy's tester logs (reported stall).

> Root cause: it was not a mid-stream stall. He repeatedly hit GoFile, which was failing for him

i did nothing! I either clicked on the black space on the Hydra app because I clicked another app or alt-tabbed back into Hydra. When i did hit the gofile was to make sure TorBox caught the hook. This didn't happen because I went back into it and redownloaded the same file. The doubling of the GBs happened while I was watching it and didn't have any peripherals in my hands

Moyasee:
> > Investigated koddy's tester logs (reported stall).
> 
> > Root cause: it was not a mid-stream stall. He repeatedly hit GoFile, which was failing for him
> 
> i did nothing! I either clicked on the black space on the Hydra app because I clicked another app or alt-tabbed back into Hydra. When i did hit the gofile was to make sure TorBox caught the hook. This didn't happen because I went back into it and redownloaded the same file. The doubling of the GBs happened while I was watching it and didn't have any peripherals in my hands

Hey! This text you are quoting is not related to your issue, no worries
But the issue that you reported was also fixed in this PR

Moyasee:
@greptile review

Hachi-R (approved):
LGTM, thanks for hardening this.

I reviewed the downloader changes and didn't find any blocker. The resume/retry/reconnect behavior looks much safer now, and the CI is green.

PR #2403, opened 2026-06-24
fix: resolve cue FILE refs by basename against cue dir

Moyasee:

**Fill in the PR content:**

PS1/PS2 disc SKU extraction during LaunchBox ROM import returned `null` whenever a `.cue` sheet's `FILE` entry pointed somewhere other than a sibling filename. Some discs were mastered with a stale absolute path, e.g.:

```
FILE "C:\WINDOWS\DESKTOP\CASTLEVANIA.BIN" BINARY
```

`parseCueReferencedFiles` fed that value straight into `path.resolve`, which discards the cue's own directory for an absolute/`\`-separated path and hands back the bogus location. `fs.open` then throws `ENOENT`, the generic catch swallows it, and the real sibling `.bin` is never scanned.

**What this does**

- Strips the `FILE` value to its basename (normalizing `\` to `/` first) and resolves it against the cue's own directory.
- If the resolved path is missing on a case-sensitive filesystem (cue says `CASTLEVANIA.BIN`, file is `Castlevania.bin`), it scans the directory and matches the basename case-insensitively, returning the actual on-disk name.
- Adds an explicit `logger.warn` in `extractPs12Sku` when the resolved sniff target does not exist, so future cases are diagnosable instead of failing silently.

**Testing**

Verified end-to-end against a real affected ROM (`Castlevania.cue` + `Castlevania.bin`): the cue's bad `C:\WINDOWS\DESKTOP\CASTLEVANIA.BIN` now resolves to the sibling `Castlevania.bin`, and the scan yields the correct SKU `SLUS-00067`. `tsc` typecheck passes.


Moyasee:
@greptile review

Hachi-R (approved):
LGTM, thanks for fixing this.

I reviewed the change and didn't find any blocker. The fix is nicely scoped, handles the bad absolute CUE refs case, and the case-insensitive fallback should help on Linux too.

PR #2404, opened 2026-06-24
Feature/lbx 794

Hachi-R:


src/big-picture/src/styles/globals.scss
```diff
@@ -167,3 +167,8 @@ li {
   margin: 0;
   padding: 0;
 }
+
+[data-bp-input-mode="mouse"] [data-focused="true"] {
```
Moyasee, line 171: This mouse-mode reset only clears outline and box-shadow, but several cards style off [data-focused="true"] with more than that: vertical-game-card, horizontal-library-game-card, horizontal-store-game-card and challenge-game-card all set background-color/border-color and reveal the action button on [data-focused]. Combined with the cleanup below (currentFocusId gets restored by syncFromService), those highlights can still leak in mouse mode. Might be cleaner to gate the card focus styles by input mode too, or drive the highlight off data-focus-visible.

src/big-picture/src/components/providers/input-mode.provider.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const prevModeRef = useRef(mode);
+
+  useLayoutEffect(() => {
+    const prevMode = prevModeRef.current;
+    prevModeRef.current = mode;
+
+    if (prevMode !== "gamepad" || mode !== "mouse") return;
+
+    (document.activeElement as HTMLElement)?.blur();
+    useNavigationStore.setState({ currentFocusId: null });
```
Moyasee: This nulls the store copy of currentFocusId but not NavigationService, which is the source of truth. getCurrentFocusId() still returns the old id, so the next navigation.subscribe -> syncFromService (node registration, route change, etc.) restores it and the card gets data-focused="true" again while in mouse mode. Clearing focus through the service would make this stick.

src/big-picture/src/hooks/use-input-mode.hook.ts
```diff
@@ -0,0 +1,15 @@
+import { useInputModeStore } from "../stores";
+
+export function useInputMode() {
```
Moyasee: useInputMode doesn't seem to be referenced anywhere, everything reads useInputModeStore directly. Drop the hook or wire it in?

Moyasee (commented):
Reviewed the input-mode (gamepad vs mouse) work. Core approach reads well: input-mode store + a set of provider detectors, gating focus visuals, nav audio and the virtual keyboard on gamepad mode. Main thing to double check is whether focus styling is fully suppressed in mouse mode (see inline on globals.scss and the focus cleanup), since the [data-focused] card styles do more than just outline/box-shadow.

A couple of smaller notes:

- src/big-picture/src/stores/input-mode.store.ts:12 - initial mode defaults to "mouse". Big picture is usually launched for controller use, so the focus ring, nav sounds and the virtual keyboard all stay inert until the first gamepad input. Is that intended? First button press flips it, so maybe fine, just flagging.
- src/big-picture/src/components/providers/navigation-input.provider.tsx - the per-DPAD setGamepadMode() calls duplicate what GamepadService.onAnyButtonPress (wired in InputModeGamepadDetector) already does on any button press. The stick/axis handlers genuinely need it, but the button ones look redundant.

Nothing blocking from me.

Hachi-R:
@greptile

chubbygrannychaser (changes requested):
Thanks for the updates here. The previous focus-cleanup/style concerns look fixed, and the local typechecks pass now.

I’m still blocking on the current overlay behavior: while in gamepad mode the full-screen overlay uses `pointerEvents: "auto"`, so a direct first mouse click targets the overlay instead of the underlying control. The mode flips to mouse, but the intended button/link never receives that click, forcing users to click a second time. Please make the transition to mouse mode not consume the first click/scroll (for example, avoid an intercepting pointer-events overlay and rely on capture-level pointer/wheel listeners, or otherwise forward/replay the intended interaction safely).

Validated locally after `yarn install --ignore-scripts --frozen-lockfile`:
- `npm run typecheck:node` ✅
- `npm run typecheck:web` ✅
- targeted ESLint on changed JS/TS files ✅ (warnings only: existing `react-hooks/exhaustive-deps` warnings in `focus-item/index.tsx`)

chubbygrannychaser (changes requested):
Thanks for the follow-up. The new capture listener moves some of the detection out of React handlers, but the previous blocker still applies for direct first interactions.

On a direct click while in gamepad mode, the browser hit-tests the full-screen `#bp-input-overlay` first because it still has `pointerEvents: "auto"`. The window capture `mousedown` handler can set `overlay.style.pointerEvents = "none"` and flip to mouse mode, but that does not retarget the already-dispatched `mousedown` to the underlying button/link, so the intended control still misses the first click. The same applies to wheel: capture can switch modes, but the scroll event target was the overlay, so the underlying scroller does not receive that first wheel event.

Please avoid putting an intercepting overlay above the app for this transition (for example keep the overlay non-interactive and use capture-level pointer/mouse/wheel listeners only for mode detection, or forward/replay the original event in a way that preserves the intended first click/scroll).

Validated current head `ad35d04` locally:
- `npm run typecheck:node` ✅
- `npm run typecheck:web` ✅
- targeted ESLint on changed JS/TS files ✅ (warnings only: existing `react-hooks/exhaustive-deps` warnings in `focus-item/index.tsx`)

CI at review time: lint/text-integrity/SonarCloud passed; build jobs were still pending.

chubbygrannychaser (approved):
LGTM — thanks for the follow-up fixes. The prior first-click/first-scroll blocker is resolved on this head: the intercepting overlay is gone, and mouse/wheel/context-menu detection now runs through capture-level listeners without putting a `pointer-events: auto` element over the app.

Validated current head `2f87e84` locally:
- `npm run typecheck:node` ✅
- `npm run typecheck:web` ✅
- targeted ESLint on changed JS/TS files ✅ (warnings only: existing `react-hooks/exhaustive-deps` warnings in `focus-item/index.tsx`)

CI at review time: lint/text-integrity/SonarCloud passed; build jobs were still pending.

PR #2405, opened 2026-06-24
feat: bind torrent client to a network adapter

### LINKED ISSUE #2023: Add VPN Binding for Torrenting
Many torrenting applications allow support to 'bind' to a network adapter, so that one can only torrent through a VPN. It would be lovely to have that feature on Hydra as well.

Moyasee:

**Fill in the PR content:**

Closes #2023.

Adds a setting to bind the torrent client to a specific network adapter, so torrent traffic only goes through it. The main use case is forcing torrents through a VPN.

It works as a kill switch: while bound to an adapter, torrents only connect through that adapter, and downloads stop if it goes down instead of falling back to your regular connection and leaking traffic.

### How it works

- The selected adapter is applied to the libtorrent session through `listen_interfaces` and `outgoing_interfaces`, the same approach qBittorrent uses, so we bind by device name and it stays stable across VPN reconnects.
- The choice is saved as a user preference (`torrentNetworkInterface`), applied live when changed, and re-applied on RPC startup before any torrent is added.
- A new `getNetworkInterfaces` IPC handler lists the available adapters (via `os.networkInterfaces()`) for the dropdown. Loopback is filtered out.
- New setting lives under Settings > Downloads, with a hint explaining the kill-switch behavior. If a previously selected adapter is currently unavailable (VPN down), it still shows in the dropdown so the saved choice is visible.
- Scope is torrents only. Direct/debrid HTTP downloads are untouched.
- Translations added for en, es, pt-BR, pt-PT and ru.

### Notes

Device-name binding works across platforms with libtorrent 2.0. On Windows the adapter friendly names from `os.networkInterfaces()` are used, which match what libtorrent expects, but it would be good to get a Windows test in.

### Testing

- Select a VPN adapter, start a torrent, and confirm peer connections egress the VPN interface.
- Bring the VPN down and confirm torrents stall instead of leaking, then restore.
- Restart the app and confirm the saved adapter persists and is re-applied before downloads resume.


Moyasee:
@greptile review

Hachi-R (approved):
I think this needs one more check before merging.

The PR says the Windows adapter names from `os.networkInterfaces()` match what libtorrent expects, but libtorrent's docs say Windows adapter device names must be passed as the adapter GUID in `{...}`, not the friendly interface name.

Since this setting is meant to work as a kill switch, I think we should either map Windows interfaces to the GUID that libtorrent expects, or confirm with an actual Windows test that the current names from `os.networkInterfaces()` are accepted by our py-libtorrent build.

Hachi-R (changes requested):
I think this needs one more check before merging.

The PR says the Windows adapter names from `os.networkInterfaces()` match what libtorrent expects, but libtorrent's docs say Windows adapter device names must be passed as the adapter GUID in `{...}`, not the friendly interface name.

Since this setting is meant to work as a kill switch, I think we should either map Windows interfaces to the GUID that libtorrent expects, or confirm with an actual Windows test that the current names from `os.networkInterfaces()` are accepted by our py-libtorrent build.

Moyasee:
Good catch, I think you're right and this is the weak spot on Windows.

libtorrent matches the string in `listen_interfaces`/`outgoing_interfaces` against the interface's device name or an IP. On Linux/macOS the device name from `os.networkInterfaces()` (tun0, utun4, eth0) is exactly what libtorrent expects, so binding by name works there. On Windows though, libtorrent's device name is the adapter GUID (`{XXXX-...}`), while `os.networkInterfaces()` returns the friendly name ("Wi-Fi", "Ethernet"), so the value we pass wouldn't match. Since this is a kill switch, "silently fails to bind" is the worst way for it to break. I wasn't able to verify against a real Windows build from here.

A few ways to close it:

1. Bind by IP instead of by name (my preference). `os.networkInterfaces()` already gives us the addresses cross-platform, and libtorrent accepts an IP in both settings. We keep storing the interface name in prefs, but resolve it to its current address at apply time (in `startRPC` and on setting change). That sidesteps the GUID problem entirely, behaves the same on all three platforms, and re-resolving on each apply also handles the VPN reconnecting with a new IP. The kill switch still holds: if the adapter is gone there's no address to bind, so we pass the stored name as-is and libtorrent refuses to bind rather than leaking.

2. Map friendly name to adapter GUID on Windows (e.g. `Get-NetAdapter | InterfaceGuid`, or a native call), pass the GUID to libtorrent, keep the friendly name in the dropdown. Keeps name-based binding but adds Windows-specific plumbing.

3. Leave as-is only if a real Windows test confirms our py-libtorrent build accepts the friendly names.

I'd lean towards option 1. Happy to rework the branch to resolve name to IP at apply time, or hold for a Windows test first, whichever you prefer.

Hachi-R (approved):
thanks for the follow-up. this is definitely better than passing the windows friendly name directly, but i think there is still one important thing to clarify before merging.

the new code resolves the selected adapter name to its current IP addresses and sends those IPs to libtorrent. that avoids the windows GUID issue in the common path, but it is now IP binding, not adapter/device binding.

libtorrent's docs say binding outgoing connections to a local IP does not necessarily make the connection go through the associated NIC/adapter. since this feature is described as a kill switch, i think we should either bind to the actual device identifier that libtorrent expects per platform, or explicitly confirm with a windows + vpn smoke test that this IP-based approach does not leak when the VPN is up/down.

also, the PR description still says we bind by device name and that windows friendly names match what libtorrent expects, which is no longer what the code does.

jazir555:
Thank you so much for implementing this guys, very much appreciated.

Issue #2337, opened 2026-06-15
Are you adding on flathub


ScorlyFly:
Will good if you add build in flathub

chubbygrannychaser:
Hydra has been refused on Flathub. It won't be added.

PR #2406, opened 2026-06-24
feat(flatpak): package Hydra as a Flatpak

VictorVow:
## What

Adds a **Flatpak** target to the Linux build, alongside the existing AppImage/snap/deb/rpm outputs, and wires it into the release pipeline.

## Why

Flatpak is the most common way Linux users install desktop apps today, and it isn't currently a first-class Hydra artifact.

## How it works

**Packaging**
- New `flatpak` target in `electron-builder.yml`, built on `org.electronjs.Electron2.BaseApp` / freedesktop runtime `23.08`.
- Release workflow builds and publishes the `.flatpak` artifact.

**Filesystem access**
- Ships with `--filesystem=host` by default, so user-picked download and game-library folders are reached as **native paths** — avoiding the document-portal FUSE layer, which adds noticeable overhead on heavy I/O (large downloads, archive extraction). This matches what most game launchers on Flathub do.
- Security-conscious users can tighten the sandbox with a `flatpak override` (`--nofilesystem=host` plus a narrower `--filesystem`). The path-grant plumbing below makes that work transparently.

**Portal-aware path handling** (active when the host grant is narrowed)
- A `PathGrants` service maps document-portal FUSE paths back to their real host paths via the `user.document-portal.host-path` xattr (read through a small native addon), so the UI always shows familiar locations.
- Startup recovery pauses affected downloads if a portal grant is lost, instead of failing writes.

**Supporting changes**
- libtorrent **posix** disk-I/O backend under Flatpak (the default mmap backend is unsafe on doc-portal FUSE paths), feature-detected with a fallback for older bindings.
- `xdg-download/Hydra` as the default downloads location under Flatpak.
- Steam/Proton and Wine-prefix path selection routed through the same grant-aware helpers.

## Notes for reviewers
- Two commits: the release-pipeline change, and the feature itself.
- The default `--filesystem=host` is a deliberate performance-vs-sandbox tradeoff; happy to discuss narrowing it by default if maintainers prefer, since the portal plumbing already supports the tighter mode.


ozekvi:
The startup grant check in loadState only runs for torrents that are already progress === 1 && status === "seeding" - active downloads skip past it entirely. If a grant gets revoked while a download is still in progress it won't get paused gracefully, it'll just hit a write failure on resume. Same blind spot for wine-prefix and proton paths. They get annotated via PathGrants.annotate but there's no equivalent verifyAccess check for them at startup only for download.downloadPath. So a lost grant there just shows up as a random failure when the user tries to launch the game, instead of the toast treatment downloads get. Rest of the PathGrants/xattr work looks solid, just want startup recovery actually covering what gets annotated

VictorVow:
Thanks, handled in the latest commit.

DRKCTRLDEV:
Can’t wait for Flatpak support. Does this add any kind of isolation for Proton prefixes, like Steam’s bubblewrap setup? That would make it much safer to run programs that could even be malicious and help keep our systems more secure.

Moyasee (commented):
Took a pass over this. The portal/PathGrants design is thoughtful and the display-path vs access-path split is handled cleanly in the renderer. A few things worth discussing before merge.

### 1. PathGrants accumulate and are never removed -> recurring false "folder access lost" toasts

`PathGrants.annotate()` persists a grant for *every* doc-portal path that goes through `showOpenDialog`, and there are ~24 callers, most of them one-off pickers (profile background image, theme editor, emulator setup, scan games, memory cards, game assets, etc.) that have nothing to do with downloads or the library.

`removeGrant()` exists but is never called, and `listBroken()` walks *all* persisted grants. Document-portal grants are ephemeral, so once one of those one-off picks lapses (app restart, portal prunes it, or the same folder is re-picked and gets a new doc id), it shows up as broken on the next launch and the user gets a "Folder access lost" toast naming a folder they picked once for something unrelated. The sublevel also grows unbounded.

Two options: only persist grants for the paths we actually track long-term (downloads path, wine prefix, proton, game executable, custom backup) instead of blanket-annotating in `showOpenDialog`, and/or prune a grant once it has been surfaced as broken / replaced. Right now `annotate` in `show-open-dialog.ts` is the wide net.

### 2. `lt.session_params(settings)` with a dict (python_rpc/main.py)

The original code passes the dict straight to `lt.session(...)`, which has a binding for that. `create_torrent_session` instead does `lt.session_params(settings)` with the same dict. Worth confirming that the `session_params` constructor accepts a plain dict on the libtorrent version we bundle (requirements pin is unpinned). If it raises, the `except` swallows it and we silently fall back to the default mmap backend, which is the exact thing this change exists to avoid on doc-portal FUSE paths. The fallback logs a warning, so it is detectable, but please verify the happy path actually engages posix under Flatpak.

### 3. `--filesystem=host` default and Flathub

Calling out what the PR already flags: with `--filesystem=host` the sandbox provides essentially no filesystem isolation, which makes most of the portal/PathGrants plumbing dormant by default. That is a reasonable perf call for a side-loaded `.flatpak`, but if there is any intent to publish on Flathub, their reviewers routinely reject `--filesystem=host`. Might be worth deciding now whether the tightened mode (`--filesystem=xdg-download` + portal) should be the shipped default so the plumbing is actually exercised.

### 4. Game launching inside the sandbox

How do proton/umu/wine actually run under the Flatpak? They are spawned as subprocesses inside the sandbox (no `flatpak-spawn --host` / `--talk-name=org.freedesktop.Flatpak` here). With `--filesystem=host` + network + dri it may well work, but this is the part most likely to break in practice. Have games been launched end-to-end from the built `.flatpak`?

### 5. i18n parity

`lost_folder_access` / `lost_folder_access_description` were only added to `en`. They also need to land in `pt-BR`, `pt-PT`, `ru`, and `es`.

### Minor
- The flatpak build step was added to `build.yml` (runs on every build, twice) in addition to `release.yml`. Expect noticeably longer CI on each run. Fine if intentional.
- `main.ts` uses `await import("./constants")` for `defaultDownloadsPath`, but `constants` is already in the module graph, so this does not defer anything. A plain top-level import reads cleaner.
- `verifyAccess` checks `R_OK`, but for a downloads target write access is what matters. `W_OK` would be a truer signal, though R_OK is a fine proxy in most cases.

VictorVow:
Thanks for the thorough review. Addressed the actionable items in separate commits, replies are below:

### 1. PathGrants accumulate -> recurring false "folder access lost" toasts

Fixed in two parts:

- **Stopped the wide net.** `show-open-dialog.ts` no longer calls `annotate()`. It now only *resolves* the display path (`PathGrants.getDisplayPath`, non-persisting) so the renderer still shows a familiar host path for any pick, but nothing is written to the sublevel. Persistence is now limited to the callers that actually track a path long-term and already annotate explicitly: downloads, wine prefix, proton, game executable, and the ludusavi backup path. The one-off pickers (profile image, theme editor, emulator setup, memory cards, etc.) no longer leave anything behind.
- **Prune on surface.** `listBroken()` now removes each grant as it reports it broken (via the previously-dead `removeGrant`). A doc-portal id only lapses once, so it gets surfaced a single time and dropped, instead of re-toasting on every subsequent launch. This also covers the "re-picked folder gets a new doc id" case - the stale entry is pruned the first time it's seen as unreachable.

### 2. `lt.session_params(settings)` with a dict

Agreed the silent fallback is the risk here. I've confirmed `lt.session_params` accepts a plain settings dict on the libtorrent build we bundle, and verified the posix backend actually engages under Flatpak (rather than silently falling back to mmap). I'll keep the warning log so any future binding drift is detectable rather than silent.

### 3. `--filesystem=host` default and Flathub

I saw on an earlier issue that [Flathub already rejected Hydra](https://github.com/hydralauncher/hydra/issues/2337) so I wasn't designing for them. The tradeoff here is performance vs sandboxing, seeing as all the existing non-Flatpak release builds also opt for performance without sandboxing I made the same decision here to get that working first. This default also means that users who already keep their Hydra library somewhere like an external drive aren't going to have to take a manual step in order to enable it.

### 4. Game launching inside the sandbox

I successfully tested this with a few games with default settings, the tightened sandbox mode however is just a stub for future work I'm leaving with an eye to land more fixes for in the future.

Pushing on the tightened mode in the latest round of fixes did surface some more bugs which have now been fixed and committed too. However there's still remaining work to be done: currently Proton's pressure-vessel can't bring up its nested `bwrap` container inside the Flatpak sandbox without more plumbing; the freedesktop 23.08 base has no i386 runtime (`Cannot determine ld.so for i386-linux-gnu`), `/dev/shm` isn't shared (flatpak#4214), and the nested namespace setup fails (`bwrap: Can't find source path /proc/self/fd/25`). Making that work means either pulling in the i386 compat + GL32 extensions and opening up `/dev/shm`/devel, or routing the launch through `flatpak-spawn --host`, which are both bigger than this PR. So tightened mode Windows-game launch stays a WIP stub path for now; default `--filesystem=host` mode is the supported one and launches fine.

### 5. i18n parity

Fixed.

### Minor

- **`verifyAccess` R_OK vs W_OK** - agreed for download targets. `verifyAccess` now takes an optional access-mode arg (defaults to `R_OK`, since proton/wine/executable grants are read/exec checks), and the two download-target checks (`get-downloads-path` and the startup `hasLostPathGrant`) pass `W_OK` so they reflect what actually matters there.
- **`main.ts` `await import("./constants")`** - switched to a plain top-level import, it wasn't deferring anything.
- **flatpak step in `build.yml` (twice)** - intentional. `build:linux` runs `electron-builder --linux`, which builds the flatpak target alongside AppImage/snap/deb/rpm — all of which already build in both the staging and production jobs. Keeping flatpak there gives the same PR-time "does the artifact still build" validation the other Linux targets get. If CI time becomes a concern I'm happy to gate it behind a path/label filter, but I'd rather not let the flatpak target silently rot relative to the others.

Moyasee:
Went through the new commits, fixes all look good. The non-persisting `getDisplayPath` in `show-open-dialog` plus pruning in `listBroken` is exactly the right split, and the W_OK distinction for download targets is a nice touch. Thanks for the thorough writeup on the tightened-mode blockers too, that context is useful.

Two corrections on the Greptile summary for anyone reading it, since both are off:

- It claims `python_rpc/main.py` assigns a string to `disk_io_constructor` and so always falls back to mmap. That's a misread. The code assigns `lt.posix_disk_io_constructor`, the actual libtorrent callable, guarded by `hasattr`. VictorVow also confirmed manually that posix engages under Flatpak.
- It says `listBroken()` has no call site. It's called by `notifyLostPathGrants` in `main.ts`.

Nothing outstanding from me.

Moyasee:
@greptile review

VictorVow:
One moment, I think I've found a fix for the tightened sandbox mode too.

VictorVow:
Windows games can now launch via Proton **without `--filesystem=host`** - verified end-to-end (game window opens) in a tightened sandbox.

**What was blocking it:** electron-builder's flatpak target can't declare runtime extensions, so the sandbox had no 32-bit loader (`Cannot determine ld.so for i386-linux-gnu`), and pressure-vessel fell back to a raw nested bwrap that couldn't pass fds.

**Fix:** a hand-authored `flatpak-builder` manifest (`flatpak/`) that ships the `Compat.i386` + `GL32` extensions, runs on the `25.08` runtime (matching the Steam Flatpak), and uses `--talk-name=org.freedesktop.Flatpak` so pressure-vessel builds its container as a portal sub-sandbox (Lutris's approach). CI now builds the Flatpak from this manifest instead of electron-builder's target.

**Tightened override (no host access):**

```sh
flatpak override --user --nofilesystem=host \
  --filesystem=xdg-download \
  --filesystem=xdg-data/Steam:create \
  --filesystem=xdg-data/umu:create \
  gg.hydralauncher.hydra   # + --filesystem=<library> if games live outside xdg-download
```

Tighter than the `--filesystem=home` scope the Lutris Flatpak uses. `xdg-download`-only still works to browse/download, just not to launch.

**Default `--filesystem=host` mode is unchanged and still the shipped default;** this is purely additive for users who want a tighter sandbox.

Moyasee:
Went through the tightened-sandbox round. Nicely done getting Proton up without `--filesystem=host`. The hand-authored manifest with the Compat.i386/GL32 extensions, the 25.08 runtime bump, and the portal sub-sandbox route all read well, and the getcwd-safe working directory fallback for doc-portal FUSE paths is a clean fix for the umu crash. The glibc strip in `build-flatpak.cjs` is in the safe direction too (runtime glibc is newer, so it only adds symbols). README covers the install prerequisites thoroughly.

One thing I want to flag: the `binaries/7zzs` swap went from 32-bit i386 to 64-bit x86-64. That binary ships in every Linux artifact (AppImage/snap/deb/rpm), not just the Flatpak, so this effectively drops 32-bit Linux host support across the board, not only in the sandbox. In practice that's almost certainly a non-issue, but since the blast radius is wider than Flatpak I wanted to confirm it's intended rather than a side effect of getting extraction working in the sandbox.

Moyasee:
Couple of cleanup items before this is good to go:

**Two bits that look like they snuck in and don't do anything for the Flatpak:**

- `.gitignore` adds `.flatpak-builder/`, `.flatpak-builder-build/`, `.flatpak-repo/`, but `build-flatpak.cjs` keeps all its scratch under `os.tmpdir()/hydra-flatpak-build/` (build/repo/state), never in the repo. The header comment even says scratch lives outside the tree, and the ignored names don't match the script's dirs, so these three lines are dead. Can drop them.
- `electron-builder.yml` excludes `!.direnv/**` and `!.venv*/**`. Those look like a workaround for a local direnv/nix venv setup rather than anything Flatpak related, and CI won't have those dirs. Probably fine to drop unless I'm missing a reason it's needed.

**SonarQube:** the 4 new issues are all the same lint, `Prefer globalThis over window`:
- `src/renderer/src/app.tsx:120`
- `src/renderer/src/hooks/use-display-path.ts:22`
- `src/renderer/src/pages/game-details/cloud-sync-files-modal/cloud-sync-files-modal.tsx:51` and `:75`

Quick swap to `globalThis.window` (matching what the rest of the renderer already uses) clears the gate.

VictorVow:
> One thing I want to flag: the binaries/7zzs swap

Yes it's intended since Hydra's Linux builds already require a 64-bit Linux host (Electron, Electron Builder, hydralauncher and ludusavi all are 64-bit), it was just `7zzs` that was 32-bit. Likely it was a mistake originally since the same commit only added 64-bit versions for macOS and Windows.

> Couple of cleanup items

Right those two snuck in because of my local environment, removed now. Also fixed the SonarQube comment.

DRKCTRLDEV:
Will this be added in 4.0.5?

Moyasee:
Did another pass now that all the earlier rounds are in. This is in really good shape, the posix backend, listBroken pruning, non-persisting getDisplayPath, and the in-progress-download recovery all check out. A few things left:

**Needs a rebase.** GitHub is showing this as conflicting with main right now, so it'll need updating before it can merge.

**Downloads is the one long-term path that never gets annotated.** proton/wine/executable/backup all call `PathGrants.annotate()`, but the downloads-path picker doesn't. So in tightened mode, when a downloads grant is lost, `getDisplayPath` can't resolve it anymore (the path is already unreachable, the xattr read fails), and the "folder access lost" toast ends up showing the raw `/run/user/.../doc/...` path for downloads instead of a friendly one. The round-1 change to stop blanket-annotating in `showOpenDialog` was right, but downloads is genuinely long-term-tracked like the other four, so annotating it explicitly at pick time would close that toast gap without bringing back the one-off-picker noise.

**Worth a line in the docs on the default sandbox posture.** The shipped `finish-args` (`--filesystem=host`, `--device=all`, `--allow=devel`, `--talk-name=org.freedesktop.Flatpak`) effectively mean no filesystem isolation by default. That matches the Steam/Lutris precedent and the perf tradeoff you already called out, so no change needed, but it might be worth stating plainly in the README/metainfo so nobody assumes the default install is sandboxed.

**Minor, tightened-mode only:** in `general-section.tsx` the custom-path `TextField` uses `value={customDisplayPath}` with `onChange` writing back into `customPath` (the access-path state). When display and access paths differ, the first keystroke overwrites the access path with the display string. Under the default `--filesystem=host` they're equal so it's dormant, just flagging for the tightened path.

None of these block except the rebase. Nice work on this overall.

VictorVow:
Thanks for the review, four new commits added to address those last comments.

Moyasee (commented):
Went through the whole thing. This is a seriously thorough PR, and the commit history shows most of the obvious review points (one-off grant accumulation, W_OK vs R_OK, covering in-progress downloads at startup, the getcwd/umu crash, the backup custom-path binding) were already caught and fixed. The portal path-grant plumbing and startup recovery are well thought out. Not blocking on anything below, but a few things worth a maintainer decision before merge:

Sandbox default. `--filesystem=host` means the default install is effectively unsandboxed, which is a real product/security call rather than a code question. The precedent you cite (Steam/Lutris/Bottles/Heroic) is solid and the README is honest about it, so I think it is defensible, but it should be an explicit maintainer sign-off and ideally called out in the release notes so users know the default install can read/write their home like the native build. Since the portal plumbing already supports the tightened mode, worth a short discussion on whether the default should be narrower.

CI cost. The "Install Flatpak build tooling" step pulls Platform + Sdk + BaseApp + Compat.i386 + GL32 on every run with no caching, and it is duplicated across both build.yml jobs and release.yml. That is a few GB and several minutes added per build. Worth caching ~/.local/share/flatpak (or the flatpak-builder state dir) and factoring the step out.

Minor, non-blocking:
- `isDocPortalMountAvailable` only checks the `/run/user/<uid>/doc` prefix, not `/run/flatpak/doc`. If a users portal mount is the latter, startup grant recovery quietly no-ops. It fails safe (no false pauses), so low impact.
- `get_xattr` goes through `String::from_utf8_lossy`, so a non-UTF8 host path would come back with replacement chars in the display path. Edge case and display-only.
- Worth double-checking the renderer default downloads path (via `prepareDefaultDownloadPathSync` fallback) matches mains `xdg-download/Hydra` under Flatpak, otherwise the settings UI could show a different default than main actually writes to.

Nice work overall.

Moyasee (commented):
Correction on my last point: scratch the "renderer default downloads path" note. I traced it, the settings pages get their default from `window.electron.getDefaultDownloadsPath()`, which returns the same `defaultDownloadsPath` constant main uses, so under Flatpak both resolve to `xdg-download/Hydra`. No mismatch. The other notes stand.

VictorVow:
Have addressed the actionable feedback, cheers.

Moyasee:
Re-reviewed after the update. The three things I flagged last time are all fixed, and fixed properly:

- `isDocPortalMountAvailable` now checks every configured prefix (49c5775), so a `/run/flatpak/doc` mount is covered. `docPortalMountRoot` is gone with no stale references left behind.
- `get_xattr` returns `None` on non-UTF-8 (7bac363), and `resolveHostPath` already falls back to the access path, so a garbled display path is no longer reachable.
- The Flatpak tooling install is a composite action with the user install dir cached (ebb1660). The cache came out at 933 MB, in line with the yarn caches the repo already keeps, so it should be fine on the cache budget. Build is green on that commit.

I also went back on my own point about the renderer default download path: it turns out to be a non-issue. The renderer gets `defaultDownloadsPath` over IPC from main's constant, so it can't drift from `xdg-download/Hydra`.

One new thing, minor:

### `listBroken()` prunes the download-path grant, so only the first lost-grant toast shows a friendly path

`notifyLostPathGrants` resolves display paths for the paused downloads first, then calls `listBroken()`, which walks every persisted grant and calls `removeGrant()` on each broken one. Download paths are annotated now (09ba6d3), so the download's own grant gets pruned on the first launch after the grant lapses, even though it is correctly filtered out of `otherBrokenDisplayPaths`.

On the next launch the grant is still broken, `hasLostPathGrant` fires again, but `getDisplayPath` has no persisted grant left to read and the xattr lookup on the now-unreachable FUSE path fails, so it falls back to the raw path and the toast reads `/run/user/1000/doc/<id>/...`. Same for the downloads path shown in settings via `useDisplayPath`. So the first toast is friendly and every one after it is opaque, which is what 09ba6d3 set out to avoid.

Simplest fix is to pass the already-handled download paths into `listBroken()` and skip pruning those. They are still referenced by a live download, and they get re-annotated whenever preferences are saved, so there is nothing stale about keeping them.

### Sandbox default

Not a code issue, just flagging where it landed: the shipped `finish-args` are `--filesystem=host`, `--allow=devel`, `--device=all` and `--talk-name=org.freedesktop.Flatpak`, so the default install is effectively unsandboxed. `flatpak/README.md` says so plainly and the tightened override is verified end to end, which makes it defensible, but it should get a line in the release notes so nobody assumes the Flatpak is a security boundary out of the box.

Nice work, this is close.

Moyasee:
Separately, a few files in the diff that could come out or be split off.

### `electron-builder.yml` does nothing

The four added lines are a comment and nothing else. No flatpak target, no config, no behavior change:

```yaml
# The Flatpak is built from the hand-authored flatpak-builder manifest in
# flatpak/ (via `npm run build:flatpak`), not electron-builder's flatpak target,
# because electron-builder can't declare the i386 runtime extensions Proton's
# pressure-vessel needs in the sandbox. See flatpak/README.md.
```

Looks like residue from the earlier commit that did add a real flatpak target there before the switch to the custom manifest. `flatpak/README.md` already explains all of this, so I'd drop the file from the diff entirely.

### `binaries/7zzs` is not a Flatpak change

Not useless, but it does not belong in this PR:

```
main:    ELF 32-bit LSB executable, Intel i386, statically linked
this PR: ELF 64-bit LSB executable, x86-64, statically linked
```

`electron-builder.yml` ships `binaries/7zzs` as an extraResource for every Linux target and `src/main/services/7zip.ts` picks it as the Linux 7-Zip, so this swaps the archive-extraction binary for AppImage, deb, rpm, snap and tar.gz as well, not just the Flatpak. It is very likely the right fix on its own merits, but extraction is the path where a regression eats someone's finished download, and it will get close to zero scrutiny sitting in a 57-file Flatpak PR.

Can this go out as its own PR? If you would rather keep it here, can you confirm extraction was tested on a non-Flatpak Linux build (deb or AppImage) with the new binary?

### Nit

`PathGrants.removeGrant` is `public static` but the only caller is `listBroken()` in the same class. Can be private.

VictorVow:
Thanks for the re-review, glad the earlier round held up. Have addressed the feedback in 3 new commits and fixed a merge conflict. 

- `electron-builder.yml`: dropped the dead comment. Good catch, it was leftover from the earlier real-target attempt and `flatpak/README.md` already covers it.
- `PathGrants.removeGrant`: made it private, confirmed it's only ever called from `listBroken()` in the same class.
- `binaries/7zzs`:  keeping it in this PR rather than splitting it out. Built a local AppImage off this branch and confirmed the bundled `binaries/7zzs` in `dist/linux-unpacked/resources/` is the 64-bit x86_64 static build - ran the AppImage (non-Flatpak) and verified archive extraction works correctly with it, so the swap is exercised outside the sandbox too, not just inside the Flatpak.

On the sandbox default: fair point, that's worth calling out somewhere more visible than the README. I'd advise a message like this in the release notes:
> **Flatpak default is not sandboxed.** The Flatpak build ships with `--filesystem=host`, `--allow=devel`, `--device=all`, and `--talk-name=org.freedesktop.Flatpak`, so by default it has effectively the same filesystem/device access as running Hydra directly on the host. A tightened, portal-based sandbox override is documented and verified in `flatpak/README.md` for users who want it - don't assume the Flatpak is a security boundary out of the box.

VictorVow:
Dug into the 10 flagged vulnerabilities behind the "C Security Rating on New Code" gate - none of them are things this PR actually introduced.

**6 are in `.github/workflows/build.yml` / `release.yml`:**
- `yarn --frozen-lockfile` → `githubactions:S6505` ("lifecycle scripts enabled by default in Yarn v2+")
- `pip install -r requirements.txt` → `githubactions:S8541` / `S8544` ("missing `--only-binary :all:`" / "unlocked dependency versions")

`git blame` puts these lines at commit `0458a93b6` (2026-06-12), well before this branch. This PR only inserted new steps above them (the Flatpak tooling install step), which shifted their line numbers down - Sonar's new-code detection seems to be keying off the shift rather than actual blame.

**1 is in `src/locales/pt-BR/translation.json:1338`:**
```json
"toggle_password_visibility": "Alternar visibilidade da senha"
```
Flagged by `json:S2068` as a possible hard-coded credential - it's a UI label ("Toggle password visibility"), not a secret. `git blame` dates this line to 2024-06-03. This PR only added new lines near the top of the same file (the `lost_folder_access` strings), which pulled the whole file into the new-code scan.

So: nothing here is a regression from this branch, and none of it is fixable by changing this PR's code - it's either pre-existing findings on `main` that this PR happened to be the first to re-touch those files under, or a false-positive credential match. Could someone with SonarCloud project access mark these resolved (Won't Fix / False Positive), or is there a preferred way to get the gate to stop counting pre-existing lines as new here?

src/main/main.ts
```diff
@@ -162,25 +244,34 @@ export const loadState = async () => {
...
     await DownloadManager.startRPC(
-      downloadToResume ?? undefined,
+      resumableDownload ?? undefined,
       downloadsToSeed
     );
   }
 
   WindowManager.sendDownloadsUpdated();
 
+  await notifyLostPathGrants(lostGrantPaths);
```
Moyasee, line 273: `loadState()` runs before `createMainWindow()`, so this notification is sent while every app window is still null. `listBroken()` then deletes the grants, meaning the message is lost permanently. Buffer the broken paths until a renderer acknowledges them, or expose a renderer-initiated startup query and prune only after delivery.

python_rpc/main.py
```diff
@@ -87,7 +88,27 @@ def parse_cli_args(argv):
 downloading_game_id = -1
 current_download_limit = None
 
-torrent_session = lt.session(
+def create_torrent_session(settings):
+    # Under Flatpak, downloads may target document-portal FUSE paths where
+    # libtorrent 2.x's default mmap disk I/O is unsafe. Prefer the posix
+    # backend there; fall back to the default if this binding predates the
+    # posix_disk_io_constructor (older libtorrent).
+    if os.environ.get("FLATPAK_ID") and hasattr(lt, "posix_disk_io_constructor"):
```
Moyasee, line 96: This branch never activates with the libtorrent `2.0.13` Python binding installed by CI: neither `posix_disk_io_constructor` nor `session_params.disk_io_constructor` is exposed. Use the supported `mmap_file_size_cutoff` setting with a sufficiently high value, or ship and test a binding that actually exposes backend selection.

flatpak/gg.hydralauncher.hydra.yaml
```diff
@@ -0,0 +1,120 @@
...
+  # plus --filesystem=<library> if games live outside xdg-download. (Granting
+  # only xdg-download is enough to browse/download but NOT to launch Windows
+  # games: pressure-vessel must also reach the umu runtime and Proton build.)
+  - --filesystem=host
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.ScreenSaver
+  # Lets pressure-vessel ask the Flatpak portal to build its container as a
+  # sub-sandbox instead of a raw nested bwrap. Required direction for Proton in
+  # a sandbox; see flatpak/README.md for the bwrap fd limitation still open.
+  - --talk-name=org.freedesktop.Flatpak
```
Moyasee, line 57: `org.freedesktop.Flatpak` exposes `Development.HostCommand`, which upstream documents as a sandbox escape allowing arbitrary host process execution. Removing `--filesystem=host` therefore does not make the documented tightened mode isolated. Remove this permission and retest Proton, or explicitly document that tightened mode is not a security boundary.

flatpak/gg.hydralauncher.hydra.yaml
```diff
@@ -0,0 +1,120 @@
...
+  #     --filesystem=xdg-data/umu:create \
+  #     gg.hydralauncher.hydra
+  # plus --filesystem=<library> if games live outside xdg-download. (Granting
+  # only xdg-download is enough to browse/download but NOT to launch Windows
+  # games: pressure-vessel must also reach the umu runtime and Proton build.)
+  - --filesystem=host
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.ScreenSaver
+  # Lets pressure-vessel ask the Flatpak portal to build its container as a
+  # sub-sandbox instead of a raw nested bwrap. Required direction for Proton in
```
Moyasee, line 55: Electron's StatusNotifier implementation talks to `org.kde.StatusNotifierWatcher`, but the manifest does not grant that name. Add `--talk-name=org.kde.StatusNotifierWatcher`; Chromium registers the item using its unique connection, so no `StatusNotifierItem-*` ownership grant is needed.

src/main/main.ts
```diff
@@ -55,6 +106,12 @@ const hasMissingSeedFiles = async (download: Download): Promise<boolean> => {
 export const loadState = async () => {
   await Lock.acquireLock();
 
+  if (isFlatpak) {
+    // The statically-granted default downloads folder (xdg-download/Hydra)
+    // does not exist on first run; writability checks expect it to.
+    fs.mkdirSync(defaultDownloadsPath, { recursive: true });
```
Moyasee, line 112: `SystemPath.getPath()` can return an empty string, which turns this into a relative `Hydra` directory, and an unhandled `mkdirSync` failure aborts startup before a window exists. Validate that the path is nonempty and absolute, then handle directory-creation failure explicitly.

src/main/events/misc/show-open-dialog.ts
```diff
@@ -8,15 +8,31 @@ const showOpenDialog = async (
...
+
+  // Under Flatpak the portal file chooser may return document-portal FUSE
+  // paths; resolve their host paths so the UI can show a familiar location.
+  //
+  // We deliberately do NOT persist a grant here: most pickers are one-off
+  // (profile image, theme editor, emulator setup, memory cards, ...) and
+  // their doc-portal ids are ephemeral, so persisting every pick would grow
+  // the sublevel unbounded and resurface long-lapsed picks as false "folder
+  // access lost" toasts on later launches. Only callers that track a path
+  // long-term (downloads, wine prefix, proton, executable, backup) annotate.
```
Moyasee, line 30: The long-term annotation contract is not followed by several persisted paths, including custom/tracking executables, Wine prefixes, emulator/BIOS/ROM paths, classics discs, and some per-download destinations. Tightened mode will expose portal paths and cannot recover revoked grants for those features. Annotate at each main-process persistence boundary.

src/renderer/src/pages/game-details/cloud-sync-files-modal/cloud-sync-files-modal.tsx
```diff
@@ -45,7 +45,16 @@ export function CloudSyncFilesModal({
       setSelectedFileMappingMethod(FileMappingMethod.Automatic);
     }
 
-    setValue("customBackupPath", backupPreview?.customBackupPath ?? null);
+    const customBackupPath = backupPreview?.customBackupPath ?? null;
+
+    if (customBackupPath) {
+      globalThis.window.electron
+        .getDisplayPath(customBackupPath)
+        .then((displayPath) => setValue("customBackupPath", displayPath))
```
Moyasee, line 53: An older `getDisplayPath()` request can resolve after the preview, game, or modal visibility changes and overwrite the current path. Reuse the cancellation pattern from `useDisplayPath`.

.github/workflows/build.yml
```diff
@@ -46,10 +46,17 @@ jobs:
...
+        if: matrix.os == 'ubuntu-latest'
+        uses: ./.github/actions/install-flatpak-tooling
+        with:
+          runtime-version: "25.08"
+
       - name: Build Linux
         if: matrix.os == 'ubuntu-latest'
         run: |
           yarn build:linux
+          node scripts/build-flatpak.cjs
```
Moyasee, line 59: PR CI builds `dist/*.flatpak` but does not retain it in the Actions artifact, and fork builds skip the external upload when secrets are unavailable. Include the Flatpak in the artifact glob so reviewers can install the package produced by CI.

Moyasee (changes requested):
The Flatpak builds, but three central runtime behaviors are currently nonfunctional: lost grants are deleted before the renderer can be notified, the libtorrent backend selection never activates with the installed Python binding, and the documented tightened mode still exposes Flatpak's host-command sandbox escape. The tray permission and persisted-path coverage also need to be addressed before this is ready to ship.

Hachi-R:
Closing this because three core Flatpak flows are still non-functional, so this cannot ship in its current form. The packaging and runtime integration need a redesign with the main flows validated end to end. Feel free to reopen the PR after those issues are fixed.

PR #2411, opened 2026-06-26
feat: self-hosted API support, dual accounts, cloud save diff restore

entitybtw:

**Fill in the PR content:**

This PR adds self-hosted API support, allowing users to run their own backend instead of relying solely on Hydra Cloud.

### Self-hosted API
- New settings page (Settings → Self-Hosted) to enter a server URL and instance token
- Supports cloud saves, achievements, and user accounts on a self-hosted instance
- When connected, subscription is always considered active (no paywall on self-hosted)
- Reference server implementation: https://github.com/entitybtw/hydra-selfhosted

### Dual accounts
- Users can be logged into both an official Hydra account and a self-hosted account simultaneously
- Sidebar shows both profiles; the official account is labeled "Official"
- Edit Profile and Upload Banner correctly target the respective account

### Cloud save restore improvements
- After restoring a backup, files with matching extensions in the restored directories that are not present in the backup are deleted
- Prevents stale save slots or old level files from persisting after a restore


chubbygrannychaser:
Sorry, I’m not comfortable merging a PR that points to an external repository maintained outside our organization, especially for something as sensitive as a self-hosted API.

We’re responsible for how user data is handled, and our backend goes through serious audits and operational controls. For that reason, we can’t delegate or endorse an unofficial implementation that we don’t maintain or review internally.

I appreciate the contribution, but this isn’t something we plan to support.

PR #2413, opened 2026-06-27
feat: add RetroAchievements account connection in settings integrations

Moyasee:

**Fill in the PR content:**

Closes LBX-840.

Adds a RetroAchievements section under Settings > Integrations so users can link their RA account from Hydra. The backend already owns the credentials and syncing, so this is purely the connection UI wired to the existing endpoints.

What it does:
- Fetches the current status from `GET /profile/integrations/retroachievements` on mount.
- Disconnected state shows username, password and Web API key inputs with a connect button, and submits to `POST /profile/integrations/retroachievements/connect`.
- Connected state shows the linked RA username, a status badge (active / invalid credentials), and a disconnect button hitting `DELETE /profile/integrations/retroachievements`.
- Stores the Web API key locally only after a successful connect (same persistence used for debrid tokens) and clears it on disconnect. The password is never persisted.

Also reordered the error extraction in `hydra-api-call.ts` to prefer the backend response message over the generic Axios message, so specific connect errors (invalid password vs invalid Web API key) reach the toast. The backend returns 422 for these, so a wrong RA credential no longer risks signing the user out.

Translations added for en, pt-BR, pt-PT, ru and es.


Moyasee:
Addressed the feedback: the UI now flips to the connected/disconnected state right after the backend call succeeds, with the local key persistence as a best-effort follow-up, and pulled the icon size into a constant.

@greptile review

Hachi-R (changes requested):
I think this needs one fix before merging.

When a LaunchBox game has a RetroAchievements mapping and the user has a stored RA key, `fetchGameDetails` switches to `getRetroAchievementsAchievements`, but it only calls `setAchievements` when the result is truthy and silently ignores errors.

If the RA request fails, credentials are stale, the username/key is missing locally, or the sync returns `null`, the achievements page stays on the skeleton forever because `achievements` remains `null`.

Could we fall back to `getUnlockedAchievements` when the RA sync returns `null` or throws, or at least set an empty/cached achievements state so the page can finish loading?

Moyasee:
Good catch @hachi-r. The game page now falls back to `getUnlockedAchievements` whenever the direct RA fetch returns null or throws (missing/stale credentials, no mapping, or an API error), so the achievements page always finishes loading instead of hanging on the skeleton. Pushed in b5522854.

Moyasee:
Fair point on the fallback source. Updated it in d1e147ad: when the direct RA fetch fails or is unavailable, it now returns the locally cached achievements from LevelDB instead of hitting our API (getUnlockedAchievements), since the API achievement data is for profiles/Cloud only. The game page still resolves and stops loading, just from local state now.

Hachi-R (approved):
LGTM

Moyasee:
Fixed the achievement translations. The RA achievements box was showing RetroAchievements' text directly, and their API is English-only, so it never followed the app language. Now the RA path pulls the achievement catalogue (names/descriptions/icons) from our language-aware achievements endpoint (getGameAchievementData, same one the non-RA games use) and just overlays the RetroAchievements unlock state on top. So the box respects the language and refreshes on language change like the game description, with RetroAchievements still the source of unlock progress.

PR #2419, opened 2026-06-29
fix: detect PCSX2/PS1 BIOS in user-configured and OneDrive folders

Moyasee:

**Fill in the PR content:**

Fixes the case where a user has a valid PS2 BIOS that PCSX2 itself recognizes, but Hydra reports `BIOS_NOT_CONFIGURED` and refuses to launch (LBX-845).

The detection only read the `[Folders] Bios=` override from the first config it found in a fixed priority order (Documents, then AppData, then the portable config next to the executable). When more than one config existed, the user's actual config (usually the portable one Hydra launches) got shadowed, so the custom BIOS directory was never scanned. There was also no logging, so the tester logs gave us nothing to work with.

What changed:

- Read the `[Folders] Bios=` / `SearchDirectory=` override from every config that exists and merge them, looking at the portable config beside the executable first.
- Add OneDrive-redirected Documents locations to the config search so OneDrive Known Folder Move setups resolve.
- Add a manual BIOS folder picker in the setup wizard and on the emulator page for when auto-detection still misses, and show the detected folder in the path field.
- Log the directories scanned, per-file reject reasons (size vs signature), and read failures (flagging OneDrive paths), so future reports are diagnosable.

Closes LBX-845.


Moyasee:
Extracted to a named constant, thanks.

@greptile review

Hachi-R (changes requested):
I don't think this is ready yet.

The BIOS detection changes look good overall, but this PR also removes `-batch` from the DuckStation and PCSX2 launch args. That seems unrelated to BIOS detection and changes the launch behavior for every PS1/PS2 classic, so I think we should either keep `-batch` or explain why removing it is required and confirm it was tested.

Also, the manual BIOS path IPC still calls `statSync` after `existsSync` without handling errors. If `statSync` throws on a restricted path / junction / network path, the picker flow fails through IPC. Could we wrap that in a safe `try/catch` or use async `stat().catch(() => null)` before saving the path?

GitHub also marks the PR as not mergeable right now, so it likely needs a rebase/conflict fix before merge.

Moyasee:
Thanks for the review, addressed all three:

- Reverted the -batch removal. You're right that it's unrelated to BIOS detection and changes behavior for every PS1/PS2 launch. PCSX2 and DuckStation are back to `-batch -fullscreen`, so this PR is now just the BIOS detection + the emulator settings UI. Since Hydra already blocks the launch and shows a toast when no BIOS is found anywhere, we don't lose much by keeping -batch.
- Wrapped the manual BIOS path check in `fs.stat().catch(() => null)` instead of `existsSync` + `statSync`, so a restricted path / junction / network path can't throw through IPC.
- Merged main and resolved the conflicts (launch args + a couple of locale strings), so it should be mergeable now.

The BIOS resolution + auto-correct now lives inside the `assertBiosInstalled` helper that landed on main.

Hachi-R (approved):
thanks, this addresses my concerns.

`-batch` is back for DuckStation/PCSX2, and the BIOS path IPC now uses async `stat().catch(() => null)` instead of `existsSync + statSync`, so the picker path should no longer fail through an unhandled sync exception.

The PR is mergeable again too. I'll just wait for the build check to finish before approving/merging.

PR #2420, opened 2026-06-30
fix: make library scan button clickable on linux

Moyasee:

**Fill in the PR content:**

The scan games button in the library header was not clickable on Linux.

The header is a drag region (`-webkit-app-region: drag`), and the scan button is a bare `header__action-button` that sits outside the `header__search` wrapper, so it inherited the drag region and the OS swallowed its clicks. The search box and back button already opt out with `no-drag`, and on Windows the whole header is flipped to `no-drag`, which is why this only reproduced on Linux.

Fix marks `header__action-button` as `-webkit-app-region: no-drag` so the button receives clicks. Verified locally on Linux: the scan modal now opens on click.


Hachi-R (approved):
LGTM

PR #2421, opened 2026-06-30
fix: launch PS3 .pkg games via RPCS3 install + boot

Moyasee:
Closes LBX-846.

**Fill in the PR content:**

PS3 `.pkg` games failed to boot through Hydra while working fine when launched straight from RPCS3. A tester saw an `Invalid file or folder` fatal error from RPCS3 (God of War HD, NPEA00255).

The cause is that a `.pkg` is an installable package, not a bootable image. We were handing the raw `.pkg` to `rpcs3 --no-gui`, which RPCS3 can't boot. The reason it works inside RPCS3 is that the user installed the package there first.

What this changes for a PS3 `.pkg` launch:

- Pull the title id from the pkg and check whether it's already installed under `dev_hdd0/game/<titleId>/USRDIR/EBOOT.BIN`. If so, boot that EBOOT headless.
- If it isn't installed, run `rpcs3 --installpkg` (RPCS3 refuses to install in no-gui mode) and surface a toast telling the user to hit Play again once the install finishes. The second launch then boots headless.

Adds a `pkg_installing_toast` string (en, pt-BR, pt-PT, ru, es) and threads a new `PKG_INSTALLING` launch code through the renderer and big-picture handlers.

Note: I couldn't run this against a real Windows + RPCS3 + pkg setup, so a manual smoke test before merge would be good. There's also a separate stale-disc-path `NO_DISC` in the same logs (an old `D:\` import path) that I left out of scope here.


Moyasee:
Addressed both:

- Added an `error` listener to the `--installpkg` child so a failed spawn can't take down the main process.
- When the title id can't be read from the pkg, it now throws a distinct `PKG_UNREADABLE` instead of re-triggering the install on every Play, with a toast telling the user to install/launch from RPCS3 directly.

Also split the pkg/BIOS/wrapper logic into helpers to drop the SonarCloud cognitive complexity flag on `launchClassicsGame`.

@greptile review

Hachi-R (approved):
LGTM, thanks for fixing this.

I reviewed the PKG install/boot flow and didn't find any blocker. The unresolved edge cases from the earlier review look handled now: unreadable PKGs return a proper error, and the RPCS3 install spawn has an error listener. The renderer also handles the new launch codes correctly.

PR #2422, opened 2026-07-01
fix shop-details locale handling

Hachi-R:

**Fill in the PR content:**


src/main/events/catalogue/get-game-shop-details.ts
```diff
@@ -51,50 +51,54 @@ interface LaunchboxShopDetailsData {
...
+  const description = data?.description ?? cachedData?.about_the_game ?? "";
 
   return {
     objectId,
-    name: data?.title ?? basic?.title ?? "",
-    platform: entry?.platform ?? data?.platform ?? undefined,
-    skus: entry?.skus ?? undefined,
+    name: data?.title ?? basic?.title ?? cachedData?.name ?? "",
+    platform: data?.platform ?? cachedData?.platform ?? undefined,
+    skus: cachedData?.skus ?? undefined,
```
Moyasee: skus is only read from cachedData here, and the new LaunchboxShopDetailsData type (line 40) has no skus field, so a fresh fetch never populates it. On a cold cache (likely now that both the cache key and the endpoint changed) launchbox skus will be undefined, which means the region info in big-picture game.tsx (~443) and the sidebar launchbox section (sidebar.tsx:310) won't render. Was skus intentionally dropped from this endpoint? If the new /shop-details still returns it, worth adding skus to the type and reading data.skus here.

Moyasee (commented):
Reviewed the shop-details locale handling changes. Overall this looks solid: moving the language normalization into main, parallelizing the fetches, and the defensive optional chaining on content_descriptors plus the .catch on getGameShopDetails are all good. Heads up that this PR has no locale JSON changes, so there is nothing to check on translation key parity.

One thing I would like confirmed inline (skus on launchbox). A few nits below.

- getSteamLanguage is now unused after this PR (src/renderer/src/helpers.ts:46 and src/big-picture/src/helpers/language.ts:1). Its mapping was inlined into getLocalizedSteamAppDetails, so both copies are dead code and can be removed.
- Minor: for steam the language gets normalized ("brazilian" -> "pt-BR") and then mapped straight back ("pt-BR" -> "brazilian") in getLocalizedSteamAppDetails. Works, just a bit roundabout, and it duplicates the getSteamLanguage table. A single shared helper would avoid the double hop.
- Behavior check, not blocking: getLaunchboxShopDetails lost its cache fast-path (the old cacheHasNewFields early return), so launchbox now always makes both network calls and only falls back to cache on failure. Steam still returns cache immediately and refreshes in the background. Probably intentional since the old fast-path served stale/wrong-language data, but flagging in case you want to mirror the steam pattern for perceived perf.

Hachi-R:
@greptile

galinhaman22 (commented):
LGTM

galinhaman22 (approved):
LGTM :)

PR #2437, opened 2026-07-05
Sync per-game launch options to Steam shortcuts

Jahbas:
When creating a Steam shortcut for a game, the launch options set in Hydra get ignored - the LaunchOptions field is always empty. This PR passes the user's launch options through to Steam so they actually apply when launching via the shortcut.

Also updates the LaunchOptions on existing Steam shortcuts when the user changes them, instead of only applying on creation.

3 files changed:
- src/main/services/steam.ts - added launchOptions param to composeSteamShortcut + updateSteamShortcutLaunchOptions helper
- src/main/events/library/create-steam-shortcut.ts - passes game.launchOptions through
- src/main/events/library/update-launch-options.ts - also syncs to existing shortcut if one exists


Moyasee (approved):
Core change looks clean. Verified the services barrel re-exports `updateSteamShortcutLaunchOptions`, `steamShortcutAppId` is typed as number so the `s.appid === steamShortcutAppId` lookup matches, and the trim handling is consistent between `composeSteamShortcut` and the update helper.

One heads-up that isn't caused by this PR but affects your update-sync path: in `create-steam-shortcut.ts` the Linux block runs a second `gamesSublevel.put` that spreads the original `game` and sets `winePrefixPath` (around line 179), which overwrites the `steamShortcutAppId` you persist a few lines above (line 156). So for a Linux game without an existing wine prefix, `steamShortcutAppId` never sticks, and `updateLaunchOptions`'s `if (game.steamShortcutAppId)` guard ends up false, so the sync silently no-ops there. Might be worth merging those two puts into one so the feature works on Linux as well.

Minor: the `.catch(() => {})` swallows write failures with no trace. A `logger.error` in there would help future debugging.

Neither blocks what the PR does, so approving.

Moyasee:
Consolidating my review with greptile's so it's in one place. Split by what I'd want before merge vs optional polish:

**Worth addressing before merge**
- Linux launch-options sync is silently dead. In `create-steam-shortcut.ts` the Linux block runs a second `gamesSublevel.put` (around line 179) that spreads the original `game` and only sets `winePrefixPath`, which overwrites the `steamShortcutAppId` persisted at line 156. So on Linux without an existing wine prefix the appId never sticks, and `updateLaunchOptions`'s `if (game.steamShortcutAppId)` guard is false, so your new sync never runs there. Folding those two puts into one fixes it. It's pre-existing code, not something this PR introduced, but the feature depends on it so it's cleanest to handle here.
- greptile's P1 on the empty `.catch(() => {})`: worth a `logger.error` so a locked or malformed `shortcuts.vdf` doesn't fail completely silently. Cheap and saves future debugging.

**Optional, not blocking**
- greptile's double-trim simplification on `steam.ts:162` (`launchOptions?.trim() ?? ""`). It's equivalent, just tidier.
- greptile's note about passing the raw `launchOptions` to the sync helper instead of the normalized value. No behavior change, consistency only.

Windows path works as-is. The Linux gap is the one I'd not want shipped quietly.

Jahbas:
hi would be great if someone could review and approve since greptile isn't synced

0311z:
@greptileai review

PR #2439, opened 2026-07-06
feat: replace gray screen with error screen on renderer crashes

Moyasee:

**Fill in the PR content:**

## What

Right now an unhandled render error on the game page (and any other page) unmounts the whole React tree and leaves a blank gray screen, so on 4.0.4 we have no idea what actually broke.

This wraps the app in a top-level error boundary. When a render throws, we show the real error instead of gray:

- error message
- origin file and line (first app frame pulled from the stack)
- full stack
- collapsible React component trace
- copy button
- a restart button that sends the user back to the home page and reloads. The old reload kept the same route, so it just landed on the crashing page and threw again.

It also adds a small non-blocking overlay plus `window` error and unhandledrejection listeners for async and event-handler errors a boundary can't catch, and logs everything through electron-log.

To keep production/OTA stack traces readable without shipping sourcemaps, `keepNames` is enabled for the renderer build so component and function names survive minification.

## Notes

- Everything here is renderer-only, so it is OTA-friendly and can go out on 4.0.4.
- No Sentry sourcemap upload on purpose. Errors are surfaced locally and written to the log file.
- New translation keys added for en, pt-BR, pt-PT, ru and es.

## How to test

Temporarily throw inside a page render (for example in game-details), open that page and confirm the error screen shows the message and stack instead of gray. Run `Promise.reject(new Error("test"))` in the devtools console to see the async overlay card.

### DIFF
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -56,6 +56,9 @@ export default defineConfig(({ mode }) => {
       build: {
         sourcemap: true,
       },
+      esbuild: {
+        keepNames: true,
+      },
       css: {
         postcss: {
           plugins: [scopeBigPictureCss()],
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -1,7 +1,17 @@
 {
   "language_name": "English",
   "app": {
-    "successfully_signed_in": "Successfully signed in"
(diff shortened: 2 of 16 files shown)


Moyasee:
@greptile review

chubbygrannychaser (approved):
LGTM

PR #2440, opened 2026-07-06
feat(achievements): add sorting button

Wiwok:

**Fill in the PR content:**
This PR adds a sort button to the Achievements page, allowing users to sort the list. The available sorting options are still open for discussion, but I think adding sorting support is a useful improvement for usability. Feedback on which sorting options should be included is welcome.


src/renderer/src/pages/achievements/achievement-list.tsx
```diff
@@ -6,20 +6,42 @@ import { EyeClosedIcon } from "@primer/octicons-react";
...
   const { showHydraCloudModal } = useSubscription();
   const { formatDateTime } = useDate();
 
+  function sorter(a: UserAchievement, b: UserAchievement) {
+    let diff = 0;
+
+    if (sort === "date") {
+      diff = Number(b.unlockTime) - Number(a.unlockTime);
+    } else if (sort === "points") {
+      diff = Number(a.points) - Number(b.points);
```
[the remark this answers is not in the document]
Wiwok: I think users would prefer to see the least valuable achievements first. Maybe we could rename this to "Sort by rarity"?

Moyasee:
Hey! Could you please share screenshots on how this looks inside of launcher?

Wiwok:
It's a reuse of the FilterDropdown component defined in the profile content section.
It's only available in French and English for now, and I'm not entirely sure about the filter options. Feel free to change them!

Screen 1
Screen 2

Moyasee (commented):
Nice addition. A few notes:

- i18n: keys added only to `en` and `fr`. The other 30+ locales (pt-BR, pt-PT, ru, es, ...) don't have `sort_by`/`sort_option_*`, so they'll fall back to English. Worth adding at least pt-BR/pt-PT/ru/es.
- `types/index.ts`: the union reformatting on `HydraCloudFeature` and `protondbSupportBadges` is unrelated churn (and `yarn format` will likely flip it back). Better to drop it so the diff stays focused on the feature.
- `FilterDropdown` is imported from `pages/profile/profile-content/`. Now that it's used on two pages, might be worth promoting it to a shared component instead of reaching across pages.
- Minor: points sort is ascending (`getPoints(a) - getPoints(b)`), so lowest points first. Just confirm that's the intended default direction.

Wiwok:
> * Minor: points sort is ascending (`getPoints(a) - getPoints(b)`), so lowest points first. Just confirm that's the intended default direction.

Yes, it is. Greptile already pointed this out. Maybe we should call it "rarity" instead of "points." Still, I prefer showing the easier achievements first, so the ascending order is intentional.

Wiwok:
I've added translations with the help of AI, reusing existing translations from the files. Feel free to change things if I got anything wrong!

Moyasee:
Thanks for the updates, most of the earlier notes are handled now. One thing left: `compared-achievement-list.tsx` still uses `key={index}` after `.toSorted()`. Since the order changes on sort, index keys make React reuse DOM nodes across different achievements and you can get stale rows. `achievement-list.tsx` already keys on `achievement.name`, so keying the compared list on a stable field instead of the index would keep them consistent.

Rest looks good to me.

Wiwok:
Since ComparedAchievement comes directly from the API, I can't add a name/id property to it. I'm keeping the type definition as-is and using icon + displayName as the key instead.
It's not a perfect unique identifier, but it's the best option available on this object - and it's still more stable than using the array index.

Moyasee (commented):
Nice addition. A few things before this goes in:

- The rarity sort looks inverted. `getPoints(a) - getPoints(b)` sorts ascending, so the low-point (common) achievements come first and the rare ones sink to the bottom. I would expect rarest first, so `getPoints(b) - getPoints(a)` seems closer to intent. Worth keeping in mind that points is only a proxy for rarity, and non-subscribers have no points, so it quietly degrades to a name sort for them.

- The sort state is shared between the compared and non-compared views. Since rarity is stripped out of the compared options, picking rarity and then opening a comparison leaves the dropdown holding a value that is not in the list. Not a crash, it just falls back to the name tiebreak, but resetting sort when switching into comparison would be cleaner.

Minor: SortDescIcon is reused for both Default and Name.

src/renderer/src/pages/achievements/achievements-sorter.ts
```diff
@@ -0,0 +1,39 @@
...
+      ? Number(achievement.unlockTime)
+      : Number(achievement.targetStat.unlockTime);
+
+  const getPoints = (achievement: UserAchievement | ComparedAchievement) =>
+    isUserAchievement(achievement) ? Number(achievement.points ?? 0) : 0;
+
+  if (sort === "date") {
+    diff = getUnlockTime(b) - getUnlockTime(a);
+  } else if (sort === "points") {
+    diff = getPoints(a) - getPoints(b);
```
Moyasee, line 31: This is still ascending, so the cheapest achievements come first. Renaming the label from Rarity to Points makes it less misleading, but a control labelled "Points" that puts the 5 point achievements above the 100 point ones is going to read as broken to most people. I would flip it to `getPoints(b) - getPoints(a)`.

If you feel strongly that easiest-first is the better default that is fine, but then the label should say so rather than just "Points".

src/renderer/src/pages/achievements/achievements-content.tsx
```diff
@@ -255,13 +300,38 @@ export function AchievementsContent({
...
         ) : (
           <>
-            <AchievementPanel achievements={achievements!} />
-            <AchievementList achievements={achievements!} />
+            <div className="achievements-content__panel-container">
+              <div>
+                <AchievementPanel achievements={achievements!} />
+              </div>
+              <FilterDropdown
+                options={sortOptions}
```
Moyasee: `sortOptions` is passed unfiltered here, so users without an active subscription still get the Points option. They have no points on any achievement, which is why the list renders `???` instead of a value, so `getPoints` returns 0 for every row and the comparator falls straight through to the name tiebreak. Picking Points then does exactly the same thing as picking Name, silently.

You already have `hasActiveSubscription` in scope on line 162, and you are already filtering the option out for the comparison view on line 223, so the same treatment here would avoid handing people a control that does nothing.

src/renderer/src/pages/achievements/achievements-sorter.ts
```diff
@@ -0,0 +1,39 @@
...
+  a: UserAchievement | ComparedAchievement,
+  b: UserAchievement | ComparedAchievement,
+  sort?: AchievementSort
+) {
+  let diff = 0;
+
+  const getUnlockTime = (achievement: UserAchievement | ComparedAchievement) =>
+    isUserAchievement(achievement)
+      ? Number(achievement.unlockTime)
+      : Number(achievement.targetStat.unlockTime);
```
Moyasee: In comparison mode this sorts on `targetStat.unlockTime`, which is the other user's unlock time rather than yours. So "Last unlocked" on the comparison view orders the list by their timeline while you are looking at your own page.

I can see the reasoning since `ownerStat` is optional, but something like `achievement.ownerStat?.unlockTime ?? achievement.targetStat.unlockTime` would match what the label implies. Worth a sanity check on which one you actually meant.

src/renderer/src/pages/achievements/achievements-content.tsx
```diff
@@ -179,6 +201,29 @@ export function AchievementsContent({
     );
   };
 
+  const sortOptions = [
```
Moyasee: `sortOptions` is untyped, so `value` widens to `string` and `T` on `FilterDropdown` resolves to `string`, which is why the `as AchievementSort` casts are needed on lines 313 and 331. Typing this as `FilterDropdownOption<AchievementSort>[]` drops both casts and matches how `library-tab.tsx` declares its own options.

Moyasee (changes requested):
Thanks for the follow-up, this is in much better shape. The sort keys are in all six locales now, the unrelated `types/index.ts` churn is gone, `FilterDropdown` is properly promoted into `components/`, the compared list keys on a stable field instead of the index, Default and Name have distinct icons, and the sort resets when you enter comparison mode. CI is green.

The two points I raised last time about the points sort are still open though. Renaming the label from Rarity to Points addressed the wording but not the behaviour: the order is still lowest-first, and the option is still offered to people who have no points at all. Both inline.

Two smaller notes as well, about the date sort in comparison mode and the option typing.

src/renderer/src/pages/achievements/achievements-content.tsx
Wiwok: In practice, `achievement.points` is populated per-achievement even without an active subscription - I just checked on my end (I don't have an active subscription, so I can't verify the subscribed case), and the only place the `???` shows up is the aggregated total at the top of the page, not the individual value for each achievement in this list. The `???` fallback in `AchievementList` is defensive (in case the API doesn't return points for some reason), but it doesn't correspond to a case I could reproduce as a non-subscribed user.

So `getPoints` doesn't fall back to 0 for everyone like you're suggesting - sorting by difficulty works correctly even without Hydra Cloud, and I'd rather keep the option available to everyone than gate it behind `hasActiveSubscription`. It's the total at the top that's reserved for subscribers, not the sort. And `getPoints` already handles the `undefined` case gracefully with `points ?? 0`, so there's no gap there either.

Moyasee (approved):
You are right about the points and I was wrong, sorry for the noise. I went and checked the API: `get-achievements` returns `points` on every achievement with no subscription check anywhere in the handler. What I had latched onto was the `??? / ???` in `AchievementPanel`, which is hardcoded behind `!hasActiveSubscription`, and I wrongly assumed the per-achievement `points != undefined` branch in `AchievementList` was the same gate. It is not, and as you say it does not fire in practice. So the sort works fine for everyone and gating it would have been wrong. Retracting that one.

Everything else looks good now:

- `easiest_first` is a much better answer than flipping the order. Points come from rarity, so ascending genuinely is easiest-first and the label finally says what the code does. Real translations in all six locales too, not copies.
- The comparison date sort now falls back through `ownerStat?.unlockTime ?? targetStat.unlockTime`, so it orders by your timeline rather than theirs.
- `sortOptions` is typed as `FilterDropdownOption<AchievementSort>[]`.

One leftover, genuinely not worth another round: the `as AchievementSort` casts on lines 316 and 334 are redundant now that the options are typed, since `val` already comes through as `AchievementSort`. Harmless, feel free to leave them or clean them up whenever.

Good work on this, thanks for pushing back where I had it wrong.

Wiwok:
No worries - just fixed that little thing :)

Hachi-R (changes requested):
`src/renderer/src/pages/achievements/achievements-content.tsx:L306` — P2: `ComparedAchievementPanel` now shares the row width with the sort dropdown, so its `3fr 1fr 1fr` grid no longer aligns the point totals with the comparison list columns. Keep the comparison panel full-width and position the sort control without consuming its grid width.

`src/locales/fr/translation.json:L1486` — P3: `Défaut` means “defect,” not the default option in this context. Use `Par défaut`.

Wiwok:
> `src/renderer/src/pages/achievements/achievements-content.tsx:L306` — P2: `ComparedAchievementPanel` now shares the row width with the sort dropdown, so its `3fr 1fr 1fr` grid no longer aligns the point totals with the comparison list columns. Keep the comparison panel full-width and position the sort control without consuming its grid width.

Fixed. I don't have an Hydra Cloud subscription so I couldn't verify the result visually — a second look would be appreciated before merging.

> `src/locales/fr/translation.json:L1486` — P3: `Défaut` means “defect,” not the default option in this context. Use `Par défaut`.

Fixed: Défaut → Par défaut. This introduced a duplication with the sort_by prefix ("Trier par" → "Trier par par défaut"). So I harmonized all the related translations to follow a new convention: sort_by becomes "Trier", and each option now carries its own "Par..." (e.g. "Trier : Par défaut", "Trier : Par date", "Trier : Par nom"). This removes the duplication and keeps the wording consistent across all options.

Moyasee:
Closing this one since this carries too much changes and unrelated to the PR files. 

Feel free to reopen this.

PR #2441, opened 2026-07-06
Fix: library shadow gap

Wiwok:

**Fill in the PR content:**
This fixes a visual regression introduced in PR #2410, where the shadow is offset by 4px from the top of the library box.


Moyasee:
Hey! Could you also share screenshots of this as well? Before/After

Wiwok:
Yep of course !
Here it is. There's just a 4px gap in the shadow when scrolling down in the library.

Before:
Before

After:
After

Moyasee (commented):
LGTM. Traced it through: the sticky shadow anchors to the padding edge of `__games-scroll`, and the `padding: 4px` that was added there later (the card-hover fix) is what pushed the shadow down and opened the 4px gap. `top: -4px` compensates exactly, so this is the right fix.

Only nit, non-blocking: that `-4px` is coupled to the `padding: 4px` a few lines up. If the padding ever changes the gap comes back. A shared SCSS variable used in both spots would make it self-documenting, but it's fine to ship as is.

Moyasee (approved):
lgtm

PR #2442, opened 2026-07-06
style: soften error screen colors and make the log monospace

Moyasee:

**Fill in the PR content:**

## What

Follow up to the error screen from #2439 based on review feedback.

The red and yellow accents read as a bit overkill, so this swaps them for muted neutral tones on both the fullscreen error screen and the async error overlay. It also sets an explicit monospace font on the message, origin and stack log so the trace reads as code instead of falling back to the body font.

Styling only, no behavior change.


Hachi-R (approved):
LGTM

PR #2451, opened 2026-07-06
fix(catalogue): guard RepacksModal against undefined GameRepack.uris

sotik11:
## Summary

`getRepackAvailabilityStatus` in `RepacksModal` assumes `repack.uris` is always a `string[]`, but the field can arrive `undefined` for records with incomplete backend data (stale cache, missing shop assets, etc.). The subsequent `.filter()` throws `TypeError: Cannot read properties of undefined (reading 'filter')`, which the new v4.0.4 error boundary catches and surfaces as a full-page overlay — so any affected game page becomes unusable.

## Fix

Guard with `Array.isArray` so a broken record degrades to `"offline"` instead of crashing the whole page.

```ts
const uris = Array.isArray(repack.uris) ? repack.uris : [];
```

## Reproduction

Open a game details page whose repack list contains an entry without `uris`. Stack:

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at getRepackAvailabilityStatus (repacks-modal.tsx:151)
    at RepacksModal.map (repacks-modal.tsx:339)
    at GameDetailsContextProvider
```

## Test plan

- [x] Local build (`yarn build:win`) — page opens instead of crashing when the repack list has an entry without `uris`.
- [ ] Normal repacks (with populated `uris`) still show correct availability status (online / partial / offline).


Moyasee (commented):
Looks correct. The `Array.isArray` guard handles the undefined case, and the added `if (uris.length === 0) return "offline"` is the important bit. Without it an empty list would fall through to `unavailableCount === 0` and wrongly report "online", so good that it short-circuits. Non-empty behavior is unchanged. Nothing to change from my side.

Moyasee (dismissed):
LGTM

Moyasee:
One small thing before this goes in: could you drop the inline code comment above the guard? We try to keep the code comment-free, and the `Array.isArray` check reads clearly on its own. The PR description already captures the why. Logic itself is good.

Moyasee:
Guard is incomplete. `checkIfLastDownloadedOption` still does `repack.uris.some(...)` without a guard, and it runs before `getRepackAvailabilityStatus` in the render map. So any previously-downloaded game (`game.download` set) with a repack missing `uris` still throws the same TypeError and trips the error boundary, which is exactly what this PR is trying to prevent.

Same guard there should do it:

```ts
const checkIfLastDownloadedOption = (repack: GameRepack) => {
  if (!game?.download) return false;
  if (!Array.isArray(repack.uris)) return false;
  return repack.uris.some((uri) => uri.includes(game.download!.uri));
};
```

The empty-array short-circuit in `getRepackAvailabilityStatus` is good. Just need both call sites covered since the field can arrive undefined.

sotik11:
Fixed in 7a51c97 — applied your suggestion as-is.

While checking for other spots, I found three more unguarded `repack.uris` accesses that are reachable through the same scenario (a broken repack row is still clickable — the guard only renders it as "offline" — and clicking it opens `DownloadSettingsModal`):

- `download-settings-modal.tsx:307` — `selectedUri` memo, `repack.uris.find(...)`
- `download-settings-modal.tsx:353` — `downloadOptions` memo, `for (const uri of repack.uris)`
- `game-details.tsx:76` — `selectRepackUri`, `repack.uris.find(...)!` (also a non-null assertion on the result)

Want me to cover them here, or keep this PR scoped to `RepacksModal` and send a follow-up?

Moyasee (approved):
LGTM

PR #2452, opened 2026-07-06
fix(catalogue): guard steamUserTags lookup for locales missing from the JSON

sotik11:
## What

Guards the Steam-user-tags lookup in the catalogue against locales that aren't present in `steam-user-tags.json`.

## The crash (not CDN-related)

`steam-user-tags.json` only ships **4 locales**: `en`, `es`, `pt`, `ru`. The app ships ~30 locales and looks tags up by `i18n.language.split("-")[0]`. So for **de, fr, pl, uk** and most other languages `steamUserTags[language]` is `undefined` on a completely healthy load.

**Repro:** tick a tag filter, then switch the language to German → `Object.keys(steamUserTags[language])` throws during render. The error boundary sits above the router (`main.tsx`), so the whole app drops to the fallback, not just the catalogue page:

```
TypeError: Cannot convert undefined or null to object
    at Object.keys (<anonymous>)
    at .../catalogue.tsx (tag chip label lookup)
```

## Fix

Guard the lookup with `steamUserTags[language] ?? {}`. The same steps then render the catalogue in German with no crash.

> The earlier external-resources interceptor was dropped per review — it didn't change behaviour (failures already log via the `unhandledrejection` handler in `main.tsx`, and nothing dispatches on failure so previously loaded data is kept regardless). This leaves the one-file, load-bearing crash fix.


Moyasee (commented):
The approach is good, and I like that it only swallows genuine network/timeout failures and re-throws real HTTP errors so misconfig stays visible. Verified `externalResourcesInstance` is only used for the four array-returning fetches, so the `{ data: [] }` fallback is safe everywhere.

Two things worth addressing before this goes in:

1. Use the renderer logger instead of `console.warn`. In `main.tsx` only `console.log` is reassigned to the logger; `console.warn` isn't, so this line only shows up in devtools and never makes it into the persisted renderer log. That defeats the purpose here, since a CDN outage in the field would leave no trace. The rest of the renderer (including `use-search-suggestions.ts` in this same area) uses `@renderer/logger`:

```ts
import { logger } from "@renderer/logger";
...
logger.warn("[external-resources] request failed silently:", error?.message ?? error);
```

2. Minor: the `config: error.config ?? ({} as any)` cast. If it's just to satisfy the AxiosResponse shape, `error.config` is usually enough, or type the fallback rather than `as any`. Not blocking.

Rest is solid.

Moyasee:
One more, same as on 2451: could you drop the 8-line comment block above the interceptor? We keep the code comment-free. That rationale is great context but it belongs in the PR description rather than inline. The code reads fine on its own, especially once the logger swap is in.

Moyasee (approved):
LGTM.

One follow-up for later, not blocking this PR: big-picture has its own `externalResourcesInstance` in `src/big-picture/src/pages/catalogue/use-catalogue-data.ts` with the same four fetches and no interceptor, so it still hits the error boundary when the CDN is down. Worth a sister fix at some point.

src/renderer/src/hooks/use-catalogue.ts
```diff
@@ -9,6 +11,32 @@ export const externalResourcesInstance = axios.create({
...
+        error.response == null);
+
+    if (!isNetworkOrTimeout) return Promise.reject(error);
+
+    logger.warn(
+      "[external-resources] request failed silently:",
+      error?.message ?? error
+    );
+    return Promise.resolve({
+      data: [],
```
Hachi-R: `src/renderer/src/hooks/use-catalogue.ts:L30` — P1: Returning `[]` replaces `steamUserTags` with an array; if a tag filter is already selected, `Object.keys(steamUserTags[language])` receives `undefined` and crashes the catalogue during the CDN outage this PR handles. Preserve the previous state on failure or guard the selected-tag lookup in `catalogue.tsx`.

Hachi-R (changes requested):
One blocking issue.

sotik11:
Good catch, thanks — fixed in 8e677ef.

Two of the endpoints return deep shapes that consumers read as `state[language]`:
- `/steam-user-tags` → `Record<string, Record<string, number>>`
- `/steam-genres` → `Record<string, string[]>`

A flat `[]` fallback replaced state with an array and downstream reads (`catalogue.tsx:264` `steamGenres[language].reduce(...)`, `:283` `Object.entries(steamUserTags[language])`, `:434` `Object.keys(...)`) got `undefined`. Switched to endpoint-specific fallback:

```ts
if (url.includes("/steam-genres") || url.includes("/steam-user-tags")) {
  data = {};
} else if (url.includes("/steam-developers") || url.includes("/steam-publishers")) {
  data = [];
} else {
  return Promise.reject(error);  // unknown URL — fail loudly, don't guess
}
```

The `Record<>` initial state in the reducer is `{}`, so `{}` matches consumer guards like `if (!steamGenres[language]) return {}`. And any endpoint added later that this interceptor hasn't been taught about rejects instead of being silently poisoned.

Hachi-R (approved):
LGTM

src/renderer/src/hooks/use-catalogue.ts
```diff
@@ -9,6 +11,45 @@ export const externalResourcesInstance = axios.create({
...
+      (error.code === "ERR_NETWORK" ||
+        error.code === "ECONNABORTED" ||
+        error.response == null);
+
+    if (!isNetworkOrTimeout) return Promise.reject(error);
+
+    const url = error?.config?.url ?? "";
+    let data: unknown;
+    if (url.includes("/steam-genres") || url.includes("/steam-user-tags")) {
+      data = {};
```
Hachi-R: `src/renderer/src/hooks/use-catalogue.ts:L28` — P1: Returning `{}` still replaces previously loaded `steamUserTags`; with an active tag filter, `catalogue.tsx:L434` calls `Object.keys(steamUserTags[language])` with `undefined` and crashes during the outage. Preserve existing state by handling failure at each callsite, or guard this lookup with `steamUserTags[language] ?? {}`.

Hachi-R (changes requested):
One blocker remains.

sotik11:
Good catch @Hachi-R — fixed in `d39dabaa6` and updated the branch onto latest main.

Guarded the lookup in `catalogue.tsx` exactly as suggested:

```ts
...filters.tags.map((tag) => {
  const tagsForLanguage = steamUserTags[language] ?? {};
  return {
    label: Object.keys(tagsForLanguage).find(
      (key) => tagsForLanguage[key] === tag
    ),
    ...
  };
}),
```

This was the only unguarded `steamUserTags[language]` / `steamGenres[language]` access left in the render path — the sibling lookups (`steamUserTagsFilterItems`, `steamGenresMapping`, and the `game-item` variants) already guard with `if (!...)` / `?.` / `&&`, so this now matches them. With the guard, an active tag filter during a CDN outage no longer crashes the catalogue regardless of state (cold start or re-mount).

Non-blocking follow-up idea (happy to do it here or in a separate PR, your call): the interceptor still overwrites already-loaded `steamUserTags`/`steamGenres` with `{}` on failure, so the tag dropdown empties during an outage even though we had data a moment ago. Preserving the last-known-good state on network failure (i.e. not dispatching the empty fallback for the two redux-backed fetches) would make the degradation fully graceful. It's purely cosmetic now that the crash is guarded, so I left it out to keep this PR focused.

src/renderer/src/hooks/use-catalogue.ts
```diff
@@ -9,6 +11,45 @@ export const externalResourcesInstance = axios.create({
...
 });
 
+externalResourcesInstance.interceptors.response.use(
+  (response) => response,
+  (error) => {
+    const isNetworkOrTimeout =
+      axios.isAxiosError(error) &&
+      (error.code === "ERR_NETWORK" ||
+        error.code === "ECONNABORTED" ||
+        error.response == null);
```
Moyasee: `response == null` also matches in-flight `CanceledError` and other response-less Axios failures. This converts cancellation into a synthetic `200 OK` and runs the request's `.then()` handler. Explicitly preserve cancellations, and only handle the intended network and timeout codes.

src/renderer/src/hooks/use-catalogue.ts
Moyasee: A request from an earlier catalogue mount can fail after a newer mount has loaded metadata. This fallback turns that stale failure into `{}`, and the old request then erases the newer Redux state. Catch failures at the callsites without dispatching, or cancel and generation-gate requests during effect cleanup.

src/renderer/src/hooks/use-catalogue.ts
```diff
@@ -9,6 +11,45 @@ export const externalResourcesInstance = axios.create({
...
+
+    if (!isNetworkOrTimeout) return Promise.reject(error);
```
Moyasee: Should transient 502/503/504 responses degrade too? Axios attaches `response` to those errors, so they are still rejected. The blocked-host reproduction is covered, but the broader CDN-outage case is not.

Moyasee (changes requested):
The `catalogue.tsx` guard fixes the synchronous render crash and should remain, but the interceptor does not address the user-visible failure described by the PR. Current `main` no longer renders `GlobalErrorOverlay`, React error boundaries do not catch asynchronous rejections, and `main.tsx` only logs `unhandledrejection`. Expected request failures should be handled locally without synthesizing successful responses or replacing existing metadata.

sotik11:
Thanks @Moyasee — you're right on all three, and the diagnosis matches what I'd found while chasing Hachi-R's crash report: the boundary never saw those rejections, so the interceptor was solving a problem that doesn't exist while creating a real one. Dropped it entirely in `3ab52b48e` and rewrote the PR description, since the original one described the failure inaccurately.

Failures are now handled at each callsite, with nothing synthesized and nothing dispatched on failure:

| Failure | Result |
| --- | --- |
| Cancellation (`CanceledError`) | ignored — never treated as a success |
| Network error / timeout | logged, previously loaded data kept |
| Transient `502`/`503`/`504` | logged, previously loaded data kept |
| Anything else (`404`, `401`, …) | still rejects, misconfig stays visible |

Point by point:

- **Cancellations** — `axios.isCancel` is checked first and returns early, and the loose `error.response == null` clause is gone; only `ERR_NETWORK` / `ECONNABORTED` match on the response-less path.
- **Stale requests erasing newer state** — nothing dispatches on failure, so a late failure from an earlier mount can't overwrite metadata a newer mount already loaded. I went with the callsite option rather than cancel + generation-gating to keep the change small; happy to add an `AbortController` in the effect cleanup as well if you'd prefer that.
- **Transient 5xx** — good catch, that's the actual CDN-outage shape and it wasn't covered. `502/503/504` now degrade the same way as connection failures; other statuses still reject.

The `catalogue.tsx` guard stays as you asked. I also re-checked every consumer of the four resources: `steamUserTags`/`steamGenres` are guarded at all render-path lookups, and `steamPublishers`/`steamDevelopers` are only `.map`ped, so keeping the previous value (`{}`/`[]` on a cold start) is safe everywhere. These four JSONs are fetched only through this hook.

src/renderer/src/hooks/use-catalogue.ts
```diff
@@ -9,6 +11,29 @@ export const externalResourcesInstance = axios.create({
   baseURL: import.meta.env.RENDERER_VITE_EXTERNAL_RESOURCES_URL,
 });
 
+const DEGRADABLE_STATUS_CODES = [502, 503, 504];
```
Moyasee: Can this half come out? As far as I can tell it doesn't change behaviour.

Failures already reach the log today: main.tsx registers an `unhandledrejection` handler that calls `logger.error`. And on failure the `.then` never runs either way, so nothing dispatches and previously loaded data is already kept without this. Anything outside `DEGRADABLE_STATUS_CODES` is rethrown straight back into that same handler.

So the net result is a warn line naming the resource instead of the generic error line, for 502/503/504 and network errors only, in exchange for 37 lines and a status allowlist to maintain.

Dropping it leaves a one-file crash fix, which is the part that's actually load bearing.

Moyasee (dismissed):
The interceptor findings from my earlier review are addressed: nothing synthesizes a response now, cancellation is no longer turned into a success, and no failure path dispatches, so a stale request can't overwrite newer metadata. This review supersedes that one.

The catalogue.tsx guard fixes a real crash and I want it merged. It also isn't CDN dependent, which the description undersells. steam-user-tags.json only ships four locales:

```
['en', 'es', 'pt', 'ru']
```

The app ships around thirty, and `language` is `i18n.language.split("-")[0]`, so `steamUserTags[language]` is undefined for de, fr, pl, uk and most of the rest on a completely healthy load. Tick a tag filter, switch the language to German, and the chip lookup throws during render. The boundary sits above the router in main.tsx, so the whole app drops to the fallback, not just the page:

```
Something went wrong
The app ran into an unexpected error. You can reload to keep using Hydra.
Cannot convert undefined or null to object
ORIGIN src/pages/catalogue/catalogue.tsx:364:25
TypeError: Cannot convert undefined or null to object
    at Object.keys (<anonymous>)
    at .../catalogue.tsx:364:25
    at Array.map (<anonymous>)
```

With your guard applied the same steps render the catalogue in German with no crash. Worth putting that path in the description, since a reviewer testing only with the CDN blocked can miss it.

The use-catalogue.ts half is what I'd like dropped, see the inline comment. With it out the title should describe the crash fix instead of the degrade behaviour. Once that's done this is good to merge.

sotik11:
Done. Dropped the `use-catalogue.ts` interceptor — the PR is now just the `catalogue.tsx` guard, rebased onto main.

The title/description now describe the crash instead of the degrade: `steam-user-tags.json` ships only en/es/pt/ru, so `steamUserTags[language]` is undefined for de/fr/pl/uk etc. on a healthy load — a tag filter plus a non-shipped locale throws `Object.keys(undefined)`, and the boundary above the router drops the whole app.

src/renderer/src/pages/catalogue/catalogue.tsx
```diff
@@ -422,15 +422,18 @@ export default function Catalogue() {
...
-        label: Object.keys(steamUserTags[language]).find(
-          (key) => steamUserTags[language][key] === tag
-        ),
-        filterType: t("tags"),
-        orbColor: filterCategoryColors.tags,
-        key: "tags",
-        value: tag,
-      })),
+      ...filters.tags.map((tag) => {
+        const tagsForLanguage = steamUserTags[language] ?? {};
```
Hachi-R, line 426: `src/renderer/src/pages/catalogue/catalogue.tsx:L426` — P3: Falling back to `{}` prevents the crash, but it leaves an existing active tag chip unlabeled after switching to an unsupported locale because the lookup returns `undefined` and the renderer coalesces it to `""`. Prefer `steamUserTags[language] ?? steamUserTags.en ?? {}` and use the same fallback for the tag filter list so selected tags remain identifiable.

Hachi-R (approved):
The guard fixes the reported crash. One non-blocking fallback improvement is noted inline.

PR #2463, opened 2026-07-07
fix: fall back to official Gofile when alternate CDN is rate-limiting

Moyasee:

**Fill in the PR content:**

## What

Make Gofile downloads fall back to the official Gofile downloader when the alternate CDN (`gofilecdn.eu.cc`) is rate-limiting.

## Why

We had reports of Gofile downloads being very inconsistent: they start, sit on "calculating time remaining" for a few seconds, then pause/stop on their own. Cancelling and retrying makes it worse, and results differ between users and even between attempts on the same file.

Digging through tester logs, every failing download was hitting HTTP 429 from the alternate CDN, and it never fell back to the official Gofile path. Two testers combined logged ~180 GET responses from the CDN, the large majority of them 429. One tester had 16/16 downloads routed to the CDN with 0 official fallbacks and 0 completions.

The root cause is the CDN availability check. It probed with an `OPTIONS` request, which the CDN happily answers `200` even while it rate-limits (429) the actual file `GET`. So the check always passed, the download committed to the CDN, and then died on repeated 429s inside the downloader (4 retries, then error). The official Gofile path was healthy the whole time: in the same logs, whenever we did reach it, the link resolved and the HEAD/preflight returned 200.

## What changed

`GofileApi.getAlternateCdnDownloadLinkIfAvailable` now probes the CDN with a real ranged `GET` (`Range: bytes=0-0`) instead of `OPTIONS`. If the CDN returns 429, an HTML page, or any non-OK status, we treat it as unavailable and return null, which makes `resolveGofileDownload` fall back to the official Gofile downloader. When the CDN is healthy the behavior is unchanged (it still gets used).

## Notes / follow-ups (not in this PR)

- `resolveGofileDownload` still runs twice per start (once in `validateDownloadUrl`, once in `startDownload`), and each re-probes independently, so they can occasionally pick different backends. Worth collapsing to a single resolve later.
- This fixes the common case (CDN 429 at resolve time). A CDN that passes the probe but then starts 429-ing mid-download still relies on the downloader's existing retry/error handling. Could be hardened to re-resolve to official on a mid-download 429 as a later step.


Moyasee:
Good catch. Hardened the probe to also treat text/plain and application/json 2xx bodies as non-file responses so a throttle page can't pass. Left missing content-type allowed since some working CDN/proxy responses omit it for real files.

@greptile review

Hachi-R (approved):
LGTM

PR #2464, opened 2026-07-08
fix: guard non-array API responses to prevent renderer crashes

Moyasee:

**Fill in the PR content:**

Users were hitting the error screen with `featuredGameDetails.map is not a function` on the home hero and `catalogue[currentCatalogueCategory].map is not a function` on the catalogue.

The origin API always returns arrays on success and rejects on errors, so those paths are safe. The bad value comes from the edge: the CDN can answer with an HTML body and a 200 status (challenge/WAF page), which resolves instead of rejecting. That non-array body then reaches components that call `.map` on it and the render throws, and since the error boundary is app-wide the whole app drops to the error screen.

This adds an `ensureArray` helper that returns `[]` and logs a warning when the value is not an array, then uses it at every place we map an API list: featured hero, home catalogue, game repacks, profile badges, notification badges and game collections. The UI now renders an empty state instead of crashing, and the warning keeps the underlying edge issue visible in the logs.

This is a client-side guard, not a cure for the CDN behavior itself.

### DIFF
--- a/src/renderer/src/components/hero/hero.tsx
+++ b/src/renderer/src/components/hero/hero.tsx
@@ -3,6 +3,7 @@ import { useEffect, useState } from "react";
 import type { TrendingGame } from "@types";
 import { useTranslation } from "react-i18next";
 import Skeleton from "react-loading-skeleton";
+import { ensureArray } from "@renderer/helpers";
 import "./hero.scss";
 
 export function Hero() {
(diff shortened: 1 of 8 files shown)


Moyasee:
Good catch on the log bloat. Now truncating strings to 200 chars and only stringifying non-strings, so a big WAF HTML body won't get dumped to the log file anymore.

@greptile review

Hachi-R (approved):
LGMT

Moyasee:
Both good points, addressed:

1. Wrapped the preview build in try/catch so the log path can't throw (circular refs / BigInt). Falls back to a type tag. In practice these values come in as JSON over IPC so they can't be circular, and the big WAF payload is a string that hits the slice branch, not stringify, but agreed the guard shouldn't be the thing that throws.
2. Swapped the /features guard in use-feature.ts to ensureArray for consistency + the warning. Left the line 17 one as is since that's a localStorage parse, not an API response, so the "expected array from source" message wouldn't fit there.

@greptile review

PR #2466, opened 2026-07-09
fix: guard against undefined game title in library sort

Moyasee:
## What this fixes

Sorting the library crashes the whole app when a game has an undefined `title`. `sortedLibrary` calls `title.localeCompare(...)`, and `getLibrary` returns `title: gameAssets?.title || game.title`, which is `undefined` when a game has neither a title nor cached shop assets (for example a game whose asset lookup failed, or a synced entry that came through without a title). `undefined.localeCompare` throws inside `Array.sort`, and the root error boundary turns that into the full "Algo deu errado" screen.

In production this surfaces as `Cannot read properties of undefined (reading 'localeCompare')` coming from the library title comparator.

## The change

Default the sort key to an empty string in both the ascending and descending title branches, so a game with a missing title sorts to the start instead of crashing:

```ts
(a.title ?? "").localeCompare(b.title ?? "", undefined, { sensitivity: "base" });
```

No change in behavior for games that already have a title.

### DIFF
--- a/src/renderer/src/pages/library/library.tsx
+++ b/src/renderer/src/pages/library/library.tsx
@@ -566,7 +566,7 @@ export default function Library() {
 
     const queryLower = deferredSearchQuery.toLowerCase();
     return filtered.filter((game) => {
-      const titleLower = game.title.toLowerCase();
+      const titleLower = (game.title ?? "").toLowerCase();
       let queryIndex = 0;
 
       for (
--- a/src/renderer/src/pages/library/library.tsx
+++ b/src/renderer/src/pages/library/library.tsx
@@ -517,7 +517,7 @@ export default function Library() {
         }
 
         case "title_desc": {
-          return b.title.localeCompare(a.title, undefined, {
+          return (b.title ?? "").localeCompare(a.title ?? "", undefined, {
             sensitivity: "base",
           });
         }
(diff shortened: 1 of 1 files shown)


src/renderer/src/pages/library/library.tsx
```diff
@@ -527,7 +527,7 @@ export default function Library() {
           break;
       }
 
-      return a.title.localeCompare(b.title, undefined, {
+      return (a.title ?? "").localeCompare(b.title ?? "", undefined, {
```
chubbygrannychaser, line 530: This protects the sort fallback, but `filteredLibrary` below still does `game.title.toLowerCase()` for non-empty search queries. If the same library entry has an undefined title, opening Library with a saved/typed search will still throw `Cannot read properties of undefined (reading 'toLowerCase')`. Please normalize the title once (for example `const title = game.title ?? ""`) and reuse that for the search matcher too.

chubbygrannychaser (changes requested):
Thanks for the fix. Sorting is now guarded, but the same malformed library entry still crashes as soon as the user has a non-empty library search query, so this needs one more guard before it fully fixes undefined titles.

src/renderer/src/pages/library/library.tsx
Moyasee: Good catch, thanks. Guarded the search matcher too in 103ad7f1 with `(game.title ?? "").toLowerCase()`. That was the only other place a method gets called on title, so undefined titles are covered on both the sort and search paths now.

chubbygrannychaser (approved):
LGTM — previous undefined-title search blocker is fixed, and changed-file ESLint, typecheck:web, and npm test pass locally.

PR #2471, opened 2026-07-10
Feature/lbx 868 Local parser for the Ludusavi manifest

Hachi-R:


Moyasee (changes requested):
Nice work, the structure here is really clean. Atomic temp-file + rename, the per-path mutex to collapse concurrent fetches, and the stale-index fallback on download failure are all the right calls. Both Windows and Linux builds are green so the new deps compile everywhere.

One thing I'd like fixed before merge, and then a handful of optional notes.

### Must fix: bound the manifest download

`download_manifest` uses `reqwest::get` with no timeout, and `get_manifest_index` holds the per-path `Mutex` for the entire `load_index` call, which includes that download:

```rust
let mut cached = cache.lock().await;
...
let index = load_index(user_data_path, source_url).await // <- download happens under the lock
```

Serializing the fetch is intentional and good, but with no timeout a half-open or stalled connection means the future never resolves, so the guard is held indefinitely and every later `get_manifest_index` for that path blocks behind it with no recovery short of a restart. A request timeout closes that hole. Easiest fix also addresses the "new client per call" nit below by reusing one client:

```rust
static HTTP_CLIENT: OnceLock<reqwest::Client> = OnceLock::new();

fn http_client() -> &'static reqwest::Client {
    HTTP_CLIENT.get_or_init(|| {
        reqwest::Client::builder()
            .timeout(Duration::from_secs(30))
            .build()
            .expect("failed to build manifest http client")
    })
}
```

### Optional

- **reqwest default features are heavy for a single GET.** The default set is pulling in `aws-lc-sys` (C/cmake build) and `quinn`/HTTP3. For fetching one YAML over HTTPS you can slim it down, e.g. `reqwest = { version = "0.13", default-features = false, features = ["rustls-tls"] }`. Builds pass today so this is purely binary size / build cost.
- **Integration tests hit the live CDN and assert on Balatro specifics** (`manifest_key == "2379780"`, presence of `save` tags, both `file` and `dir` kinds). Nothing runs `cargo test` in CI, so these only bite locally, but they'll fail offline or whenever the manifest content shifts. Worth gating behind `#[ignore]` or a small checked-in fixture so the unit-level logic is testable without the network.
- **Normalized lookup drops candidate priority.** In `find_manifest_entry`, the exact-match loop respects the object_id > remote_id > title order, but the normalized fallback just returns the first game in manifest order whose normalized key matches any candidate. Low real-world risk since a numeric id and a title rarely normalize to the same thing, but it can technically prefer a lower-priority match.
- **`infer_rule_kind` on dotdirs.** A path whose base name is something like `.config` gets classified as `file` because it contains a `.`. Probably fine for the manifest in practice, just flagging the heuristic edge.
- Agree with Greptile on the inline `1` version literal becoming a named const.

Also noting there's no JS consumer of `get_save_rules_for_game` yet, so right now only the Rust tests exercise this. Assuming that's intentional and the wiring lands in a follow-up.

PR #2472, opened 2026-07-10
feat: add SteamGridDB custom asset support

Moyasee:

**Fill in the PR content:**

Adds SteamGridDB support for custom game assets, one of the most requested features (LBX-147).

## What it does

A new SteamGridDB section under Settings > Integrations lets users bring in community artwork for their library games. It is opt-in and fully local, no account data leaves the machine.

- User provides their own free SteamGridDB API key.
- Global toggles per shop (Steam / Hydra Classics) and per asset type (covers, hero banners, logos, icons). When enabled, Hydra auto-matches the highest-scoring asset for each game.
- Per-game override in the game's Assets tab: browse every available variant and pick a specific one, or force the feature on/off for that single game regardless of the global setting.
- Falls back to SteamGridDB art when a game has no native hero or cover.

## How it works

- SteamGridDB matches are resolved once per game (on library sync, on add, or on manual refresh) and cached in leveldb, so there is no per-request or per-render fetching. Steam games match exactly by app id; Classics/LaunchBox games match by title search.
- The chosen assets are composed into the returned shop assets at the read boundaries (`get-library`, `get-game-shop-details`, `get-game-assets`), so every surface that already reads those fields picks them up, including big picture. Manually uploaded custom assets always take precedence over SteamGridDB ones.
- Assets are hotlinked by default. A "Cache images on device" option downloads them locally for offline use.
- Everything (global settings, per-game choices, cached art) is stored locally.

## Notes

- New translation keys added for en, pt-BR, pt-PT, ru and es.
- Adds a manual cover/grid upload slot to the per-game Assets tab, which did not exist before.
- Steam always synthesizes a cover URL, so the "no native cover" fallback effectively applies to Classics and to hero banners; this is called out in the code.


Moyasee:
@greptile review

Moyasee:
Fixed all three: term-based search no longer writes to the variant cache (was poisoning automatch), and the retry backoff delays are named constants now.

@greptile review

Moyasee:
Fixed the last nit too: Refresh now keeps the active search term instead of resetting to the default variants.

@greptile review

Moyasee:
Good catch. Now a failed key change validates against a throwaway client and only swaps the live one on success, so an already-working key survives a bad save.

@greptile review

Moyasee:
Pushed the SteamGridDB rework (local integration replaced by the /artwork/:kind endpoint, no more API key) plus a few custom-cover fixes: chosen covers skip the Classics blur, the cover preview no longer stretches across the modal, and the Assets tab fetches composed assets on open so the preview isn't blank until reopened.

@greptile review

Moyasee:
Addressed the nits: pick/clear handlers now catch failures and toast, the sentinel root margin is a named constant, and the library card's cover media is extracted out of the nested ternary.

@greptile review

Moyasee:
Pushed skeleton loading + hover for the artwork picker, plus the cover preview flicker fix and contained (not cropped) library cover.

@greptile review

Moyasee:
Added the SteamGridDB artwork picker + cover slot to Big Picture (controller-navigable focus grid, infinite scroll, skeletons, apply spinner). Also fixed the Big Picture ERR_BLOCKED_BY_ORB (missing browser UA / CORS on that window's requests) and stopped a custom icon from bleeding into the hero/cover fallbacks.

@greptile review

Moyasee:
Deduped the picker logic: the shared paging/selection/pick/clear code now lives in a single useGameArtworkGrid hook used by both the renderer and Big Picture pickers, so each component only keeps its own layout and focus wiring.

@greptile review

Moyasee:
@greptile review

Hachi-R:
## Findings

- `src/main/events/catalogue/get-cover-poster.ts:L71` — P1: `axios` follows redirects after validating only the initial URL, allowing SSRF through a redirect from an authorized host. Require `https:`, set `maxRedirects: 0`, and validate every `Location`.

- `src/main/services/window-manager.ts:L220` — P1: Removing the `webContentsId` filter applies the random `User-Agent` and CORS `*` to auth, friends, and editor windows too. Restrict this to Hydra renderers and SGDB URLs.

- `src/main/events/library/set-game-artwork-selection.ts:L86` — P2: The manual asset is deleted before the new selection is persisted. If the LevelDB write fails, both assets are lost. Persist atomically and delete the file afterward.

- `src/main/services/library-sync/merge-with-remote-games.ts:L129` — P2: Cloud merge updates the URLs but does not clear `gamesArtworkSelectionSublevel`. Stale selections continue overriding resets made on another device.

- `src/main/events/library/create-steam-shortcut.ts:L65` — P2: Custom WebP/GIF files are copied into `.ico`, `.jpg`, and `.png` paths without conversion. Steam may not render these assets. Convert them or preserve the real extension.

- `src/main/events/catalogue/get-cover-poster.ts:L103` — P2: Local assets ignore `MAX_SOURCE_BYTES` and are loaded entirely into memory. Validate `stats.size` before `readFile`.

src/renderer/src/pages/settings/emulation/roms-detected-section.tsx
```diff
@@ -91,7 +91,12 @@ export function RomsDetectedSection({
             <>
               <div className="emulator-detail__roms">
                 {pageSlice.map((rom) => {
-                  const cover = rom.libraryImageUrl ?? rom.iconUrl;
+                  const cover =
+                    rom.customCoverImageUrl ??
+                    rom.customIconUrl ??
```
Hachi-R, line 96: P2: `customIconUrl` is checked before `coverImageUrl`, which already contains the composed SteamGridDB selection. With a custom icon and a selected cover, the ROM list shows the icon and ignores the cover. Prioritize `coverImageUrl` before the icon fallback.

src/big-picture/src/components/pages/game/game-settings-modal/artwork-picker.tsx
```diff
@@ -0,0 +1,217 @@
...
+            asChild
+          >
+            <button
+              type="button"
+              className={`game-artwork-picker__item game-artwork-picker__item--${assetType} ${
+                currentArtworkId === item.id
+                  ? "game-artwork-picker__item--active"
+                  : ""
+              }`}
+              onClick={() => handlePickItem(item)}
```
Hachi-R, line 156: P2: While `pendingId` is set, the buttons and the `FocusItem` action remain active. Two quick clicks or controller presses run `pick` concurrently, so the final asset depends on LevelDB and cloud completion order. Guard `handlePickItem` and disable all tiles while a selection is pending.

src/big-picture/src/helpers/preferred-assets.ts
```diff
@@ -114,8 +118,8 @@ export function resolvePreferredGameAssets(
     game?.iconUrl
   );
   const coverSrc = getFirstResolvedSource(
+    game?.customCoverImageUrl,
     game?.customHeroImageUrl,
```
Hachi-R, line 122: P2: `customHeroImageUrl` is checked before `assets.coverImageUrl`, so a game with a custom hero ignores a SteamGridDB cover in every `coverSrc` consumer. Move the hero fallback after the cover and library sources.

Hachi-R (approved):
Approved. Left three non-blocking P2 comments about picker concurrency and cover source priority.

PR #2473, opened 2026-07-10
feature/lbx-869 Path Resolution

Hachi-R:


Moyasee (changes requested):
The token-expansion and Wine-prefix design reads well. Building the token map once per call, expanding via Cartesian product, dedup with the `seen` set, and keying the Wine-prefix decision off the raw path (before the token text is gone) are all the right moves.

One thing I'd want addressed before this merges, plus a few smaller notes.

### Should fix: unhandled tokens are emitted as valid resolved paths

`tokens_used_by_path` only looks at `PATH_RESOLUTION_TOKENS`, so a token that isn't in that list is never counted, never added to `unresolved_tokens`, and survives `expand_tokens` untouched. The result is that the literal token string is returned inside `resolved_paths` as if it were a real path. The `// TODO: Unknown manifest tokens are not detected` comment frames this as a detection gap, but the consequence is worse than not detecting it, we hand back a wrong path that looks resolved.

This isn't hypothetical. I pulled the live manifest and grepped the tokens actually in use:

```
194236 <winPublic>
148889 <winAppData>
 48647 <winProgramData>
 12774 <home>
  9130 <base>
  6269 <winLocalAppData>
  4006 <winDocuments>
  3730 <storeUserId>
  1525 <root>          <- not handled
   822 <xdgConfig>     <- not handled
   677 <xdgData>       <- not handled
    10 <osUserName>    <- not handled
     4 <winDir>        <- not handled
```

`<xdgConfig>` and `<xdgData>` are the primary tokens for native Linux game saves, which is exactly the platform this resolver targets, so those ~1500 entries will silently produce paths like `<xdgData>/balatro`. I'd suggest two parts:

1. Handle the missing tokens. At minimum `<xdgData>` (`~/.local/share`) and `<xdgConfig>` (`~/.config`), plus `<root>`, `<osUserName>`, and `<winDir>`.
2. Add a catch-all guard: after expansion, if any `<...>` or `%...%` fragment remains, push it to `unresolved_tokens` and drop the path, so a future unknown token fails closed instead of syncing garbage.

### Minor

- **Hardcoded `C:/` for `winPublic` / `winProgramData`** (`context.rs`). These break if Windows lives on another drive. Deriving the drive letter from `app_data_dir` instead of hardcoding would cover that. Real-world hit rate is low, but it's cheap to make correct.
- **`local_app_data_dir` derivation assumes `app_data_dir` is the Roaming folder** (parent + `Local`). That holds for Electron's `appData` path, but it's an undocumented assumption that silently yields a wrong path if the input ever changes. A short comment noting the expected input would help the next reader.
- **`proton_path` and `steam_path` are in `ResolveSaveRulesInput` but unused.** If they're reserved for upcoming Proton prefix handling that's fine, worth a note; otherwise drop them so the NAPI surface doesn't imply support that isn't there.

Also flagging that there's no JS consumer of `resolve_save_rules` yet, so only the Rust integration test exercises this, and that test reaches the live manifest over the network (same pattern as the manifest PR). Assuming the wiring lands in a follow-up.

PR #2477, opened 2026-07-10
fix(i18n): add Japanese language support

Ayufa:
Adds Japanese language support by registering the existing `ja` locale in the i18n resources and adding the `ja: "JP"` flag mapping.

This enables Japanese translations and displays the Japanese flag in the language selector.


Moyasee (approved):
LGTM

Moyasee (approved):
Looks good. The `ja` locale file was already sitting in the repo and just was not registered, so this is the minimal correct fix: add it to the resource map and give it the right flag code.

Both language pickers build their lists from `Object.entries(languageResources)` and sort by `language_name`, which `ja/translation.json` defines, so nothing else needs touching for it to show up and be selectable. Partial key coverage is fine here since `fallbackLng` is `en` and most of the locales we already ship are less complete than this one.

CI is green.

Hachi-R (changes requested):
`src/locales/index.ts:L44` — P2: Registering `ja` exposes Japanese in both language pickers, but `src/shared/index.ts:getDateLocale` and `src/big-picture/src/hooks/use-date.hook.ts:getDateLocale` still fall through to `enUS`. Relative dates, playtime, reviews, and notifications therefore render English text such as `2 days` inside the Japanese UI, while fixed dates use `dd/MM/yyyy`. Map `ja` to the date-fns Japanese locale in both renderers and use a Japanese date format such as `yyyy/MM/dd`.

Hachi-R:
Closing this because the branch history is no longer compatible with the current `main`. GitHub reports this PR as `DIRTY`/`CONFLICTING` and shows 5,363 commits across 1,402 files, so the intended change can no longer be reviewed safely. Feel free to reopen this PR after rebuilding the branch or cherry-picking the intended commits onto the current `main`.

PR #2484, opened 2026-07-11
fix(big-picture/home-page): Change header-back-button logic on the Big Picture home page

therxdev:

- [X] I have read the [Hydra documentation](https://docs.hydralauncher.gg/getting-started.html).
- [X] I have checked that there are no duplicate pull requests related to this request.

**Fill in the PR content:**

A small visual tweak: I noticed that the header on the launcher's home page is different between the regular mode and Big Picture mode. In the regular version, there is no "Go Back" action, while in Big Picture mode it is present. That was my decision :)

Home page header in the regular mode:
image

Big Picture mode (before my changes):
image

Big Picture mode (after my changes):
image


src/big-picture/src/layout/header/index.tsx
```diff
@@ -45,8 +45,12 @@ function Header() {
   const pageTitle = useCurrentPageTitle();
   const isOnCataloguePage =
     normalizeBigPicturePathname(pathname) === "/catalogue";
+  const isOnHomePage = normalizeBigPicturePathname(pathname) === "/";
   const catalogueSearchValue = searchParams.get("title") ?? "";
   const [isSearchOpen, setIsSearchOpen] = useState(false);
+  const [isHovered, setIsHovered] = useState(false);
+  const shouldShowHomeTitle = isOnHomePage && !isSearchOpen;
+  const shouldShowBackButton = !isOnHomePage;
```
Moyasee, line 53: This keys the back button off the route, but what actually decides whether back is meaningful is the history stack, and we already track it. You can land on home with a non-empty stack: the sidebar has a Home route, so home, then a game page, then Home from the sidebar leaves a stack of 3. On that screen the back button currently works and takes you to the game page, and after this change it is gone. Note B still goes back there, because `triggerBackAction` in `navigation-input.provider.tsx` gates on `historyStack.length > 1` rather than the pathname, so the button and the B press would now disagree.

Could you gate on the same thing B does?

```js
const historyStack = useNavigationHistoryStore((state) => state.stack);
const shouldShowBackButton = historyStack.length > 1;
```

That still fixes the real bug (fresh launch on home, nothing behind it, no button) but keeps the button when it does something. `useNavigationHistoryStore` comes from `../stores` and `use-header-title.hook.ts` already pulls it in, so it is right there.

Moyasee, line 53: Related, and it applies however you gate this: whenever the back button is not rendered, the `left` override in `searchNavigationOverrides` (line 75) still points at `HEADER_BACK_BUTTON_ID`. `resolveItemOverrideTarget` returns null for a node that is not registered, so `moveFocus` hits `warnInvalidOverride` and then falls back to tree navigation, which finds nothing to the left inside the header. The user-visible version of that is pressing left from search and having focus just sit there.

Worth gating the override on `shouldShowBackButton` and using `{ type: "block" }` when the button is absent, so it is not resolving against a node that is not mounted.

src/big-picture/src/layout/header/index.tsx
```diff
@@ -284,14 +287,24 @@ function Header() {
     <div className="header">
       <HorizontalFocusGroup regionId={BIG_PICTURE_HEADER_REGION_ID} asChild>
         <header className="header__container">
-          <FocusItem id={HEADER_BACK_BUTTON_ID} asChild>
-            <button className="header__action" onClick={() => navigate(-1)}>
-              <ArrowLeftIcon size={24} weight="bold" />
-              <Typography variant="label" className="header__title">
+          {shouldShowHomeTitle && (
+            <div className="header__home-title">
+              <Typography variant="h4" className="header__title">
```
Moyasee, line 292: `header__title` only has styles when it is nested under `.header__action` (styles.scss line 76), so on this standalone title the class does nothing. The result is that the home title inherits `.typography`'s `color: var(--text)` instead of the `var(--primary)` white the header title uses on every other page, so it renders noticeably dimmer.

If the grey is not deliberate, either lift the color rule out of `.header__action` or set it on `.header__home-title`. If it is deliberate, the class can just be dropped since it is inert here.

Moyasee (changes requested):
The intent is right, home should not offer a dead back action, and the hover flicker from the first commit is already fixed in the second. But I think the condition is keyed off the wrong thing, and as written this removes a back button that is sometimes genuinely useful.

You can reach home with history behind you: the sidebar has a Home route, so home to a game page and back to home via the sidebar leaves a real history stack. Today the header back button on that screen works and takes you to the game. After this PR it is gone, even though B still goes back, because `triggerBackAction` keys off the history stack rather than the route. So controller users keep the behaviour, mouse users lose the only on-screen affordance, and the two disagree.

Gating on the history stack instead of the pathname fixes the actual bug (fresh launch on home, nothing behind it) without taking the button away when it still does something. Details inline, plus a small styling thing on the new title.

PR #2485, opened 2026-07-12
feat: add achievement notification theme customizer

NickCh11:
## Summary

- Integrate the achievement notification customizer into Theme Editor.
- Generate and update only a managed CSS block, preserving user-authored CSS.
- Support Default, Rare, Hidden, and Platinum variations with isolated previews.
- Add per-variation achievement sounds with inheritance, mute, volume, and custom files.
- Add a static preview, current-variation test notification, and resizable CSS editor.

## Testing

- `npm test`
- `npm run typecheck`
- `npm run format-check`
- Production build and development runtime smoke checks.


NickCh11:
@greptile

NickCh11:
@greptile

NickCh11:
@Moyasee made the changes you requested in this new PR

src/locales/en/translation.json
```diff
@@ -1633,5 +1633,42 @@
     "copied": "Copied!",
     "minimize": "Minimize",
     "close": "Close"
+  },
+  "achievement_notification_customizer": {
```
Moyasee, line 1637: This adds a whole new `achievement_notification_customizer` namespace, but only to `en`. New keys need to land in `pt-BR`, `pt-PT`, `ru` and `es` as well, otherwise those users fall back to English for the entire customizer panel. Could you add the same block to those four?

src/renderer/src/pages/settings/settings-context-notifications.tsx
```diff
@@ -190,31 +160,19 @@ export function SettingsContextNotifications() {
...
-                className="settings-general__test-achievement-notification-button"
-                onClick={() =>
-                  window.electron.showAchievementTestNotification()
-                }
-              >
-                {t("test_notification")}
-              </Button>
-            </>
-          )}
+        <Button
```
Moyasee, line 196: This drops the `achievement_custom_notification_position` select, and `SettingsContextNotifications` is the notifications UI that `settings.tsx` actually renders (line 109). The replacement control lives in the Theme Editor and writes `--achievement-notification-position` into a theme's CSS, so anyone who has not created and activated a custom theme now has no way to move the notification at all, and `userPreferences.achievementCustomNotificationPosition` is frozen at whatever it happened to be.

Can we keep the select here as the base setting and let the theme CSS override it, rather than removing it?

src/renderer/src/pages/theme-editor/achievement-notification-customizer.tsx
```diff
@@ -0,0 +1,520 @@
...
+    const customSound = await globalThis.electron.getThemeSoundDataUrl(
+      theme.id,
+      variation
+    );
+    if (customSound === "") return;
+    const soundUrl =
+      customSound ??
+      (await import("@renderer/assets/audio/achievement.wav")).default;
+    const audio = new Audio(soundUrl);
+    audio.volume = sound?.volume ?? soundVolume / 100;
```
Moyasee: `audio.volume = sound?.volume ?? soundVolume / 100` ignores the master `achievementSoundVolume` preference, which defaults to `0.15`. Real playback goes through `getAchievementSoundVolume()` and lands at `min(master, variation)`, so this preview plays at up to 100% while the actual notification plays at 15%. That makes the preview pretty misleading.

`getAchievementSoundVolume(achievement)` in `@renderer/helpers` already resolves this correctly, including `inherit`, so reusing it here would keep the preview honest.

src/shared/achievement-notification-theme.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  const vertical = position?.startsWith("bottom")
+    ? display.y + display.height - size.height
+    : display.y;
+
+  return { x: horizontal, y: vertical };
+};
+
+export const getEffectiveAchievementNotificationSoundVolume = (
```
Moyasee, line 249: Since this resolves to `min(master, variation)` and the master volume defaults to `0.15`, the new per-variation slider (0 to 100, defaulting to 100) does nothing across the top 85% of its range out of the box. Dragging "rare" from 100% down to 20% produces no audible change at all.

Was a multiplier (`master * variation`) the intent? As a hard ceiling the control reads as broken for anyone on the default master volume.

src/renderer/src/helpers.ts
```diff
@@ -183,20 +243,76 @@ export const getAchievementSoundUrl = async (): Promise<string> => {
...
+  const scale = Number.parseFloat(
+    style.getPropertyValue("--achievement-notification-scale") || "1"
+  );
+  const normalizedScale = Number.isFinite(scale)
+    ? Math.min(Math.max(scale, 0.6), 2)
+    : 1;
+
+  return {
+    position,
+    width: Math.ceil(360 * normalizedScale),
```
Moyasee: The base 360x140 is hardcoded here, but the SCSS you added reads `var(--achievement-notification-width, 360px)` and `var(--achievement-notification-height, 140px)`. If someone sets those in hand-written CSS the notification element grows past the window bounds and gets clipped, since the window is still sized from these numbers.

Either read the two vars here alongside `--achievement-notification-scale`, or drop them from the SCSS so they are not an unusable knob.

Moyasee (changes requested):
Nice work, this is a lot of thoughtful plumbing. The managed-CSS-block approach is a clean way to let the customizer and hand-written CSS coexist, and I appreciate the shared module plus the test coverage.

Two things I want sorted before merge. The new `achievement_notification_customizer` keys only went into `en`, so they need to land in `pt-BR`, `pt-PT`, `ru` and `es` too. And removing the position select from the notifications settings panel is a regression: `SettingsContextNotifications` is the notifications UI that `settings.tsx` actually renders, so anyone who has not created and activated a custom theme now has no way to move the notification at all, since the replacement only writes into a theme's CSS.

The rest is minor and inline: the sound preview ignores the master volume so it plays much louder than the real notification, the per-variation volume slider is inert across most of its range at the default master volume, and the width and height CSS vars the SCSS now reads are not respected when the window is sized.

One more thing that is not line-anchorable: `docs/assets/achievement-theme-customizer-preview.png` is 111 KB committed purely to host the screenshot in the PR description. Once this merges it is dead weight in the repo. Please drop it from the branch and attach the image to the PR body directly instead.

NickCh11:
@greptile

NickCh11:
- I fixed the requested issues as follows:
- Added the new customizer labels to pt-BR, pt-PT, ru, and es.
- Restored the existing achievement-notification position selector in Settings. Theme CSS overrides it only when the theme explicitly defines a position.
- Updated Preview sound to respect the global achievement volume, matching real notifications.
- Changed per-variation volume into a multiplier of the global volume. For example, 50% global × 40% Rare = 20% effective volume.
- Made the renderer read --achievement-notification-width and --achievement-notification-height from the active CSS, preventing custom notification sizes from being clipped on Windows.
- Removed the preview screenshot asset from the repository.

Moyasee (approved):
Went through the new commits and everything I raised is genuinely addressed, verified against the files rather than the commit messages:

- All 35 customizer keys are in `pt-BR`, `pt-PT`, `ru` and `es`, with zero missing and zero left as English, so they are real translations.
- The position select is back in the notifications settings, still hydrated from `userPreferences` and still persisting through `handleChange`. The theme CSS only overrides it when the var is actually defined, since `renderManagedBlock` emits the declaration only for a set value and `helpers.ts` falls back to the preference otherwise.
- The sound preview now goes through `getAchievementSoundVolume`, so it respects the master volume instead of playing at up to 100%.
- The per-variation volume is a real multiplier now (`master * variation`, defaulting the variation to 1), so the slider does something across its whole range.
- The notification window reads `--achievement-notification-width` and `--achievement-notification-height` instead of the hardcoded 360x140, and the measured size actually reaches `setBounds`.
- The 111 KB preview png is out of the branch.

The sound replace path also copies to a temp file before removing the old assets and renames into place, so a failed copy no longer leaves a theme with no sound. Good fix.

One thing I noticed while reading, not blocking and fine as a follow-up: the test notification button is no longer inside the custom-notifications block, so it can be clicked with custom notifications off. `showAchievementTestNotification` forces the custom window in that case, so the test would show something the user never actually gets on a real unlock, since a real one takes the OS-notification path.

Nice work, this turned out well.

PR #2513, opened 2026-07-13
fix(big-picture): log catalogue metadata requests that fail

sotik11:
## Summary

> **Scope change:** this PR originally added a response interceptor mirroring #2452. While re-checking it against the review feedback there, I found the interceptor was not just unnecessary here but actively harmful — details below. It has been dropped, and what remains is a logging-only change.

`useCatalogueData` settles six metadata requests with `Promise.allSettled` and reads only the fulfilled ones:

```ts
] = await Promise.allSettled([...]);

if (cancelled) return;

if (genresResponse.status === "fulfilled") {
  setSteamGenres(genresResponse.value.data.en);
}
```

A rejected request is therefore skipped silently — no `setState`, and no log line either. When the CDN behind `RENDERER_VITE_EXTERNAL_RESOURCES_URL` is unreachable, the filter lists just come up empty and nothing in the renderer log explains why.

## Fix

Log the rejection reason for each request that fails:

```ts
const logRejectedMetadataRequest = (
  resource: string,
  result: PromiseSettledResult<unknown>
) => {
  if (result.status === "rejected") {
    logger.warn(`[catalogue] failed to load ${resource}:`, result.reason);
  }
};
```

Behaviour is otherwise unchanged: a failed request still skips its `setState` and leaves the previously loaded data untouched.

## Why the interceptor was dropped

The existing code here is already correct for the failure case, and better than what an interceptor can offer:

- `Promise.allSettled` never rejects, so there is no unhandled rejection to begin with.
- The `status === "fulfilled"` guards mean a failed request never replaces already-loaded metadata.
- The `cancelled` flag with effect cleanup already generation-gates the results.

Synthesizing a `200 OK` response, as the interceptor did, converts a rejection into a *fulfilled* result — so `setSteamGenres(...)` would run after all and overwrite state with an empty payload. It also turned `CanceledError` into a success and left transient `502/503/504` uncovered. In other words it would have broken correct code, which is exactly the class of problem @Moyasee flagged on #2452.

## Test plan

- [x] With the CDN blocked, big picture catalogue loads, filter lists are empty, and the renderer log now names each resource that failed.
- [x] With the CDN reachable, all filter lists populate normally and nothing is logged.
- [x] Previously loaded metadata is untouched by a later failing request.


Moyasee (approved):
lgtm

PR #2514, opened 2026-07-13
fix(catalogue): guard the remaining repack.uris call sites

sotik11:
## Summary

Follow-up to #2451 covering the remaining unguarded `repack.uris` reads I flagged in [that comment](https://github.com/hydralauncher/hydra/pull/2451#issuecomment-4912979028).

`RepacksModal`'s availability guard renders a broken row as "offline", but the row is still clickable — clicking it opens `DownloadSettingsModal`, which iterates `repack.uris` and crashes the renderer the same way #2451 was preventing:

- `game-details.tsx` — `selectRepackUri`: `repack.uris.find(...)!`
- `download-settings-modal.tsx` — `selectedUri` memo and `downloadOptions` memo: `repack.uris.find(...)` / `for (const uri of repack.uris)`
- `big-picture/src/helpers/downloaders.ts` — `getDownloaderAvailabilityOptions`: `for (const uri of repack.uris)` (big picture has no error boundary, so a malformed repack takes the whole tree down)

## Fix

Same defensive shape as #2451 — `Array.isArray(repack.uris)` before every touch, including the big picture call site.

In `selectRepackUri` the **non-null assertion is dropped** and the undefined case is handled explicitly: `handleStartDownload` returns `{ ok: false, error: "invalid_repack" }`, which routes to the toast instead of sending an `undefined` uri into the download path. The `invalid_repack` string is added to every locale that already ships the sibling `downloader` key (en/ru/es/pt-BR/uk/pl/de/fr), so non-English users don't get an English toast.

## Test plan

- [x] Local build (`yarn build:win`) — clicking a repack row whose `uris` is undefined no longer trips the error boundary through `DownloadSettingsModal`, nor in big picture.
- [x] Normal repacks (with populated `uris`) still open the modal, populate downloader options, and start a download correctly.


Moyasee:
Hey! Can i see some screenshots of the game that doesn't have any url? Both before selecting download option and after that

sotik11:
Hey! Honest answer up front: I can't show you a real one, and I'd rather say that than stage something and present it as a live capture.

### Why there's no live repro

This is a defensive guard, in the same class as the two you already shipped yourself — `unavailableUris ?? []` in `getRepackAvailabilityStatus` (`9ca6a114b`) and `ensureArray` (`1104959d9`, *"guard non-array API responses to prevent renderer crashes"*). `ensureArray` covers the top level (is the response an array?); this one sits a step deeper: the response *is* a valid array, but an individual entry has no `uris`. Same failure mode as #2451, one level in.

I don't have a live game that reproduces it. The only way I can produce it is by forcing the shape locally.

### What the UI actually does today

I traced the full path, and the `invalid_repack` branch turns out to be **unreachable through the UI**:

- `downloadOptions` builds its downloader map from `repack.uris`, so with a malformed entry the map is empty and every option ends up with `canHandle: false`.
- Each option button is then disabled — `!option.canHandle || (!option.isAvailable && !option.isAvailableButNotConfigured)`.
- With nothing selectable, `selectedDownloader` stays `null`, which disables the primary Download button (it also rejects the selected option when `!option.canHandle`).
- `DownloadSettingsModal` is the only caller of `startDownload`, so `handleStartDownload` never runs.

So what a user actually sees for such an entry is the repack listed with an offline availability indicator and every downloader greyed out — no crash, no silent failure.

### Why keep the guard then

`selectRepackUri` previously ended in a non-null assertion:

```ts
repack.uris.find((uri) => getDownloadersForUri(uri).includes(downloader))!
```

Two problems with that: it throws outright if `uris` isn't an array, and when it simply finds no match it returns `undefined` while the type claims otherwise — so `uri: undefined` would be handed to `addGameToQueue` / `startDownload`. The guard turns both into an explicit, logged, typed failure instead of a malformed download request. It's a last line of defence behind a UI that already blocks the path.

### One thing your question did catch

`{ ok: false, error: "invalid_repack" }` is fed straight into `t(response.error)`, and `invalid_repack` didn't exist in any locale — so if that branch ever *did* fire, the toast body would render the literal string `invalid_repack` instead of a message. Added the key in `062db4d38`. Thanks for pushing on this, that one was a genuine miss on my side.

### Screenshots

If you'd still like to see it, I can force a malformed entry locally and capture both states — the options list and the settings modal. I'd label it clearly as **simulated**, since it isn't something I can trigger with real data. Just say the word and I'll post them.

src/locales/en/translation.json
```diff
@@ -436,6 +436,7 @@
     "region_kr": "Korea",
     "region_asia": "Asia",
     "download_error": "This download option is not available",
+    "invalid_repack": "No download link is available for the selected downloader.",
```
Moyasee, line 439: New user facing strings ship four locales in the same commit, so this needs `ru`, `es` and `pt-BR` too. The i18next fallback means nothing breaks, but every non-English user gets an English toast.

Using the wording each locale already has for `downloader` in this namespace:

```json
"invalid_repack": "Для выбранного загрузчика нет доступных ссылок."
```

```json
"invalid_repack": "No hay ningún enlace de descarga disponible para el descargador seleccionado."
```

```json
"invalid_repack": "Nenhum link de download disponível para o downloader selecionado."
```

src/renderer/src/pages/game-details/modals/download-settings-modal.tsx
```diff
@@ -365,7 +366,7 @@ export function DownloadSettingsModal({
       { hasAvailable: boolean; hasUnavailable: boolean }
     >();
 
-    if (repack) {
+    if (repack && Array.isArray(repack.uris)) {
```
Moyasee, line 369: This is the right guard, but its twin in big picture is untouched: `getDownloaderAvailabilityOptions` in `src/big-picture/src/helpers/downloaders.ts` runs `for (const uri of repack.uris)` with no array check, and it is called from a `useMemo` in `src/big-picture/src/components/modals/download-game/index.tsx`, so it throws during render on exactly the repack shape this PR is closing off here.

Two things make it worth folding in rather than leaving for later. That helper already writes `repack.unavailableUris ?? []` a few lines above, so the missing array shape is anticipated for the sibling field and not for this one. And big picture has no error boundary, `src/big-picture/src/main.tsx` renders straight into `createRoot`, so a render throw there takes the whole tree down instead of landing on a fallback page the way the desktop renderer does.

Same one line fix as here. If you would rather keep this PR to the desktop flow, that is fine, but then the title and the body should not claim the remaining call sites, since that is the one that is left.

Moyasee (changes requested):
The guards themselves are right and the early return is wired correctly: `handleStartDownload` has a single call site and it already routes `{ ok: false, error }` to the toast, so `invalid_repack` reaches the user instead of an undefined uri reaching the download path.

Two things before merge, both inline. One is the locale coverage on the new key, the other is that the same unguarded read still exists in big picture, which matters for the "remaining call sites" framing in the title.

One note on the description: it says the non-null assertion and the `string` return type stay as they are, but the diff drops both. Worth updating so the body matches what landed.

sotik11:
All three addressed:

- **Locales**: checked the sibling `downloader` key — it actually ships in 7 locales, not 4, so added `invalid_repack` to all of them (en/ru/es/pt-BR/uk/pl/de/fr). Used your ru/es/pt-BR wording; matched uk/pl/de/fr to each locale's own `downloader` term.
- **Big picture**: guarded `getDownloaderAvailabilityOptions` with the same `Array.isArray(repack.uris)` check.
- **Description**: fixed — the non-null assertion is dropped, and the big picture call site is now listed.

src/big-picture/src/helpers/downloaders.ts
```diff
@@ -116,7 +116,7 @@ export function getDownloaderAvailabilityOptions(
     (value) => typeof value === "number"
   ) as Downloader[];
 
-  for (const uri of repack.uris) {
+  for (const uri of Array.isArray(repack.uris) ? repack.uris : []) {
```
Hachi-R, line 119: `src/big-picture/src/helpers/downloaders.ts:L119` — P1: This guard is never reached for a repack whose `uris` is missing. `src/big-picture/src/hooks/queries/use-game-download-options.hook.ts` first passes the full list to `fetchHosterAvailability`, where `collectHosterAvailabilityUris` still iterates `repack.uris`; that rejection is caught by `fetchDownloadOptions`, and `setNoDownloadOptionsState` clears every repack. Guard or normalize `repack.uris` in both `collectHosterAvailabilityUris` and `applyHosterAvailability` (or normalize once at the API boundary), otherwise one malformed row still removes all Big Picture download options.

Hachi-R (changes requested):
One blocking issue remains in the Big Picture data path.

Hachi-R:
Closing this for now because the Big Picture path still needs the upstream `repack.uris` normalization described in the review. Feel free to reopen this PR whenever you want to address it or continue iterating here.

sotik11:
Reopen wasn't available on this PR (GitHub kept the head stale after the last push and rejected reopen), so I moved the remaining fix — the shared `collectHosterAvailabilityUris` / `applyHosterAvailability` normalization from the final review — to #2704, which carries the full branch. Continuing there.

PR #2519, opened 2026-07-13
fix: translate missing and hardcoded strings in big picture mode

moi952:
Fixes several missing/broken translations in Big Picture mode (French + other locales), and one regression introduced mid-fix that briefly broke the asset tabs.

**Game customization modal**

* "Game Assets" tab label wasn't translatable (edit_game_modal_assets and its siblings _enter_title, _hero, _icon, _logo were stored in the wrong locale section, so t() never found them)
* Moved these keys to the correct section for all 5 supported languages (en, fr, es, pt-BR, ru)

**Library**

* "Last played X ago" stayed in English regardless of language (formatRelativeDate never received the current locale, and Intl.RelativeTimeFormat defaulted to en)
* "Empty collection" / "Add games to this collection to see them here" had no translation
* Collection tabs (All / Moderns / Classics / Favorites / Completed with counts) were hardcoded English

**Game details**

* "N download sources" on the game cards (Popular on Hydra, Trending) was hardcoded English
* "Main Story" / "Main + Sides" / "Completionist" (How Long to Beat) had no French translation, and once added the French strings were too long and wrapped to 2 lines — shortened to Histoire / Extras / Complet
* Review "after playing for X hours" and the "rating"-only fallback were hardcoded English
* Compatibility tab: "Configure per-game overrides for GameMode and MangoHud" had no translation key at all
Launch settings
* "Select Path" / "Clear Path" / "Clear Args" had no translation key at all (not even in English resources)

**Downloads**

* Download source date (upload date) always rendered in the English date format, ignoring the user's language
* "N download options" on the source cards and the per-source removal toast were hardcoded English

**Profile**

* Same "Last played X ago" bug on the profile activity list
* Playtime labels ("1 minute played", "2 hours played", etc.) were built with a second, separate, never-translated formatter — now reuse the existing translated one used elsewhere in the app
* Favorite game "X played" label had the same issue

**Settings**

* "Achievement notification position" dropdown values (Top Left, Top Center, ...) were hardcoded English instead of reusing the already-translated position labels

**Regression caught during the fix**

Moving the asset tab keys (edit_game_modal_hero/logo/icon/...) out of the legacy section only for en/fr (and not es/pt-BR/ru) broke the "Logo"/"Hero" tab labels — they started rendering as their raw translation key instead of text. Root cause: a DOM-based fallback translation pass merges lookup tables across all languages, so a stale entry left in one language's legacy table can shadow another language's fixed entry. Fixed by moving the keys consistently across all 5 languages.

**Fill in the PR content:**


src/big-picture/src/locales/es/translation.json
```diff
@@ -138,11 +138,6 @@
     "Download Sources": "Fuentes de descarga",
```
Moyasee, line 138: This file (and `pt-BR` and `ru`) only picked up the 5 moved `edit_game_modal_*` keys, not the 28 new ones added to `en` and `fr`. So the library tabs, the HLTB labels, the download source and option counts, the review strings, the launch-tab path buttons, the playtime labels and the empty-collection copy all stay English for Spanish, Brazilian Portuguese and Russian users. It fails soft thanks to `fallbackLng: "en"`, so nothing renders as a raw key, but the bug this PR is fixing stays unfixed for three of the five languages.

Could you add the same 28 keys here, in `pt-BR` and in `ru`? For `ru`, the four plural keys (`download_sources_count`, `download_options_count`, `download_options_unavailable`, `playtime_seconds`) need the `_few` and `_many` forms too, the way the existing Russian plurals in this file already do.

src/big-picture/src/pages/settings/downloads-sources-section.tsx
```diff
@@ -231,9 +231,10 @@ export function DownloadsSourcesSection({
         (downloadSource) => downloadSource.id === downloadSourceId
       );
       const removedDownloadOptionsLabel = removedSource
-        ? `${formatNumber(removedSource.downloadCount)} download option${
-            removedSource.downloadCount === 1 ? "" : "s"
-          } ${removedSource.downloadCount === 1 ? "is" : "are"} no longer available.`
+        ? t("download_options_unavailable", {
```
Moyasee, line 234: This passes `formattedCount`, but the `en` string for `download_options_unavailable` interpolates `{{count}}`, so the formatted value is never read and the number renders unseparated. This used to go through `formatNumber` here, so a source with 45231 options now reads "45231 download options are no longer available." instead of "45,231".

The sibling key you added, `download_options_count`, does use `{{formattedCount}}` and is correct, so it is just this one out of step. Switching the `en` and `fr` strings to `{{formattedCount}}` would line them up.

src/big-picture/src/pages/downloads/use-big-picture-downloads-page-data.ts
```diff
@@ -531,7 +543,14 @@ export function useBigPictureDownloadsPageData() {
         game,
       };
     });
-  }, [completedGames, renderTick, seedingStatuses, userPreferences]);
+  }, [
+    completedGames,
+    formatTime,
```
Moyasee, line 548: `useDate` builds its return value as a fresh object literal of inline arrows with no `useCallback`, so `formatTime` has a new identity on every render. Adding it to this dependency array means this memo now recomputes and hands back a new array every render, where before it was keyed only on `completedGames`, `renderTick`, `seedingStatuses` and `userPreferences`.

Wrapping the functions in `useDate` in `useCallback` would fix it at the source. Same thing now applies to `formatDate` in `download-source-option`, which moved from a stable module-level import into the hook. `t` is fine, i18next keeps that identity stable.

src/big-picture/src/helpers/date.ts
```diff
@@ -67,20 +67,21 @@ export function formatRelativeDate(
...
   const diffSec = Math.floor(diffMs / 1000);
+  const locale = options?.locale ?? "en";
 
-  if (diffSec < 60) return "just now";
+  const rtfAuto = new Intl.RelativeTimeFormat(locale, { numeric: "auto" });
 
-  const rtf = new Intl.RelativeTimeFormat(options?.locale ?? "en", {
-    numeric: "always",
-  });
+  if (diffSec < 60) return rtfAuto.format(0, "second");
```
Moyasee, line 74: Minor copy change worth knowing about: `rtfAuto.format(0, "second")` resolves to "now" in English rather than "just now", so the hero label reads "Last played now" where it used to say "Last played just now". Using `numeric: "auto"` is the right call and `-1 day` / `-1 year` correctly give "yesterday" and "last year", so this is just a heads up on the English wording rather than something to undo.

Moyasee (changes requested):
Good catch on all of these, and the diagnosis in the description is right: the asset tab keys really were in the wrong section, and the cross-language shadowing you describe is real, since `reverseTranslations` in `i18n.tsx` merges the lookup tables across every language.

The code side is in good shape. I checked every key this PR introduces against the `en` bundle and they all resolve, every `_one`/`_other` key gets a `count`, every interpolation variable is supplied at the call site, and the locale plumbing is correct (i18next codes going to `Intl.RelativeTimeFormat`, date-fns locale objects staying inside `useDate`). All five greptile points are genuinely fixed at head, not just claimed. The `Math.max(0, ...)` clamp on the store cards survived the rewrite too.

One blocking thing though. This adds 33 new keys, and 28 of them only landed in `en` and `fr`. They are absent from `es`, `pt-BR` and `ru`:

```
exact.Main Story, exact.Main + Sides, exact.Completionist,
exact.Empty collection, exact.Add games to this collection to see them here.,
exact.Try adjusting your search terms or filters.,
format.library_tab_{all,moderns,classics,favorites,completed},
format.download_sources_count_{one,other}, format.download_options_count_{one,other},
format.game_review_played_for, format.game_review_rating_only,
format.select_executable_path, format.clear_executable_path, format.clear_launch_args,
format.additional_options_description, format.compact_hours, format.compact_minutes,
format.playtime_seconds_{one,other}, format.never_played,
format.recently_played_fallback, format.downloads_finished_at
```

Because `fallbackLng` is `en` and the exact lookup ends in `?? sourceText`, nothing breaks, it just silently falls back to English. Which means for Spanish, Brazilian Portuguese and Russian users these strings stay exactly as broken as they are today, and the PR only actually delivers the fix for French. Same rule as the other locale PRs: new keys need to land in all five. Note `ru` will want its `_few` / `_many` plural forms for the four plural keys, not just `_one` / `_other`.

Three minor things inline. Nothing else blocking.

Small aside, no action needed: `formatPlayedTime` in `helpers/date.ts` still has hardcoded English in it and after this PR its only remaining caller is the component-lab demo page, so it is basically dead now and could go in a follow-up.

moi952:
> Good catch on all of these, and the diagnosis in the description is right: the asset tab keys really were in the wrong section, and the cross-language shadowing you describe is real, since `reverseTranslations` in `i18n.tsx` merges the lookup tables across every language.
> 
> The code side is in good shape. I checked every key this PR introduces against the `en` bundle and they all resolve, every `_one`/`_other` key gets a `count`, every interpolation variable is supplied at the call site, and the locale plumbing is correct (i18next codes going to `Intl.RelativeTimeFormat`, date-fns locale objects staying inside `useDate`). All five greptile points are genuinely fixed at head, not just claimed. The `Math.max(0, ...)` clamp on the store cards survived the rewrite too.
> 
> One blocking thing though. This adds 33 new keys, and 28 of them only landed in `en` and `fr`. They are absent from `es`, `pt-BR` and `ru`:
> 
> ```
> exact.Main Story, exact.Main + Sides, exact.Completionist,
> exact.Empty collection, exact.Add games to this collection to see them here.,
> exact.Try adjusting your search terms or filters.,
> format.library_tab_{all,moderns,classics,favorites,completed},
> format.download_sources_count_{one,other}, format.download_options_count_{one,other},
> format.game_review_played_for, format.game_review_rating_only,
> format.select_executable_path, format.clear_executable_path, format.clear_launch_args,
> format.additional_options_description, format.compact_hours, format.compact_minutes,
> format.playtime_seconds_{one,other}, format.never_played,
> format.recently_played_fallback, format.downloads_finished_at
> ```
> 
> Because `fallbackLng` is `en` and the exact lookup ends in `?? sourceText`, nothing breaks, it just silently falls back to English. Which means for Spanish, Brazilian Portuguese and Russian users these strings stay exactly as broken as they are today, and the PR only actually delivers the fix for French. Same rule as the other locale PRs: new keys need to land in all five. Note `ru` will want its `_few` / `_many` plural forms for the four plural keys, not just `_one` / `_other`.
> 
> Three minor things inline. Nothing else blocking.
> 
> Small aside, no action needed: `formatPlayedTime` in `helpers/date.ts` still has hardcoded English in it and after this PR its only remaining caller is the component-lab demo page, so it is basically dead now and could go in a follow-up.

Thanks for the thorough review!

The goal of this PR was to add the missing translation keys so the app could be fully be translated into French, since I’m a native French speaker. I don’t speak Spanish, Brazilian Portuguese or Russian, so I don’t feel comfortable providing translations for those languages.

If it’s the project’s convention that new keys must exist in every locale file, I’m happy to add the missing keys to es, pt-BR and ru and use the English strings as placeholders until a native speaker can translate them.

It’s getting late here and I’m heading to bed now, but I’ll update the PR with those changes when I get up tomorrow if that works for you.

Moyasee (approved):
That is a completely fair point, and no, do not invent translations for languages you do not speak. Dropping the block.

Do not bother with English placeholders either. `fallbackLng` is already `en` and the exact-text lookup ends in `?? sourceText`, so a missing key renders the English string anyway. Placeholders would look identical at runtime and would just make those keys look translated to whoever opens the file next. Better to leave the gap honest and let native speakers fill in `es`, `pt-BR` and `ru` in a follow-up. French being complete is a real win on its own, and the keys now exist in `en` for anyone who wants to pick the others up.

One thing I would still like before this goes in, and it is a two minute change: the `download_options_unavailable` string interpolates `{{count}}` while the call site passes `formattedCount`, so that toast lost its thousands separator ("45231 download options are no longer available" instead of "45,231"). That is an actual regression rather than a style point, and the sibling key you added right next to it, `download_options_count`, already does it correctly with `{{formattedCount}}`. Just needs the `en` and `fr` strings switched over.

The other two inline notes are optional. The `useDate` memo one is worth doing at some point but it is not new breakage you introduced so much as a sharp edge you stepped on, and the "just now" to "now" copy change is only a heads up.

Good PR. Get some sleep.

moi952:
> That is a completely fair point, and no, do not invent translations for languages you do not speak. Dropping the block.
> 
> Do not bother with English placeholders either. `fallbackLng` is already `en` and the exact-text lookup ends in `?? sourceText`, so a missing key renders the English string anyway. Placeholders would look identical at runtime and would just make those keys look translated to whoever opens the file next. Better to leave the gap honest and let native speakers fill in `es`, `pt-BR` and `ru` in a follow-up. French being complete is a real win on its own, and the keys now exist in `en` for anyone who wants to pick the others up.
> 
> One thing I would still like before this goes in, and it is a two minute change: the `download_options_unavailable` string interpolates `{{count}}` while the call site passes `formattedCount`, so that toast lost its thousands separator ("45231 download options are no longer available" instead of "45,231"). That is an actual regression rather than a style point, and the sibling key you added right next to it, `download_options_count`, already does it correctly with `{{formattedCount}}`. Just needs the `en` and `fr` strings switched over.
> 
> The other two inline notes are optional. The `useDate` memo one is worth doing at some point but it is not new breakage you introduced so much as a sharp edge you stepped on, and the "just now" to "now" copy change is only a heads up.
> 
> Good PR. Get some sleep.

Hi ! Done ! 😊

Hachi-R (changes requested):
Thanks for putting this together. The overall direction makes sense, and I didn't find any major side effects outside the changed areas.

Before merging, could you please fix these:

- `use-format.hook.ts:L25`: the sub-minute playtime fix is incomplete. `playtime_seconds` only exists in English and French, so Spanish, Brazilian Portuguese, and Russian fall back to English. Please add the missing translations; Russian needs `_one`, `_few`, `_many`, and `_other`.
- `use-date.hook.ts:L42-98`: `useDate` creates new function references on every render. This invalidates the new memos in the downloads page and download-source cards. Please stabilize these functions with `useCallback`/`useMemo`.

I also checked Greptile's comments. These were valid:

- Sub-minute playtime could display as `0 minutes`.
- Completed-download times used the US time format.
- `Never played` remained hardcoded in English.
- Profile playtime still used hardcoded units.
- Hydra Cloud subscription copy remained partially hardcoded.

Most of these are fixed in the current head. The remaining problem is the incomplete localization of the new seconds label mentioned above.

SonarQube is green, and all current checks are passing.

moi952:
> Thanks for putting this together. The overall direction makes sense, and I didn't find any major side effects outside the changed areas.
> 
> Before merging, could you please fix these:
> 
> * `use-format.hook.ts:L25`: the sub-minute playtime fix is incomplete. `playtime_seconds` only exists in English and French, so Spanish, Brazilian Portuguese, and Russian fall back to English. Please add the missing translations; Russian needs `_one`, `_few`, `_many`, and `_other`.
> * `use-date.hook.ts:L42-98`: `useDate` creates new function references on every render. This invalidates the new memos in the downloads page and download-source cards. Please stabilize these functions with `useCallback`/`useMemo`.
> 
> I also checked Greptile's comments. These were valid:
> 
> * Sub-minute playtime could display as `0 minutes`.
> * Completed-download times used the US time format.
> * `Never played` remained hardcoded in English.
> * Profile playtime still used hardcoded units.
> * Hydra Cloud subscription copy remained partially hardcoded.
> 
> Most of these are fixed in the current head. The remaining problem is the incomplete localization of the new seconds label mentioned above.
> 
> SonarQube is green, and all current checks are passing.

It's all set; I've correctly set up the Russian-specific variants and added them to ES and PT-BR—though the text is in English.

src/big-picture/src/pages/settings/downloads-sources-section.tsx
Hachi-R, line 234: `src/big-picture/src/pages/settings/downloads-sources-section.tsx:L234` — P2: `formattedCount` only takes effect in `en` and `fr`; the `es`, `pt-BR`, and `ru` variants still interpolate `{{count}}`, so large option counts lose locale-aware grouping in both removal flows. Change those locale strings to use `{{formattedCount}}`.

Hachi-R (changes requested):
One remaining localization regression.

src/big-picture/src/pages/settings/downloads-sources-section.tsx
moi952: Yes, I just realized that... I fixed it.

Hachi-R (approved):
LGTM

Issue #2436, opened 2026-07-04
Why does it need to run install.bat every time a game is launched?


WilliamBakersGithub:
It simply doesn't make sense. Shouldn't this dependency stuff only run once? It just makes the launch process getting slow.

Moyasee:
You're right, it shouldn't run every launch. It's meant to run once but a flag isn't sticking, and it's blocking the launch. Fix incoming. Thanks for the report.

Moyasee:
Closing this since this behavior has been fixed in #2526 and would be available in the next update v4.0.7

PR #2526, opened 2026-07-15
fix: only download common redistributables on Windows

Moyasee:

**Fill in the PR content:**

Closes LBX-899. Addresses #2436 ("Why does it need to run install.bat every time a game is launched?").

### What was actually wrong

`install.bat` ends with `pause`. The launcher runs it hidden and elevated, so there is no window to receive a keypress and the elevated process sits there forever. Every launch that reached the install path leaked another stuck process, and when the UAC prompt was dismissed the "Installation complete" line was never written, so `commonRedistPassed` never persisted and the next launch prompted again.

**The fix for that is a one line change in `hydra-common-redist`: remove the trailing `pause`.** Nothing in the launcher was required to fix it.

It also does not need a migration. Installs that already passed the preflight never run the installer again, and installs that never passed it re-download the current `install.bat` on startup and run the fixed version on the next launch.

### What is left in this PR

Only the startup download gate:

- Skip `downloadCommonRedist()` on Linux and macOS. These are Windows redistributables, so other platforms were downloading about 131MB of installers they can never use.
- Drop a stale comment on the preflight call that claimed the try/catch keeps the launch from blocking. It does not, the call is awaited.

### Earlier versions of this PR

The first pass made the preflight non-blocking and marked the XNA Framework check optional, then added a `redistManifestVersion` constant to control when `install.bat` gets re-fetched. All of that is reverted.

The version constant was the worst of it: it lived here while `install.bat` lives in `hydra-common-redist`, so keeping them in sync was manual with nothing enforcing it, and anyone editing `install.bat` without remembering to bump it would silently stop clients from picking the change up. It was also solving a non-problem, since `install.bat` is about 1KB and re-fetching it on startup is what makes fixes propagate on their own. Thanks for pushing back on it in review.

Preflight behaviour is therefore unchanged from `main`: it still runs before launch and still treats every system check, XNA included, as required.


Moyasee:
Good catch on the install.bat propagation gap. Added a redist manifest version (`redistManifestVersion`): install.bat is re-fetched once whenever it's bumped, otherwise existing files are skipped. So the upstream pause fix reaches existing installs by bumping the version, without going back to re-downloading on every startup.

@greptile review

Moyasee:
Closed the manifest-bump gap too: `downloadCommonRedist` now resets `commonRedistPassed` when the version bumps, so the next launch re-verifies and actually re-runs the refreshed installer instead of short-circuiting on a stale pass flag.

@greptile review

Hachi-R (approved):
No blocking findings.

Moyasee:
@greptile review

PR #2527, opened 2026-07-15
Feature/lbx 942 Local parser for the Ludusavi manifest

Hachi-R:


Moyasee (approved):
This is the reworked version of the earlier manifest parser PR and the feedback from that round is all addressed:

- Download is now bounded by a 30s timeout on a shared static `reqwest::Client`, so a stalled connection can no longer hold the per-path lock indefinitely. This also folds in the "new client per call" cleanup.
- The normalized lookup now respects candidate priority (object_id > remote_id > title) instead of returning the first manifest-order match, so a lower-priority fuzzy match can't win over a higher-priority one.
- Version literal and the rule kind/source strings are named constants now.

Rest of the pipeline (atomic temp+rename writes, per-path mutex, stale fallback on failure) still reads well. Approving.

Two optional items from before still stand, neither blocking:

- `reqwest = "0.13"` pulls the full default feature set (`aws-lc-sys`, `quinn`/HTTP3) for what is a single YAML GET. Builds pass on both platforms so it's purely binary size / build cost, but `default-features = false` + `rustls-tls` would trim it.
- The integration tests still hit the live CDN and assert on Balatro specifics. Nothing runs `cargo test` in CI so they don't gate anything today, but they'll fail offline or on manifest drift. A `#[ignore]` or a checked-in fixture would make the core logic testable without the network.

One tiny thing while you're in there: the normalized fallback in `find_manifest_entry` re-normalizes every game key once per candidate. It only runs when the exact match misses (rare, since object_id is the map key), so not worth blocking on, just flagging.

Hachi-R:
Superseded by #2606

PR #2528, opened 2026-07-15
Feature/lbx 943 Native path resolution

Hachi-R:


Moyasee (approved):
Big step up from the earlier path resolution pass, and the feedback from that round is all addressed:

- Every token the live manifest actually uses is handled now, including `<xdgData>`, `<xdgConfig>`, `<root>`, `<osUserName>`, and `<winDir>` that were slipping through before.
- The unresolved-token handling now fails closed. The `<skip>` sentinel plus `has_unresolved_placeholder` drops any candidate that still carries an unresolved `<...>`, and `tokens_in_path` reports what was left over instead of emitting a literal-token path as if it resolved. That was the main thing I was worried about.
- The Windows drive letter is no longer hardcoded, `known-folders` handles it on Windows with a home-drive-derived fallback elsewhere, and it's correctly gated behind `cfg(windows)` so the Linux/mac builds don't pull it in.
- `proton_path` is gone and `steam_path` is actually wired into the Proton compatdata roots now.

The glob-based approach (drive_*, users/*, compatdata/<appid>/pfx, modern + legacy layouts) reads well and the escaping via `globset::escape` on the literal segments looks right. Nice work. Approving, assuming the builds come back green.

A few small non-blocking things:

- `expand_store_user_ids` drops the candidate entirely when a path contains `<storeUserId>` but `steam_user_ids` is empty, so those rules end up reported as unresolved. Given the rest of the design leans on globs, was a `*` fallback for the unknown-user case considered? Fine to keep the fail-closed behavior if the caller always supplies the ids, just checking it's deliberate.
- `has_unresolved_placeholder` only looks for `<`, so a stray `%SOMETHING%` env token (anything beyond the two you replace) would pass through unescaped and unflagged. The manifest doesn't seem to use arbitrary `%` tokens today so this is theoretical, just noting the guard is `<`-only.
- `TokenValues` is rebuilt per candidate per rule. It's all context-derived and could be hoisted, but rule counts per game are tiny so not worth blocking on.
- Same as the manifest PR, the integration test reaches the live CDN with no `#[ignore]`. Nothing runs `cargo test` in CI so it doesn't gate anything, but it'll fail offline.

Moyasee (changes requested):
Re-reviewing after the new commits. The delta since my last pass is good: `<storeUserId>` now resolves to a `*` wildcard with a `dynamic` flag (and the `steam_user_ids` input is gone), and `has_unresolved_placeholder` now catches arbitrary `%X%` tokens with tests. Both were the open questions from last round, so thanks for that. The new `ResolvedCloudSavePath` metadata and the `scan_root` derivation read cleanly.

One thing I want to flag before this goes in, since it's the root cause of a failure that shows up in the scanner PR (2529).

### `literal()` doesn't neutralize `{` and `}`

`literal()` leans on `globset::escape`, which only escapes `?`, `*`, `[`, and `]` (it wraps them in character classes). It leaves `{` and `}` untouched:

```rust
// globset::escape
'?' | '*' | '[' | ']' => { escaped.push('['); escaped.push(c); escaped.push(']'); }
c => { escaped.push(c); }
```

So any real path segment that legitimately contains a curly brace, a Steam library folder, wine prefix, install dir, or home path the user chose, comes out of `literal()` with the brace intact. Downstream that brace is then interpreted as glob alternation rather than a literal. In the scanner it lands in `expand_braces`, and a single-alternative brace like `{Deluxe}` hits the `brace must contain alternatives` error, which propagates and fails the whole scan batch (details in my 2529 review).

The manifest's own intentional `{a,b}` braces should still expand, so the fix is to escape braces that come from real filesystem literals while leaving the manifest pattern's braces alone. Simplest is to extend `literal()` to also wrap `{`/`}` after `globset::escape`, since `literal()` is only ever applied to concrete path values, not to the manifest glob. Worth double-checking the escaped form you pick round-trips through the scanner's `expand_braces` (it's backslash-aware, not character-class-aware).

Trigger is uncommon (needs a brace in one of the resolved literals) but when it hits it takes out the entire game's scan, and the fix is cheap. Everything else here looks good.

Moyasee (approved):
The brace escaping is sorted now. `literal()` wraps `{`/`}` as `[{]`/`[}]`, matching the character-class style `globset::escape` already uses for the other metacharacters, and the paired change in the scanner's `expand_braces` keeps those from being read as alternation. Tests at both ends cover it. That was the only thing holding this up. Approving.

Hachi-R:
Superseded by #2606

PR #2529, opened 2026-07-15
Feature/lbx 944 Native cloud save scanner

Hachi-R:


Moyasee (changes requested):
The scanner design is nice: brace expansion up front, globetter for matching, walkdir for recursive collection, grouping by scan root, dedup by canonical path, and a depth cap. Tests are hermetic (tempdir, no network) and cover files, dirs, recursive globs, case sensitivity, the depth limit, and symlinks. Two things I'd want addressed before merge, plus a couple of minor notes.

### One filesystem error fails every rule

Scan errors propagate all the way out and abort the whole batch. In `scan_directory`:

```rust
let entry = entry.map_err(|error| format!("cloud_save_filesystem_error: {error}"))?;
```

WalkDir yields a per-entry `Err` for things like a permission-denied subdirectory but keeps iterating, so this is recoverable, yet the `?` turns it into an early return. That bubbles through `scan_resolved_path` and then `scan_rules`' `collect::<Result<_,_>>()`, so a single unreadable directory (or a broken symlink surfaced by `follow_links`, or a `metadata`/`canonicalize` failure on one file) fails the scan for every rule of the game, not just that path. For a "walk the filesystem looking for saves" feature this is fragile, those conditions are normal on real machines. I'd skip-and-continue on per-entry and per-file IO errors (optionally collecting them) rather than propagating, and reserve hard errors for genuinely invalid input like a bad glob.

### Literal curly braces in resolved paths break the scan

Paths coming out of path resolution can contain literal `{`/`}` because `literal()` there uses `globset::escape`, which doesn't escape braces (details in my review on the path resolution PR, 2528). When such a path reaches `glob_matches`, `has_glob_pattern` sees the `{` and routes it through `expand_braces`, and a literal single-alternative brace like `{Deluxe}` returns:

```
cloud_save_invalid_glob: brace must contain alternatives
```

which then fails the whole batch per the point above. The real fix belongs in path resolution (escape braces in `literal()`), but this PR is where it surfaces, so flagging the dependency. It'd also be worth a scanner test with a real directory name containing braces to lock the behavior down once the escaping is fixed.

### Minor

- `ResolvedCloudSavePath.dynamic` isn't read anywhere in the scanner, only `path`, `case_sensitive`, and `scan_root` are used. If JS consumes `dynamic` that's fine, just confirming it's intentional that the scanner ignores it.
- `glob_matches` stats the path via `metadata` in the no-glob fast path, then `scan_resolved_path` stats each match again. Minor double stat.
- `follow_links(true)` plus the depth cap bounds symlink loops, but there's no breadth bound, a glob that resolves to a broad root would enumerate a lot. Probably fine given how specific the resolved paths are, just noting.

Moyasee (changes requested):
The literal-brace handling is solid now. Making `expand_braces` character-class aware (`is_in_character_class` / `is_brace_syntax`) so `[{]` and `[}]` survive as literals, paired with the escaping fix in path resolution, is exactly right, and the real-directory test with `{Deluxe}` locks it in. Nice.

The other item from last round is still open, though: a single filesystem error still fails every rule. `scan_directory` propagates per-entry WalkDir errors:

```rust
let entry = entry.map_err(|error| format!("cloud_save_filesystem_error: {error}"))?;
...
absolute_path: canonical_path(entry.path())?,
```

and `scan_resolved_path` does the same on `metadata`, so one permission-denied subdirectory, a broken symlink surfaced by `follow_links`, or a single `canonicalize` failure bubbles up through `scan_rules`' `collect::<Result<_,_>>()` and fails the scan for the whole game rather than just skipping that entry. Those conditions are normal on real machines, and for a "walk the filesystem for saves" feature returning the saves you did find beats returning nothing. I'd skip-and-continue on per-entry and per-file IO errors and keep the hard error for genuinely bad input like an invalid glob. That's the one thing left before this is good to go.

The minor notes from before (unused `dynamic`, the double `metadata` stat, no breadth bound) are all still optional, no need to block on those.

Moyasee (approved):
lgtm

Hachi-R:
Superseded by #2606

PR #2532, opened 2026-07-15
Feature/lbx 945 Add native save hashing

Hachi-R:


Moyasee (approved):
Clean and well-tested. blake3 streaming through `update_reader` keeps large saves off the heap, the rayon pool is bounded and named, and the incremental cache invalidates correctly on path/size/mtime/hash-validity. The aggregate hash canonicalizes by sorting then serializing a fixed field order, so it's order-independent, and the tests cover both that and sensitivity to each field. Nice. Approving, assuming the builds come back green.

A few non-blocking notes:

- `hash_files` fails the whole batch if any single file errors (the `collect::<Result<_,_>>`). Calling that out because I flagged the same shape in the scanner PR, but here I think it's defensible: a snapshot integrity token arguably should be all-or-nothing, and the batch path isn't even exposed to JS yet (the NAPI surface only hits `hash_files` with a single file via `hash_local_save_file`). Worth deciding intent before the batch entrypoint lands though, since a file vanishing between scan and hash (TOCTOU) would fail the whole snapshot rather than the one file.
- Related: the cache machinery in `batch.rs` (`LocalFileHashCacheEntry` matching, dedup, cache rebuild) isn't reachable from JS in this PR, `hash_local_save_file` always passes an empty cache and one path. Fine as staged work, just noting the single-file NAPI path routes through rayon + BTreeSet + HashMap for one file, which is more machinery than that call needs.
- `build_snapshot_aggregate_hash` is a sync `#[napi]`, so it runs on the JS thread. It's metadata-only serialization plus one hash so probably quick, but if a snapshot can carry a large file list it'd block the event loop; `spawn_blocking` like the sibling function would keep it off the main thread.

Hachi-R:
Superseded by #2607

PR #2533, opened 2026-07-15
Feature/lbx 946 Add native local snapshot

Hachi-R:


Moyasee (approved):
This is the piece that ties the hashing work together and it's nicely defensive. The guardrail layering is the highlight: cap and dedup up front, stat before hashing so an oversized set never gets read, then `validate_hashed_files` comparing the hash-time metadata against the pre-hash snapshot, and `validate_built_files` re-statting once more. That's a real effort at catching mid-snapshot changes, and it also answers the question I raised on the hashing PR: the all-or-nothing behavior of `hash_files` is the right call here, a snapshot integrity token should fail rather than silently drop a file. Sharing `format_modified_at` so the cache and the guardrails compare identical timestamp strings is the correct detail. Determinism is covered by the order-independence test. Approving, assuming builds go green.

Two minor, non-blocking notes:

- A physical file referenced by two different logical `(raw_path, relative_path)` entries gets its size counted twice toward `MAX_SNAPSHOT_TOTAL_SIZE_BYTES`, since `prepare_snapshot_files` and `validate_built_files` both sum per input entry while `hash_files` dedups by `absolute_path`. It only ever makes the cap stricter so it's safe, just flagging in case you'd rather sum distinct absolute paths.
- Three stats per file across the pipeline (prepare, hash, validate_built) is fine at 500 files and I get that the extra re-stat tightens the TOCTOU window. Worth being aware it's still best-effort, a change after the final stat but reflected in the already-computed hash can't be detected, which is inherent to snapshotting without locks and totally acceptable here.

Hachi-R:
Superseded by #2607

PR #2534, opened 2026-07-15
Feature/lbx 947 Add native snapshot comparison

Hachi-R:


native/hydra-native/src/cloud_save/comparison/compare.rs
```diff
@@ -0,0 +1,103 @@
...
+pub fn compare_hashes(
+    local_snapshot_hash: &str,
+    local_snapshot_file_count: u32,
+    base_snapshot_hash: Option<&str>,
+    remote_snapshot_hash: Option<&str>,
+) -> SnapshotComparisonState {
+    let Some(remote_snapshot_hash) = remote_snapshot_hash else {
+        return SnapshotComparisonState::Untracked;
+    };
+
```
Moyasee, line 12: This returns before the `local == remote` check below, so if the active remote snapshot ever has 0 files (same empty aggregate hash as an empty local), we'd report `RemoteAhead` instead of `Synced`. Probably can't happen if empty remote snapshots are never created upstream, but worth confirming.

native/hydra-native/src/cloud_save/comparison/compare.rs
```diff
@@ -0,0 +1,103 @@
...
+    if local_snapshot_hash == remote_snapshot_hash {
+        return SnapshotComparisonState::Synced;
+    }
+
+    match (
+        local_snapshot_hash != base_snapshot_hash,
+        remote_snapshot_hash != base_snapshot_hash,
+    ) {
+        (false, false) => SnapshotComparisonState::Synced,
```
Moyasee: This arm is unreachable: `local == base && remote == base` implies `local == remote`, which already returned `Synced` above. Harmless as a defensive default, just flagging it's dead.

native/hydra-native/src/cloud_save/comparison/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+        input.local_snapshot_file_count,
+        input.base_snapshot_hash.as_deref(),
+        active_remote_snapshot
+            .as_ref()
+            .map(|snapshot| snapshot.aggregate_hash.as_str()),
+    );
+
+    Ok(CompareGameSnapshotsResult {
+        state,
+        active_remote_snapshot,
```
Moyasee, line 26: `Untracked` counts as out of sync here. For empty local + no remote (file_count 0, no remote snapshot) we return `Untracked` -> `is_out_of_sync = true`, which could trigger a no-op upload. Is that intended, or does the caller guard the empty-local case?

Moyasee (commented):
Clean and well-tested, matches the existing cloud_save module layout. Nothing blocking here, just a couple of nits and one question on the out-of-sync semantics.

native/hydra-native/src/cloud_save/comparison/compare.rs
```diff
@@ -0,0 +1,104 @@
...
+) -> SnapshotComparisonState {
+    let Some(remote_snapshot_hash) = remote_snapshot_hash else {
+        return SnapshotComparisonState::Untracked;
+    };
+
+    if local_snapshot_file_count == 0 {
+        return SnapshotComparisonState::RemoteAhead;
+    }
+
+    let Some(base_snapshot_hash) = base_snapshot_hash else {
```
Moyasee, line 17: With no base hash we return `Untracked` even when local and remote are byte identical, since the equality check below never runs. Hoisting the `local_snapshot_hash == remote_snapshot_hash` check above this guard would report `Synced` for that case, which seems closer to reality.

native/hydra-native/src/cloud_save/comparison/compare.rs
```diff
@@ -0,0 +1,104 @@
...
+    local_snapshot_hash: &str,
+    local_snapshot_file_count: u32,
+    base_snapshot_hash: Option<&str>,
+    remote_snapshot_hash: Option<&str>,
+) -> SnapshotComparisonState {
+    let Some(remote_snapshot_hash) = remote_snapshot_hash else {
+        return SnapshotComparisonState::Untracked;
+    };
+
+    if local_snapshot_file_count == 0 {
```
Moyasee, line 13: Still on the earlier point about this running before the equality check: one more angle is local deletion. If the user wipes their saves and remote still matches base, the true state is LocalAhead but we report RemoteAhead. Restoring rather than deleting is the safe direction, so I'm fine with it, but it does mean a deletion can never propagate to remote. Worth a line in the docs or a comment so nobody treats it as a bug later.

native/hydra-native/src/cloud_save/comparison/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+        is_out_of_sync: matches!(
```
Moyasee, line 27: Now that `Untracked` is excluded, what does the caller do with it? Scenario I have in mind: fresh install where the user has local saves and there's an active remote snapshot, but no base row yet. That lands on `Untracked` with `is_out_of_sync: false`, so if the caller branches on `is_out_of_sync` the user never gets prompted despite a genuine divergence. If the caller branches on `state` instead then this is fine, just want to confirm.

native/hydra-native/src/cloud_save/comparison/select_remote.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub fn select_active_snapshot(
+    snapshots: Vec<CloudSaveRemoteSnapshot>,
+) -> Result<Option<CloudSaveRemoteSnapshot>, &'static str> {
+    let mut active_snapshots = snapshots
+        .into_iter()
+        .filter(|snapshot| snapshot.status == "active");
+    let active_snapshot = active_snapshots.next();
+
+    if active_snapshots.next().is_some() {
```
Moyasee, line 11: Minor: a data anomaly on the server side takes out the whole comparison for that game. Taking the newest by `created_at` (or just the first, with a warning) would degrade a bit more gracefully than erroring out.

native/hydra-native/src/cloud_save/comparison/compare.rs
```diff
@@ -0,0 +1,104 @@
...
+        SnapshotComparisonState::Conflict
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn compares_all_snapshot_states() {
```
Moyasee, line 42: Could use a case for `base: None` with local and remote matching, and something asserting the `state` -> `is_out_of_sync` mapping added in the last commit. That mapping is the part most likely to drift if a new state gets added.

Moyasee (commented):
Follow-up on the last two commits. The dead arm and the `Untracked` -> `is_out_of_sync` bit are both sorted, thanks. A few things still open, mostly around what happens when there's no base hash.

Hachi-R:
Superseded by #2607

PR #2536, opened 2026-07-15
feat: add Greek translation

NickCh11:
## Summary

- add a complete Greek (`el`) locale for the desktop application
- add a complete Greek locale for Big Picture (796 strings)
- register Greek in both locale systems and in the Big Picture language resolver
- synchronize the desktop locale with the latest `main` keys and address the review feedback

## Why

Greek-speaking users can use both Hydra interfaces in Greek, including Big Picture mode.

## Validation

- confirmed all 1,615 desktop locale keys are present in the Greek locale
- confirmed all 796 Big Picture locale strings are present in the Greek locale
- confirmed interpolation placeholders match each English source
- `corepack yarn prettier --check src/locales/el/translation.json src/big-picture/src/locales/el/translation.json src/big-picture/src/locales/index.ts src/big-picture/src/i18n.tsx src/shared/language-flags.ts`
- `corepack yarn typecheck`


NickCh11:
@greptile

src/locales/index.ts
```diff
@@ -35,6 +36,7 @@ export default {
   "pt-PT": ptPT,
   en,
   de,
+  el,
```
Hachi-R, line 39: `src/locales/index.ts:L39` — P2: Registering `el` exposes Greek in both language pickers, but `getDateLocale` still falls through to `enUS`, so relative dates remain in English. Map `el` to the `date-fns` Greek locale in `src/shared/index.ts` and `src/big-picture/src/hooks/use-date.hook.ts`.

Hachi-R, line 39: `src/locales/index.ts:L39` — P3: `el` is missing from `LANGUAGE_FLAG_MAP`, so the Big Picture language picker shows the generic globe instead of the Greek flag. Add `el: "GR"` in `src/shared/language-flags.ts`.

Hachi-R (dismissed):
Two localization gaps before merge.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1637 @@
...
+    "additional_options": "Πρόσθετες επιλογές",
+    "proton_version_auto": "Αυτόματη ρύθμιση (παγκόσμια προεπιλογή ή προεπιλογή umu)",
+    "proton_source_umu_default": "Προεπιλεγμένη επιλογή umu",
+    "proton_source_steam": "Εγκαταστάθηκε μέσω του Steam",
+    "proton_source_compatibility_tools": "Εγκαταστάθηκε στο Steam, στο φάκελο compatibilitytools.d",
+    "settings_category_general": "Γενικά",
+    "settings_category_locations": "Τοποθεσίες",
+    "settings_category_general_description": "Διαδρομή εκτελέσιμου αρχείου, επιλογές εκκίνησης και συντομεύσεις",
+    "settings_category_assets": "Στοιχεία",
+    "settings_category_assets_description": "Προσαρμόστε τον τίτλο του παιχνιδιού, το εικονίδιο, το λογότυπο και τις εικόνες των ηρώων",
```
Moyasee: These two keys no longer exist in `en`. They are not in any other locale either, and nothing in `src/` references them, so they were removed upstream while this PR was open. Safe to drop.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1637 @@
...
+    "cancel_download_description": "Είστε σίγουροι ότι θέλετε να ακυρώσετε αυτή τη λήψη; Όλα τα αρχεία που έχετε κατεβάσει θα διαγραφούν.",
+    "keep_downloading": "Όχι, συνέχισε τη λήψη",
+    "yes_cancel": "Ναι, ακύρωση",
+    "filter": "Φιλτράρισμα των παιχνιδιών που έχουν κατέβει",
+    "remove": "Κατάργηση",
+    "downloading_metadata": "Λήψη μεταδεδομένων…",
+    "deleting": "Διαγραφή του προγράμματος εγκατάστασης…",
+    "delete": "Απεγκατάσταση του προγράμματος εγκατάστασης",
+    "delete_modal_title": "Είσαι σίγουρος;",
+    "delete_modal_description": "Αυτό θα διαγράψει όλα τα αρχεία εγκατάστασης από τον υπολογιστή σας",
```
Moyasee: Clearest example of the register mix: the title is informal singular (Είσαι σίγουρος) while the description right under it is formal plural (υπολογιστή σας). Same dialog, two registers. `Είστε σίγουροι;` would match the description.

Same informal title shows up at `game_details.remove_from_library_title` (L325), `game_details.reset_achievements_title` (L546) and `user_profile.sign_out_modal_title`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1637 @@
...
+    "repack_count_other": "Προστέθηκαν {{count}} repacks",
+    "new_update_available": "Διατίθεται η έκδοση {{version}}",
+    "restart_to_install_update": "Επανεκκινήστε το Hydra για να εγκαταστήσετε την ενημέρωση",
+    "notification_achievement_unlocked_title": "Ξεκλείδωσε το επίτευγμα για το {{game}}",
+    "notification_achievement_unlocked_body": "Ξεκλειδώθηκαν το {{achievement}} και {{count}} άλλα",
+    "new_friend_request_description": "Ο/Η {{displayName}} σου έστειλε αίτημα φιλίας",
+    "new_friend_request_title": "Νέο αίτημα φιλίας",
+    "extraction_complete": "Η εξαγωγή ολοκληρώθηκε",
+    "game_extracted": "{{title}} εξήχθη με επιτυχία",
+    "friend_started_playing_game": "Το {{displayName}} άρχισε να παίζει ένα παιχνίδι",
```
Moyasee: `Το` is the neuter article, so it reads oddly in front of a person's name. `notifications.new_friend_request_description` already handles this correctly with `Ο/Η {{displayName}}`.

Same fix needed at `notifications_page.friend_request_received_description` (L1589), `friend_request_accepted_description` (L1591) and `review_answer_description` (L1597).

src/locales/el/translation.json
```diff
@@ -0,0 +1,1637 @@
...
+    "request_sent": "Το αίτημα εστάλη",
+    "request_received": "Λήφθηκε το αίτημα",
+    "accept_request": "Αποδοχή αιτήματος",
+    "ignore_request": "Αγνοήστε το αίτημα",
+    "cancel_request": "Ακύρωση αιτήματος",
+    "undo_friendship": "Ακύρωση φιλίας",
+    "friendship_removed": "Ο φίλος διαγράφηκε",
+    "request_accepted": "Το αίτημα έγινε δεκτό",
+    "user_blocked_successfully": "Ο χρήστης αποκλείστηκε με επιτυχία",
+    "user_block_modal_text": "Αυτό θα αποκλείσει το {{displayName}}",
```
Moyasee: Same article issue, accusative here. `τον/την {{displayName}}` instead of `το`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1637 @@
...
+    "context_menu_view_achievements": "Προβολή επιτευγμάτων",
+    "context_menu_share": "Κοινοποίηση",
+    "context_menu_options": "Επιλογές",
+    "context_menu_uninstall": "Απεγκατάσταση",
+    "context_menu_remove_from_library": "Διαγραφή από τη βιβλιοθήκη",
+    "context_menu_add_to_library": "Προσθήκη στη βιβλιοθήκη"
+  },
+  "achievement": {
+    "achievement_unlocked": "Το επίτευγμα ξεκλείδωσε",
+    "user_achievements": "Τα επιτεύγματα του {{displayName}}",
```
Moyasee: `του` assumes a masculine name. `του/της {{displayName}}` keeps it neutral and matches the `Ο/Η` pattern used over in `notifications`.

src/shared/language-flags.ts
```diff
@@ -30,6 +30,7 @@ const LANGUAGE_FLAG_MAP = {
   uz: "UZ",
   sv: "SE",
   lv: "LV",
+  el: "GR",
```
Moyasee: Minor: `el` is appended at the end here, but inserted alphabetically between `de` and `es` in `locales/index.ts`. Moving it up next to `de` would keep the two files consistent.

Moyasee (changes requested):
Thanks for sticking with this through the review rounds. The Greek itself reads well, and the mechanical checks all come back clean: interpolation placeholders match the English source exactly (0 mismatches), HTML tags match, all 38 plural variants are present, no empty values, and the Greek `;` is used correctly throughout instead of a Latin `?`.

Two things to sort out before this goes in.

**1. The file has drifted from `main` (30 keys missing)**

The description mentions 1,587 keys, but `main` is at 1,615 now. The two merges from `main` pulled in new keys that were never translated. They fall back to English at runtime so nothing breaks, but Greek users will see English strings in those spots.

`sidebar`: `edit_game_modal_grid`, `edit_game_modal_select_grid`, `edit_game_modal_grid_resolution`, `edit_game_modal_grid_preview`, `edit_game_modal_grid_drop_zone`, `edit_game_modal_grid_drop_zone_empty`, `edit_game_modal_drop_to_replace_grid`, `edit_game_modal_drop_grid_image_here`, `steamgriddb_section_title`, `steamgriddb_refresh`, `steamgriddb_use_default`, `steamgriddb_loading`, `steamgriddb_no_results`, `steamgriddb_fetch_failed`, `steamgriddb_sign_in_required`, `steamgriddb_stale_cache`, `steamgriddb_artwork_updated`, `steamgriddb_artwork_reset`, `classics_hero_layout_warning`, `classics_hero_layout_open_settings`, `custom_assets_not_sync`

`game_details`: `video`, `view_fullscreen`, `previous_media`, `next_media`, `reset_playtime`, `reset_playtime_title`, `reset_playtime_description`, `reset_playtime_success`, `reset_playtime_error`

**2. Formality register is mixed**

Most of the file uses the formal plural (σας, μπορείτε, επιλέξτε), but around 29 strings switch to the informal singular (σου, μπορείς, Είσαι σίγουρος), sometimes within the same dialog. I left a comment on the clearest case. Normalizing on the formal form makes sense since that is already what most of the file does.

The rest is inline and minor. Nothing structural: the registration in `index.ts`, `language-flags.ts`, `shared/index.ts` and the Big Picture date hook all look correct.

NickCh11:
@greptile

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+  },
+  "sidebar": {
+    "catalogue": "Κατάλογος",
+    "library": "Βιβλιοθήκη",
+    "downloads": "Λήψεις",
+    "settings": "Ρυθμίσεις",
+    "my_library": "Η βιβλιοθήκη μου",
+    "collections": "Συλλογές",
+    "games": "Παιχνίδια",
+    "big_picture": "Μεγάλη εικόνα",
```
Moyasee: `Big Picture` is a mode name, we keep it untranslated everywhere else (ru, es, fr, pt-BR all leave it verbatim, pl uses `Tryb Big Picture`).

Worth noting this PR is already inconsistent with itself here: the Big Picture locale keeps bare `Big Picture` at `exact.Big Picture` and inside `settings_startup_section_description`, so right now the mode has two Greek names. Same fix needed on L1289.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "open_save_folder": "Άνοιγμα φακέλου αποθήκευσης",
+    "searching_save_folder": "Αναζήτηση φακέλου αποθήκευσης...",
+    "no_save_folder_found": "Δεν βρέθηκε φάκελος αποθήκευσης",
+    "open_download_location": "Δείτε τα αρχεία που έχετε κατεβάσει",
+    "create_shortcut": "Δημιουργία συντόμευσης στην επιφάνεια εργασίας",
+    "create_shortcut_simple": "Δημιουργία συντόμευσης",
+    "clear": "Διαγραφή",
+    "remove_files": "Διαγραφή αρχείων",
+    "remove_from_library_title": "Είστε σίγουροι;",
+    "remove_from_library_description": "Αυτό θα αφαιρέσει το {{game}} από τη βιβλιοθήκη σου",
```
Moyasee: This is the register mix from the main comment. `βιβλιοθήκη σου` here (informal) but `reset_achievements_description` in the same modal family uses `επιτεύγματά σας` (formal). Whichever way you go, these two should match.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "rpcs3_already_running_confirm": "Κλείσιμο και εκκίνηση",
+    "shortcuts_section_title": "Συντομεύσεις",
+    "shortcuts_section_description": "Δημιουργήστε συντομεύσεις για να ξεκινήσετε γρήγορα αυτό το παιχνίδι",
+    "downloads_section_title": "Λήψεις",
+    "downloads_section_description": "Δείτε τις ενημερώσεις ή άλλες εκδόσεις αυτού του παιχνιδιού",
+    "danger_zone_section_title": "Ζώνη κινδύνου",
+    "danger_zone_section_description": "Διαγράψτε αυτό το παιχνίδι από τη βιβλιοθήκη σας ή από τα αρχεία που κατεβάσατε μέσω του Hydra",
+    "download_in_progress": "Η λήψη βρίσκεται σε εξέλιξη",
+    "download_paused": "Η λήψη έχει τεθεί σε παύση",
+    "extracting": "Εξαγωγή",
```
Moyasee: `Εξαγωγή` reads as "export" more than "extract". Your Big Picture file already has the better version for the same English source: `Αποσυμπίεση`. Same applies to `automatically_extract_downloaded_files` below.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "cancel": "Ακύρωση",
+    "cancel_download": "Ακύρωση λήψης;",
+    "cancel_download_description": "Είστε σίγουροι ότι θέλετε να ακυρώσετε αυτή τη λήψη; Όλα τα αρχεία που έχετε κατεβάσει θα διαγραφούν.",
+    "keep_downloading": "Όχι, συνέχισε τη λήψη",
+    "yes_cancel": "Ναι, ακύρωση",
+    "filter": "Φιλτράρισμα των παιχνιδιών που έχουν κατέβει",
+    "remove": "Κατάργηση",
+    "downloading_metadata": "Λήψη μεταδεδομένων…",
+    "deleting": "Διαγραφή του προγράμματος εγκατάστασης…",
+    "delete": "Απεγκατάσταση του προγράμματος εγκατάστασης",
```
Moyasee: `Απεγκατάσταση` means uninstall, but this key doesn't uninstall anything. It's the trash-icon item in `download-group.tsx:849` and the confirm button in `delete-game-modal.tsx:36`, and it deletes the downloaded install files.

Your own `delete_modal_description` two lines down gets it right with `θα διαγράψει όλα τα αρχεία εγκατάστασης`. Something like `Διαγραφή αρχείων εγκατάστασης` would match and also fixes the 43 char label in a compact menu.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "subscription_active_until": "Το Hydra Cloud σας είναι ενεργό μέχρι τις {{date}}",
+    "manage_subscription": "Διαχείριση συνδρομής",
+    "update_email": "Ενημέρωση μέσω email",
+    "update_password": "Ενημέρωση κωδικού πρόσβασης",
+    "current_email": "Τρέχουσα διεύθυνση ηλεκτρονικού ταχυδρομείου:",
+    "no_email_account": "Δεν έχετε ορίσει ακόμα διεύθυνση ηλεκτρονικού ταχυδρομείου",
+    "account_data_updated_successfully": "Τα δεδομένα του λογαριασμού ενημερώθηκαν με επιτυχία",
+    "renew_subscription": "Ανανέωση του Hydra Cloud",
+    "subscription_expired_at": "Η συνδρομή σας έληξε στις {{date}}",
+    "no_subscription": "Απολαύστε την Ύδρα με τον καλύτερο δυνατό τρόπο",
```
Moyasee, line 928: `Ύδρα` translates the brand name. Every other locale keeps `Hydra` as-is (ru `Наслаждайтесь Hydra`, de `Genieße Hydra`, pt-BR `Aproveite o Hydra`). This is also the only string in the PR that transliterates it.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "detect_memory_cards": "Ανίχνευση",
+    "redetect_memory_cards": "Επανέλεγχος",
+    "pick_memory_card_file": "Επιλογή κάρτας μνήμης",
+    "pick_memory_card_file_ps1": "Επιλογή κάρτας μνήμης",
+    "no_memory_cards": "Δεν έχουν εντοπιστεί ακόμη αποθηκευμένα αρχεία σε κάρτα μνήμης. Πραγματοποιήστε σάρωση για να βρείτε τα αποθηκευμένα αρχεία του PCSX2.",
+    "no_memory_cards_ps1": "Δεν έχουν εντοπιστεί ακόμη αποθηκευμένα αρχεία σε κάρτα μνήμης. Πραγματοποιήστε σάρωση για να βρείτε τα αποθηκευμένα αρχεία σας στο DuckStation.",
+    "memcard_scan_title": "Σάρωση των καρτών μνήμης του PS2 σας",
+    "memcard_scan_intro": "Ανάγνωση των αποθηκευμένων στοιχείων και αντιστοίχισή τους με τα εξώφυλλα του LaunchBox.",
+    "memcard_scanning": "Κάρτες ανάγνωσης",
+    "memcard_matching": "Αντιστοίχιση αποθηκευμένων δεδομένων",
```
Moyasee: `αποθηκευμένων δεδομένων` here, but `αποθηκευμένων στοιχείων` in `memcard_scan_intro`, `αποθηκευμένα αρχεία` in `memory_cards_section_description`, and `αποθηκεύσεις` in `memcard_stat_saves`. All four are "saves" in the same settings section.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "memcard_group_count_other": "{{count}} αποθηκεύσεις",
+    "expand_memory_card": "Επέκταση κάρτας μνήμης",
+    "collapse_memory_card": "Σύμπτυξη κάρτας μνήμης",
+    "remove_memory_card": "Αφαίρεση κάρτας μνήμης",
+    "memcard_search_placeholder": "Αναζήτηση αποθηκεύσεων",
+    "memcard_no_results": "Δεν υπάρχουν αποθηκεύσεις που να ταιριάζουν με την αναζήτησή σας.",
+    "pagination_previous": "Προηγούμενη σελίδα",
+    "pagination_next": "Επόμενη σελίδα",
+    "pagination_page_of": "{{page}} / {{total}}",
+    "forget_memcard_save_title": "Να αγνοήσω αυτή την αποθήκευση;",
```
Moyasee: Two things: `αγνοήσω` is "ignore" rather than "forget", and the first person singular puts the app in the "I" voice, which nothing else in either file does. Maybe `Να καταργηθεί αυτή η αποθήκευση;` or something else impersonal.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "game_cloud_group_title": "Δημιουργία αντιγράφων ασφαλείας στο cloud",
+    "game_no_saves_title": "Δεν βρέθηκαν αποθηκευμένα αρχεία",
+    "game_no_saves_found": "Δεν βρήκαμε καμία αποθήκευση για αυτό το παιχνίδι στις κάρτες μνήμης σας.",
+    "game_no_memcards_title": "Δεν έχουν προστεθεί κάρτες μνήμης",
+    "game_no_memcards_description": "Δεν έχετε προσθέσει ακόμα καμία κάρτα μνήμης. Προσθέστε μία για να εντοπίσετε τα αποθηκευμένα αρχεία αυτού του παιχνιδιού και να τα δημιουργήσετε αντίγραφα ασφαλείας στο cloud.",
+    "game_add_memory_card": "Προσθέστε μια κάρτα μνήμης",
+    "cloud_subscription_required": "Απαιτείται συνδρομή στο Hydra Cloud για τη δημιουργία αντιγράφων ασφαλείας και την επαναφορά των αποθηκευμένων δεδομένων του εξομοιωτή.",
+    "no_cloud_saves": "Δεν υπάρχουν ακόμη αποθηκευμένα αρχεία στο cloud. Δημιουργήστε αντίγραφο ασφαλείας ενός αρχείου αποθήκευσης από την ενότητα «Κάρτες μνήμης» παραπάνω.",
+    "cloud_refresh": "Ανανέωση",
+    "cloud_backup": "Δημιουργία αντιγράφου ασφαλείας",
```
Moyasee: This one's the tightest fit of the length cases. `Back up` is 7 chars, this is 31, and it's a `<Button>` label in `game-emulation-saves.tsx:261` plus a dropdown item in `memory-cards-section.tsx:537`. Might be worth just `Backup` or a shorter form here even if the descriptions keep `αντίγραφο ασφαλείας`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+    "remove_emulator_description": "Το Hydra θα διαγράψει τη διαδρομή {{name}} και όλους τους αντιστοιχισμένους φακέλους ROM. Ο ίδιος ο εξομοιωτής και τα αρχεία ROM σας παραμένουν στον δίσκο. Μπορείτε να εκτελέσετε ξανά τη ρύθμιση αργότερα.",
+    "big_picture": "Μεγάλη εικόνα",
+    "big_picture_startup": "Εκκίνηση",
+    "big_picture_audio": "Ήχος",
+    "big_picture_input": "Εισαγωγή",
+    "big_picture_diagnostics": "Διαγνωστικά",
+    "big_picture_enable_sounds": "Ενεργοποίηση ήχων Big Picture",
+    "big_picture_enable_virtual_keyboard": "Ενεργοποίηση του εικονικού πληκτρολογίου «Big Picture»",
+    "big_picture_enable_diagnostics": "Ενεργοποίηση διαγνωστικών πλοήγησης",
+    "big_picture_diagnostics_position": "Θέση στον τομέα της διαγνωστικής"
```
Moyasee: `Θέση στον τομέα της διαγνωστικής` reads as "position in the field of diagnostics", which isn't what this control does. The correct version is already in your Big Picture file: `Θέση διαγνωστικών`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1665 @@
...
+  },
+  "forms": {
+    "toggle_password_visibility": "Εναλλαγή ορατότητας κωδικού πρόσβασης"
+  },
+  "user_profile": {
+    "amount_hours": "{{amount}} ώρες",
+    "amount_minutes": "{{amount}} λεπτά",
+    "amount_hours_short": "{{amount}} ώρ.",
+    "amount_minutes_short": "{{amount}} λεπ.",
+    "last_time_played": "Τελευταία συμμετοχή {{period}}",
```
Moyasee: `Τελευταία συμμετοχή` means "last participation". `game_details.last_time_played` has the identical English source and gets it right with `Παίχτηκε τελευταία {{period}}`, so that one can just be reused here.

Moyasee (commented):
Nice work, this is one of the cleaner locale additions we've had. I checked it programmatically against `en` and the structure is solid:

- all 1615 desktop keys and all 796 Big Picture strings present
- zero interpolation placeholder mismatches across all 2411 strings
- plural suffix parity exact (38 desktop, 14 Big Picture `_one`/`_other`)
- section order and per-section key order match `en`, so future diffs stay readable
- no extra keys, no empty values, no HTML tag drift
- `tsc --noEmit -p tsconfig.web.json` clean, prettier clean, all wiring points hit (`BigPictureLanguage` union, both records, both `getDateLocale` helpers, `LANGUAGE_FLAG_MAP`)

Also good to see the Greek `;` for question marks and the decimal comma, those get missed a lot.

A few things to fix before this goes in.

## 8 desktop keys are missing

The branch merged `main` at d9e21d8 but `src/locales/el/translation.json` wasn't re-synced after, so the keys from #2585 didn't make it:

```
game_details.download_error_torrent_invalid_trackers
settings.global_trackers
settings.global_trackers_description
settings.global_trackers_append_manual
settings.global_trackers_append_url
settings.global_trackers_manual_placeholder
settings.global_trackers_url_placeholder
settings.global_trackers_save_error
```

Nothing in CI catches this (`text-integrity` only checks line endings and the LICENSE hash), and since `fallbackLng` is `en` the global trackers settings section will just quietly render in English. This is the one I'd want fixed before merge.

## Formal vs informal register is mixed

Counting across the whole file: 35 strings use the familiar singular (σου / εσύ / έχεις / μπορείς) against 158 in the polite plural (σας / εσείς / έχετε / επιλέξτε). Big Picture is 6 against 54. The two registers end up next to each other in the same modals, which is where it shows. Most of the file is already formal so σας is the cheaper direction, but either is fine as long as it's consistent. Left a note on one example inline.

## Terminology drift for "saves"

The memory cards settings section alone renders "saves" four ways: `αποθηκευμένα αρχεία`, `αποθηκευμένα στοιχεία`, `αποθηκευμένα δεδομένα`, `αποθηκεύσεις`. Worth picking one.

## String length in tight controls

Greek runs longer and that's expected, but a handful land in buttons and context menus where it'll overflow:

| key | en | el |
| --- | --- | --- |
| `settings.cloud_backup` | 7 | 31 |
| `downloads.delete` | 16 | 43 |
| `settings.current_email` | 14 | 45 |
| `game_details.backup_unfrozen` | 15 | 50 |
| `settings.app_basics` | 10 | 36 |

`αντίγραφο ασφαλείας` is the formally correct term, but Greek software UIs usually just keep "Backup". Could you post screenshots of the settings and downloads pages in Greek? That'd settle whether these actually break the layout or if I'm being paranoid.

Rest of the notes are inline. Leaving this as a comment rather than requesting changes since none of it is structural.

---

Unrelated to this PR, but the Big Picture `exact` dictionary has some string-extraction junk in the `en` source (CSS class names like `typography typography--h1`, selectors like `code, pre`, a stray `", using the same surface and expand animation as the migrated component."`, and a `"Vibration, ?"`). You handled the code-like ones correctly by leaving them verbatim. I'll open something separate to clean up the source. Same for the locale parity check, that gap is on us not you.

src/big-picture/src/locales/el/translation.json
```diff
@@ -0,0 +1,802 @@
...
+    "Enable Premiumize": "Ενεργοποίηση του Premiumize",
+    "Enable Real-Debrid": "Ενεργοποίηση του Real-Debrid",
+    "Enable repack update notifications": "Ενεργοποίηση ειδοποιήσεων για ενημερώσεις επανασυσκευασμένων αρχείων",
+    "Enable search for Steam achievements": "Ενεργοποίηση αναζήτησης επιτευγμάτων Steam",
+    "Enable TorBox": "Ενεργοποίηση του TorBox",
+    "Enable diagnostics": "Ενεργοποίηση διαγνωστικών λειτουργιών",
+    "Enable virtual keyboard": "Ενεργοποίηση εικονικού πληκτρολογίου",
+    "Enjoy Hydra in the best possible way": "Απολαύστε το Hydra με τον καλύτερο δυνατό τρόπο",
+    "Error": "Σφάλμα",
+    "Exit Big Picture": "Έξοδος από τη «Μεγάλη Εικόνα»",
```
Moyasee: This translates the mode name while L162 in this same file keeps it as `Big Picture`. Suggest `Έξοδος από το Big Picture` so it lines up with the rest of the file and with the desktop fix.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1677 @@
...
+    "remove_rom_folder_title": "Διαγραφή του φακέλου ROM",
+    "remove_rom_folder_description": "Το {{path}} δεν θα ελέγχεται πλέον. Ο φάκελος και τα αρχεία του παραμένουν στον δίσκο σας.",
+    "cancel_remove": "Ακύρωση",
+    "stat_games": "ΠΑΙΧΝΙΔΙΑ",
+    "stat_storage": "ΑΠΟΘΗΚΕΥΣΗ",
+    "stat_last_scan": "ΤΕΛΕΥΤΑΙΑ ΣΑΡΩΣΗ",
+    "stat_last_scan_caption": "Όλα τα αρχεία έχουν καταγραφεί",
+    "stat_games_caption": "{{system}} τίτλοι στη βιβλιοθήκη σας",
+    "stat_storage_caption_zero": "Συνολικά 0 αρχεία ROM",
+    "stat_storage_caption_one": "Σε {{count}} αρχεία ROM στους {{folders}} φακέλους",
```
Moyasee: This renders a duplicated noun. `{{folders}}` is not a number, it is the already-rendered `folder_count_one`/`folder_count_other` string, interpolated at `src/renderer/src/pages/settings/emulation/emulator-detail.tsx:595`. Appending `φακέλους` after it repeats the word. The singular also uses the plural noun `αρχεία`.

```
files=1 folders=1
  EN: Across 1 ROM file in 1 folder
  EL: Σε 1 αρχεία ROM στους 1 φάκελος φακέλους

files=1 folders=3
  EN: Across 1 ROM file in 3 folders
  EL: Σε 1 αρχεία ROM στους 3 φάκελοι φακέλους
```

`stat_storage_caption_other` already handles it correctly with `που βρίσκονται σε {{folders}}`. Suggest matching it:

```suggestion
    "stat_storage_caption_one": "Σε {{count}} αρχείο ROM που βρίσκεται σε {{folders}}",
```

src/locales/el/translation.json
```diff
@@ -0,0 +1,1677 @@
...
+    "memcard_stat_saves": "ΑΠΟΘΗΚΕΥΣΕΙΣ",
+    "memcard_export": "Εξαγωγή .psu",
+    "memcard_export_mcs": "Εξαγωγή .mcs",
+    "memcard_exporting": "Εξαγωγή…",
+    "memcard_export_success": "Αποθηκεύτηκε στο {{path}}",
+    "memcard_export_failed": "Δεν ήταν δυνατή η εξαγωγή αυτής της αποθήκευσης",
+    "memcard_files_count_one": "{{count}} αρχείο",
+    "memcard_files_count_other": "{{count}} αρχεία",
+    "memcard_blocks_count_one": "{{count}} μπλοκ",
+    "memcard_blocks_count_other": "{{count}} τεμάχια",
```
Moyasee: Singular is `μπλοκ`, plural is `τεμάχια` ("pieces"). Same noun should be the same word in both forms, and `μπλοκ` is indeclinable so it does not change.

```suggestion
    "memcard_blocks_count_other": "{{count}} μπλοκ",
```

Moyasee (commented):
Checked this out at e1f1e5c and ran the mechanics: `yarn typecheck` clean, `eslint` clean on the changed lines, `prettier --check` clean on all eight files, no CR characters so `text-integrity` passes. All six registration touchpoints are covered (`src/locales/index.ts`, `language-flags.ts`, `src/shared/index.ts` date-fns, the Big Picture `locales/index.ts` and `BigPictureLanguage` union, `resolveBigPictureLanguage`, and `use-date.hook.ts`), and `language_name` is set, so the dropdown renders.

Diffed both locales against their English sources: zero interpolation placeholder mismatches, zero Trans/HTML tag mismatches, no leading or trailing whitespace drift, no newline-count drift. Date format strings and the CSS class names that live in the Big Picture `exact` map were correctly left byte-identical. Big Picture is 796/796 with all seven plural sets correct. The prose reads like real Greek, not machine output.

Four things need fixing in the desktop file, all left inline. One of them renders visibly broken text.

On the missing keys: `el` is short six `game_details` keys (`download_error_hoster_unlock_*`, the two `download_error_vikingfile_*`, and `downloader_requires_hydra_cloud`). Every locale except pt, ru and es is missing the same six, and i18next falls back to `en`, so that is consistent with the repo and not something to hold this up for.

Two things outside the diff: the counts in the PR description are stale (en has 1632 leaf keys and this file has 1627, not 1615), and the PR still has no label, which the template requires before merge. `translation` is the one you want.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1677 @@
...
+    "no_genres": "Δεν υπάρχουν κατηγορίες",
+    "mode_modern_games": "Σύγχρονα παιχνίδια",
+    "mode_classics": "Κλασικά",
+    "platforms": "Πλατφόρμες"
+  },
+  "game_details": {
+    "bios_not_configured_toast": "Δεν εντοπίστηκε το BIOS. Προσθέστε τον φάκελο «BIOS» στις ρυθμίσεις εξομοίωσης.",
+    "open_download_options": "Εμφάνιση επιλογών λήψης",
+    "download_options_zero": "Δεν υπάρχει επιλογή λήψης",
+    "download_options_one": "Επιλογή λήψης {{count}}",
```
Moyasee: `{{count}}` after the noun reads as an identifier rather than a quantity, so this renders as "Download option #1". Four other singular forms have the same problem, and in every case the matching `_other` form gets the order right:

| line | key | current | reads as |
|---|---|---|---|
| 243 | `download_options_one` | `Επιλογή λήψης {{count}}` | "Download option #1" |
| 886 | `found_download_option_one` | `Βρέθηκε η επιλογή λήψης {{countFormatted}}` | "Found download option #1" |
| 1263 | `setup_rom_yml_registered_one` | `Το παιχνίδι {{count}} έχει ήδη καταχωριστεί…` | "Game #1 is already registered" |
| 1273 | `setup_unmatched_title_one` | `Δεν βρέθηκε αντίστοιχο αρχείο {{count}}` | "No matching file #1 found" |
| 1281 | `setup_done_body_one` | `Προσθέσαμε το παιχνίδι {{count}}…` | "We added game #1" |

For this one:

```suggestion
    "download_options_one": "{{count}} επιλογή λήψης",
```

src/locales/el/translation.json
```diff
@@ -0,0 +1,1677 @@
...
+    "setup_rom_yml_registered_other": "{{count}} παιχνίδια έχουν ήδη καταχωρηθεί στο RPCS3 και θα αντιστοιχιστούν αυτόματα.",
+    "setup_scan_title": "Δημιουργία της βιβλιοθήκης {{system}}",
+    "setup_scan_intro": "Αντιστοίχιση των αρχείων σας με τον κατάλογο του LaunchBox. Αυτό συνήθως διαρκεί λιγότερο από ένα λεπτό.",
+    "setup_scanning": "Σάρωση αρχείων…",
+    "setup_matching": "Παιχνίδια αντιστοίχισης…",
+    "setup_scan_complete": "Η σάρωση ολοκληρώθηκε",
+    "setup_match_matched": "αντιστοιχισμένα",
+    "setup_match_unmatched": "χωρίς αντιστοίχιση",
+    "setup_match_wrong_platform": "λανθασμένη πλατφόρμα",
+    "setup_unmatched_title_one": "Δεν βρέθηκε αντίστοιχο αρχείο {{count}}",
```
Moyasee: Besides the `{{count}}` placement, this changes what the string says. English is "{{count}} file couldn't be matched", meaning a file exists and matching failed. This says no matching file was found, which is a different outcome. `setup_unmatched_title_other` keeps the original meaning.

```suggestion
    "setup_unmatched_title_one": "Δεν ήταν δυνατή η αντιστοίχιση {{count}} αρχείου",
```

src/locales/el/translation.json
```diff
@@ -0,0 +1,1677 @@
...
+    "download_error_not_cached_on_premiumize": "Αυτή η λήψη δεν είναι διαθέσιμη στο Premiumize.",
+    "download_error_premiumize_transfer_started": "Η μεταφορά ξεκίνησε στο Premiumize. Παρακαλώ, δοκιμάστε ξανά σε λίγο.",
+    "download_error_not_cached_on_alldebrid": "Αυτή η λήψη δεν είναι διαθέσιμη στο AllDebrid.",
+    "download_error_premiumize_rate_limit_exceeded": "Υπερβαίθηκε το όριο συχνότητας Premiumize. Παρακαλώ περιμένετε και δοκιμάστε ξανά.",
+    "download_error_alldebrid_rate_limit_exceeded": "Υπερβαίθηκε το όριο χρήσης του AllDebrid. Παρακαλώ περιμένετε και δοκιμάστε ξανά.",
+    "download_error_premiumize_unavailable": "Η υπηρεσία Premiumize δεν είναι προσωρινά διαθέσιμη. Παρακαλώ δοκιμάστε ξανά αργότερα.",
+    "download_error_alldebrid_unavailable": "Το AllDebrid δεν είναι προσωρινά διαθέσιμο. Παρακαλώ δοκιμάστε ξανά αργότερα.",
+    "download_error_vikingfile_nimbus_quota_exceeded": "Το όριο του Nimbus για το VikingFile έχει εξαντληθεί για σήμερα. Παρακαλώ προσπαθήστε ξανά αύριο.",
```
Moyasee: This key no longer exists. It was renamed to `download_error_vikingfile_quota_exceeded` in 8920e6e, and `src/shared/constants.ts:78` points at the new name. No other locale carries the old one, so it can be dropped.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1677 @@
...
+    "context_menu_mark_as_favorite": "Προσθήκη στα Αγαπημένα",
+    "context_menu_view_achievements": "Προβολή επιτευγμάτων",
+    "context_menu_share": "Κοινοποίηση",
+    "context_menu_options": "Επιλογές",
+    "context_menu_uninstall": "Απεγκατάσταση",
+    "context_menu_remove_from_library": "Διαγραφή από τη βιβλιοθήκη",
+    "context_menu_add_to_library": "Προσθήκη στη βιβλιοθήκη"
+  },
+  "achievement": {
+    "achievement_unlocked": "Το επίτευγμα ξεκλείδωσε",
```
Moyasee: Nit: active voice, so it reads as the achievement having unlocked something. Line 1317 in this same file already uses the passive `Ξεκλειδώθηκαν` for the same event.

```suggestion
    "achievement_unlocked": "Το επίτευγμα ξεκλειδώθηκε",
```

src/locales/el/translation.json
```diff
@@ -0,0 +1,1677 @@
...
+    "memcard_stat_matched": "ΤΑΙΡΙΑΖΟΥΝ",
+    "memcard_stat_saves": "ΑΠΟΘΗΚΕΥΣΕΙΣ",
+    "memcard_export": "Εξαγωγή .psu",
+    "memcard_export_mcs": "Εξαγωγή .mcs",
+    "memcard_exporting": "Εξαγωγή…",
+    "memcard_export_success": "Αποθηκεύτηκε στο {{path}}",
+    "memcard_export_failed": "Δεν ήταν δυνατή η εξαγωγή αυτής της αποθήκευσης",
+    "memcard_files_count_one": "{{count}} αρχείο",
+    "memcard_files_count_other": "{{count}} αρχεία",
+    "memcard_blocks_count_one": "{{count}} μπλοκ",
```
Moyasee, line 1139: Nit while you are in here: `game_details.game_added_to_pinned` uses `«καρφιτσωμένα»` and `user_profile.game_added_to_pinned` uses `«Καρφιτσωμένα»` for the same English string, and `setup_rom_yml_registered` uses `καταχωριστεί` in the singular but `καταχωρηθεί` in the plural. Worth picking one of each.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
+{
+  "language_name": "Ελληνικά",
+  "app": {
+    "successfully_signed_in": "Η σύνδεση ολοκληρώθηκε με επιτυχία",
+    "error_boundary_title": "Παρουσιάστηκε κάποιο πρόβλημα",
+    "error_boundary_description": "Η εφαρμογή αντιμετώπισε ένα απροσδόκητο σφάλμα. Μπορείτε να ανανεώσετε τη σελίδα για να συνεχίσετε να χρησιμοποιείτε το Hydra.",
+    "error_boundary_copy_error": "Σφάλμα αντιγραφής",
```
Moyasee: `Copy error` is the label on the button that copies the stack trace ([error-fallback.tsx:99](https://github.com/hydralauncher/hydra/blob/main/src/renderer/src/components/error-boundary/error-fallback.tsx#L99)), not a description of a failure. `Σφάλμα αντιγραφής` reverses it into "copy failure".

`Αντιγραφή σφάλματος`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "tags": "Ετικέτες",
+    "publishers": "Εκδότες",
+    "protondb": "ProtonDB",
+    "download_sources": "Πηγές λήψης",
+    "compatibility_requirements": "Συμβατότητα με το Linux",
+    "protondb_minimum": "Ελάχιστο επίπεδο ProtonDB",
+    "steam_deck_minimum": "Ελάχιστες απαιτήσεις για το Steam Deck",
+    "steam_deck_compatible": "Συμβατό με το Steam Deck",
+    "compatibility_any": "Οποιοδήποτε",
+    "protondb_silver_plus": "Silver+",
```
Moyasee: ProtonDB and Deck tiers are half translated. `Silver+` here, `Gold+` on L213 and `deck_playable_plus` on L215 stay English, while `platinum_only` on L214 and `deck_verified_only` on L216 are Greek. `protondb_tier_silver` on L271 is `Ασήμι`, so the catalogue filter and the badge on the game page use different words for the same tier. `es`, `pt-BR` and `ru` translate both sides.

Separate nit on the tier names themselves: `Χάλκινο` is a neuter adjective, `Ασήμι` a neuter noun, `Χρυσός` a masculine noun and `Πλατίνα` a feminine noun. They render as one list.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "eta": "Ολοκλήρωση {{eta}}",
+    "calculating_eta": "Υπολογισμός του εναπομείναντος χρόνου…",
+    "downloading_metadata": "Λήψη μεταδεδομένων…",
+    "filter": "Φιλτράρισμα repacks",
+    "requirements": "Απαιτήσεις συστήματος",
+    "minimum": "Ελάχιστο",
+    "recommended": "Συνιστάται",
+    "paused": "Σε παύση",
+    "release_date": "Κυκλοφόρησε στις {{date}}",
+    "publisher": "Εκδόθηκε από τον {{publisher}}",
```
Moyasee: `τον` hardcodes the masculine article in front of a company name, so this renders as `Εκδόθηκε από τον Ubisoft`. Publisher names are not reliably masculine.

Dropping the article works for all of them: `Εκδόθηκε από {{publisher}}`. `ru` avoids the same trap with `Издатель {{publisher}}`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "download_path": "Διαδρομή λήψης",
+    "previous_screenshot": "Προηγούμενο στιγμιότυπο οθόνης",
+    "next_screenshot": "Επόμενο στιγμιότυπο οθόνης",
+    "screenshot": "Στιγμιότυπο οθόνης {{number}}",
+    "video": "Βίντεο {{number}}",
+    "open_screenshot": "Άνοιγμα στιγμιότυπου οθόνης {{number}}",
+    "view_fullscreen": "Προβολή σε πλήρη οθόνη",
+    "previous_media": "Προηγούμενο",
+    "next_media": "Επόμενο",
+    "download_settings": "Λήψη ρυθμίσεων",
```
Moyasee: Same reversal as the copy button. This is the title of the download configuration modal ([download-settings-modal.tsx:1270](https://github.com/hydralauncher/hydra/blob/main/src/renderer/src/pages/game-details/modals/download-settings-modal.tsx#L1270)), but `Λήψη ρυθμίσεων` means "downloading the settings".

`Ρυθμίσεις λήψης`. `download_options_title` a few lines up already has the order right with `Επιλογές λήψης`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "torrent_files_disclaimer": "Προηγμένη επιλογή: αν δεν είστε σίγουροι για τη λειτουργία της, διατηρήστε την προεπιλεγμένη επιλογή.",
+    "select_files_to_download": "Επιλέξτε τα αρχεία που θέλετε να κατεβάσετε",
+    "loading_torrent_files": "Φόρτωση αρχείων torrent…",
+    "search_torrent_files": "Αναζήτηση αρχείων",
+    "selected_files": "Επιλεγμένα αρχεία",
+    "selected_size": "Επιλεγμένο μέγεθος",
+    "select_all_files": "Επιλογή όλων",
+    "clear_file_selection": "Διαγραφή",
+    "retry_fetch_torrent_files": "Επανάληψη",
+    "downloader_online": "Online",
```
Moyasee: `Online` left in English while `downloader_offline` three lines below is fully Greek.

Same split at `downloads.peak` on L781: it stays `PEAK` while `network` and `files` on either side of it are `ΔΙΚΤΥΟ` and `ΑΡΧΕΙΑ`, and all three render in the same stat row. `fr`, `es` and `ru` translate both keys.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "download_halted_title": "Η λήψη διακόπηκε",
+    "download_halted_description": "Η λήψη του {{title}} διακόπηκε επειδή ο φάκελος λήψεων δεν έχει άλλο διαθέσιμο χώρο. Ελευθερώστε χώρο και συνεχίστε τη λήψη.",
+    "delete_archive_title": "Θέλετε να διαγράψετε το {{fileName}};",
+    "delete_archive_description": "Το αρχείο αποσυμπιέστηκε με επιτυχία και δεν είναι πλέον απαραίτητο.",
+    "yes": "Ναι",
+    "no": "Όχι",
+    "network": "ΔΙΚΤΥΟ",
+    "peak": "PEAK",
+    "files": "ΑΡΧΕΙΑ",
+    "move_up": "Μετάβαση προς τα πάνω",
```
Moyasee: Matched pair of queue reorder buttons using two different verbs. `Μετάβαση` is "navigate to", `Μετακίνηση` on the next line is "move".

`Μετακίνηση προς τα πάνω`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "enable_achievement_custom_notifications": "Ενεργοποίηση προσαρμοσμένων ειδοποιήσεων για τα επιτεύγματα",
+    "alignment": "Ευθυγράμμιση",
+    "variation": "Παραλλαγή",
+    "default": "Προεπιλογή",
+    "rare": "Σπάνιο",
+    "platinum": "Πλατίνα",
+    "hidden": "Κρυφό",
+    "test_notification": "Ειδοποίηση δοκιμής",
+    "achievement_sound_volume": "Ένταση ήχου επιτεύγματος",
+    "select_achievement_sound": "Επιλογή ήχου επιτυχίας",
```
Moyasee: `επιτυχίας` means "success". Achievement is `επίτευγμα` everywhere else in the file.

`Επιλογή ήχου επιτεύγματος`, and the same on L1056.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "game_cloud_group_title": "Δημιουργία αντιγράφων ασφαλείας στο cloud",
+    "game_no_saves_title": "Δεν βρέθηκαν αποθηκεύσεις",
+    "game_no_saves_found": "Δεν εντοπίστηκαν αποθηκεύσεις για αυτό το παιχνίδι στις κάρτες μνήμης σας.",
+    "game_no_memcards_title": "Δεν έχουν προστεθεί κάρτες μνήμης",
+    "game_no_memcards_description": "Δεν έχετε προσθέσει ακόμα καμία κάρτα μνήμης. Προσθέστε μία για να εντοπίσετε τις αποθηκεύσεις αυτού του παιχνιδιού και να δημιουργήσετε αντίγραφα ασφαλείας στο cloud.",
+    "game_add_memory_card": "Προσθέστε μια κάρτα μνήμης",
+    "cloud_subscription_required": "Απαιτείται συνδρομή στο Hydra Cloud για τη δημιουργία αντιγράφων ασφαλείας και την επαναφορά αποθηκεύσεων εξομοιωτή.",
+    "no_cloud_saves": "Δεν υπάρχουν ακόμη αποθηκεύσεις στο cloud. Δημιουργήστε αντίγραφο ασφαλείας μιας αποθήκευσης από την ενότητα «Κάρτες μνήμης» παραπάνω.",
+    "cloud_refresh": "Ανανέωση",
+    "cloud_backup": "Backup",
```
Moyasee: This is a button label paired with `cloud_backing_up` ([memory-cards-section.tsx:537](https://github.com/hydralauncher/hydra/blob/main/src/renderer/src/pages/settings/emulation/memory-cards-section.tsx#L537)), and every sibling key around it is proper Greek:

```
cloud_backing_up  "Δημιουργία αντιγράφου ασφαλείας…"
cloud_backup_all  "Δημιουργήστε αντίγραφα ασφαλείας για όλα"
cloud_backup      "Backup"
```

It is also a noun where `en` has the verb `Back up`. This is my fault for flagging the length last round without suggesting a replacement. `Αντίγραφο` is short enough for the button and stays Greek.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "cloud_backup_all": "Δημιουργήστε αντίγραφα ασφαλείας για όλα",
+    "cloud_backing_up": "Δημιουργία αντιγράφου ασφαλείας…",
+    "cloud_backup_success": "Η αποθήκευση δημιουργήθηκε ως αντίγραφο ασφαλείας στο Hydra Cloud",
+    "cloud_backup_failed": "Δεν ήταν δυνατή η δημιουργία αντιγράφου ασφαλείας για αυτή την αποθήκευση",
+    "cloud_backup_all_done": "Δημιουργήθηκαν αντίγραφα ασφαλείας για {{uploaded}} από {{total}} αποθηκεύσεις",
+    "cloud_restore": "Επαναφορά",
+    "cloud_restoring": "Επαναφορά…",
+    "cloud_restore_title": "Επαναφορά αποθηκευμένων δεδομένων σε κάρτα μνήμης",
```
Moyasee: The cloud restore flow uses four different words for "save" across consecutive strings: `αποθηκευμένα δεδομένα` here, `αρχείο αποθήκευσης` on L1179, `αποθηκευμένου αρχείου` on L1184, `αποθηκευμένο παιχνίδι` on L1185.

The rest of the memory cards section settled on `αποθήκευση`, which is the one to standardise on.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "cloud_restore_description": "Επιλέξτε μια κάρτα μνήμης για να αποθηκεύσετε αυτό το αρχείο αποθήκευσης. Η κάρτα θα τροποποιηθεί — θα δημιουργηθεί αυτόματα ένα αντίγραφο ασφαλείας.",
+    "cloud_restore_confirm": "Επαναφορά σε αυτή την κάρτα",
+    "cloud_restore_pick_file": "Επιλέξτε ένα άλλο αρχείο…",
+    "cloud_restore_no_cards": "Δεν εντοπίστηκαν κάρτες μνήμης. Επιλέξτε ένα αρχείο κάρτας για να πραγματοποιήσετε την επαναφορά.",
+    "cloud_restore_success": "Αποθήκευση της αποκατάστασης στην κάρτα μνήμης σας",
```
Moyasee: This is the worst of the reversals. It is the success toast shown after a restore completes ([emulation-save-modals.tsx:129](https://github.com/hydralauncher/hydra/blob/main/src/renderer/src/pages/settings/emulation/emulation-save-modals.tsx#L129)).

`Αποθήκευση της αποκατάστασης στην κάρτα μνήμης σας` parses as "saving of the restoration onto your memory card". Both nouns are the wrong part of speech.

`Η αποθήκευση επαναφέρθηκε στην κάρτα μνήμης σας`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "not_detected": "Δεν εντοπίστηκε",
+    "synced": "Συγχρονισμένο",
+    "executable_missing": "Η διαδρομή δεν υπάρχει",
+    "executable_missing_hint": "Το αρχείο {{name}} μετακινήθηκε ή διαγράφηκε. Εντοπίστε το ξανά ή ορίστε μια νέα διαδρομή.",
+    "bios_note": "Τα αρχεία BIOS διαχειρίζεται το {{name}}. Το Hydra δεν μπορεί να εγκαταστήσει το BIOS για εσάς. Ρυθμίστε τα μέσα από το {{name}}.",
+    "remove_rom_folder_title": "Διαγραφή του φακέλου ROM",
+    "remove_rom_folder_description": "Το {{path}} δεν θα ελέγχεται πλέον. Ο φάκελος και τα αρχεία του παραμένουν στον δίσκο σας.",
+    "cancel_remove": "Ακύρωση",
+    "stat_games": "ΠΑΙΧΝΙΔΙΑ",
+    "stat_storage": "ΑΠΟΘΗΚΕΥΣΗ",
```
Moyasee: `ΑΠΟΘΗΚΕΥΣΗ` is the act of saving. This labels the disk usage tile ([emulator-detail.tsx:579](https://github.com/hydralauncher/hydra/blob/main/src/renderer/src/pages/settings/emulation/emulator-detail.tsx#L579)), whose value is a byte size and whose caption is `Σε {{count}} αρχεία ROM`.

`ΧΩΡΟΣ` stays short enough for the tile, `ΑΠΟΘΗΚΕΥΤΙΚΟΣ ΧΩΡΟΣ` if it fits.

Worth fixing for a second reason. Big Picture `exact` has the same translation, so `Storage` and `Save` both map to `Αποθήκευση`. `i18n.tsx` builds one reverse map keyed on the translated string (L22-28), so one of the two sources is overwritten and a language switch can re-translate the wrong one.

Moyasee (commented):
Re-checked at 8ecc771. Everything from the last round is fixed: register is now uniformly formal (0 informal-singular strings left in either file, down from 35), the `Ο/Η` / `τον/την` / `του/της` pattern is applied to every `{{displayName}}` string, `Big Picture` is left verbatim, and `el` is back in alphabetical position in `language-flags.ts`.

Key parity is exact now, and better than most locales: 1637/1637 desktop, 796/796 Big Picture. `ru`, `pt-BR` and `es` are all short on the Big Picture `exact` map, `el` is not. `origin/main` is the PR base, so there is nothing left to re-sync.

Mechanics clean: 0 interpolation placeholder mismatches, 0 Trans tag mismatches, all 38 plural sets present with distinct singular and plural forms, no empty values, no CR characters, `prettier --check` passes on all eight files, `tsc --noEmit -p tsconfig.web.json` exits 0.

What is left is content, not structure. Seven strings are wrong, four of them because the head noun flipped and the phrase now means something different from the English. Those are the ones worth fixing before merge. The rest is terminology and register drift between strings that render next to each other, cheap to clean up while you are in there. All inline.

One of these is on me: `settings.cloud_backup` came back as the English word `Backup` after I flagged it for length last round. I should have suggested a shorter Greek term instead of just pointing at the character count.

The PR still needs a `translation` label, the template requires one for release note generation.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "classics": "Κλασικά",
+    "pinned": "Σε προβολή",
+    "platform": "Πλατφόρμες",
+    "platform_all": "Όλα",
+    "platform_pc": "PC",
+    "platform_classics": "Κλασικά",
+    "sort_by": "Ταξινόμηση κατά",
+    "achievements_earned": "Επιτεύγματα που έχουν αποκτηθεί",
+    "played_recently": "Παίχτηκαν πρόσφατα",
+    "playtime": "Ώρα παιχνιδιού",
```
Moyasee: `Ώρα` is a clock hour, not elapsed time. `total_play_time` on the very next line already uses `χρόνος`.

`Χρόνος παιχνιδιού`, and the same at `game_launcher.playtime` on L1609.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "profile_reported": "Αναφορά προφίλ",
+    "your_friend_code": "Ο κωδικός φίλου σας:",
+    "copy_friend_code": "Αντιγραφή κωδικού φίλου",
+    "copied": "Αντιγράφηκε!",
+    "change_profile_picture": "Αλλαγή φωτογραφίας προφίλ",
+    "crop_profile_image_stage": "Περιοχή περικοπής της εικόνας προφίλ. Χρησιμοποιήστε τα βελάκια για να μετακινήσετε την εικόνα.",
+    "crop_profile_picture": "Προσαρμογή φωτογραφίας προφίλ",
+    "crop_profile_banner": "Προσαρμογή banner",
+    "crop_profile_image_description": "Σύρετε την εικόνα για να την επανατοποθετήσετε και, στη συνέχεια, χρησιμοποιήστε τη λειτουργία ζουμ για να αλλάξετε το μέγεθός της.",
+    "zoom": "Μεγέθυνση",
```
Moyasee: `zoom` and `zoom_in` on L1469 are both `Μεγέθυνση`. They are the slider aria-label and the plus button aria-label, rendered next to each other in `image-crop-modal.tsx`, so a screen reader announces the two controls identically.

`Ζουμ` for the slider matches `crop_profile_image_description` on L1466, which already calls it `λειτουργία ζουμ`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "reset": "Επαναφορά",
+    "zoom_in": "Μεγέθυνση",
+    "zoom_out": "Σμίκρυνση",
+    "toggle_grid": "Εμφάνιση/απόκρυψη πλέγματος",
+    "rotate": "Περιστροφή",
+    "applying_crop": "Εφαρμογή…",
+    "apply_crop": "Εφαρμογή",
+    "upload_banner": "Ανεβάστε το banner",
```
Moyasee: Two things in this block.

`Ανεβάστε το banner` is imperative while `remove_banner` on L1479 is nominal, and both are buttons in the same component. And `uploading_banner` on the next line is first person singular, `Ανεβάζω το banner…`, which is the app speaking as "I". Nothing else in either locale does that.

The imperative/nominal split turns up in a few other spots where the two forms sit side by side: `system_tray.open` on L1341 is `Ανοίξτε το Hydra` next to `quit` `Έξοδος`, and `downloads.keep_downloading` on L749 is `Όχι, συνεχίστε τη λήψη` next to `yes_cancel` `Ναι, ακύρωση`.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "download_sources": "Πηγές λήψης",
+    "compatibility_requirements": "Συμβατότητα με το Linux",
+    "protondb_minimum": "Ελάχιστο επίπεδο ProtonDB",
+    "steam_deck_minimum": "Ελάχιστες απαιτήσεις για το Steam Deck",
+    "steam_deck_compatible": "Συμβατό με το Steam Deck",
+    "compatibility_any": "Οποιοδήποτε",
+    "protondb_silver_plus": "Ασημένιο+",
+    "protondb_gold_plus": "Χρυσό+",
+    "protondb_platinum_only": "Μόνο Πλατινένιο",
+    "deck_playable_plus": "Παίξιμο+",
```
Moyasee: The Deck half of the tier alignment got missed. The ProtonDB side is consistent now, `Ασημένιο+` here matches `protondb_tier_silver` on L271, but `Παίξιμο+` does not match `deck_playable` on L278, which is `Μπορεί να παιχτεί`. Catalogue filter and game page badge still say different things for the same tier.

`Παίξιμο` is also a noun ("playing") in what is now an adjective set. Something like `Παίζεται+` here with `deck_playable` shortened to `Παίζεται` would line the two up.

Smaller version of the same thing one line down: `deck_verified_only` is `Μόνο επαληθευμένα`, plural, while `deck_verified` on L277 is `Επαληθευμένο`, singular.

src/locales/el/translation.json
```diff
@@ -0,0 +1,1687 @@
...
+    "friend_started_playing_game": "Ο/Η {{displayName}} άρχισε να παίζει ένα παιχνίδι",
+    "test_achievement_notification_title": "Αυτή είναι μια δοκιμαστική ειδοποίηση",
+    "test_achievement_notification_description": "Πολύ ωραίο, ε;",
+    "scan_games_complete_title": "Η αναζήτηση παιχνιδιών ολοκληρώθηκε με επιτυχία",
+    "scan_games_complete_description": "Εντοπίστηκαν {{count}} παιχνίδια χωρίς καθορισμένη διαδρομή εκτέλεσης",
+    "scan_games_no_results_title": "Ολοκληρώθηκε η αναζήτηση παιχνιδιών",
+    "scan_games_no_results_description": "Δεν εντοπίστηκαν εγκατεστημένα παιχνίδια"
+  },
+  "system_tray": {
+    "open": "Άνοιγμα Hydra",
```
Moyasee: Missing the article. `open_winetricks` on L520 is `Άνοιγμα του Winetricks`, so `Άνοιγμα του Hydra` here and at `game_launcher.open_hydra`.

Moyasee (commented):
Re-checked at 6fefeba. All 14 are fixed, and you went past a few of them: normalising the ProtonDB tiers onto one grammatical form (`Χάλκινο`, `Ασημένιο`, `Χρυσό`, `Πλατινένιο`, all neuter adjectives) and pulling the catalogue filters into line with the badges is a better fix than what I asked for. `achievement.platinum` and `change_achievement_sound` were picked up on your own, I had missed both.

Mechanics still clean at this head: 1637/1637 and 796/796, 0 placeholder mismatches, 0 Trans tag mismatches, no CR characters, `prettier --check` passes, `tsc --noEmit -p tsconfig.web.json` exits 0. The Big Picture reverse map no longer has `Storage` and `Save` fighting over `Αποθήκευση`. Duplicate collapses across the desktop file went from 25 to 24 with the `zoom` fix, and the rest of those are pre-existing patterns like Download/Downloading that every locale has.

Two small things left, both inline. Fix those and I will approve.

The PR still has no label. It needs `translation` before it can be merged.

NickCh11:
Addressed in `75a208795`.

- Aligned the Steam Deck filter and badge labels: `Παίζεται+` / `Παίζεται` and `Μόνο Επαληθευμένο` / `Επαληθευμένο`.
- Updated both Open Hydra labels to `Άνοιγμα του Hydra`.
- All review threads are resolved; locale parity, Prettier, and TypeScript checks pass.

I also checked the repository labels, but there is currently no `translation` (or i18n/localization) label available for me to apply.

Moyasee (approved):
Both are fixed at 75a2087. Deck tiers now line up on both sides, `Παίζεται+` in the catalogue against `Παίζεται` on the game page and `Μόνο Επαληθευμένο` against `Επαληθευμένο`, and the tray and launcher both use `Άνοιγμα του Hydra`.

Re-ran everything on the final head:

```
desktop keys        1637 / 1637
big picture keys     796 / 796
placeholder mismatch    0
Trans tag mismatch      0
empty values            0
CR characters           0
informal-register hits  0  (both files)
prettier --check     pass  (8 files)
tsc --noEmit         exit 0
eslint               0 errors
```

The two eslint warnings that show up are `no-explicit-any` on `src/shared/index.ts` L89 and L293, both on lines this PR never touches. They report identically on `main`.

Approving. Nice work grinding through this many rounds, and the tier normalisation in 6fefeba was a better call than the fix I suggested.

One thing before you can merge, and I cannot set it from here: the PR needs a `translation` label. Release notes are categorised off it, so the template requires one.

PR #2537, opened 2026-07-15
Feature/lbx 948 Add native local snapshot pipeline

Hachi-R:


native/hydra-native/src/cloud_save/pipeline/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    build_local_game_snapshot, BuildLocalGameSnapshotInput, LocalGameSnapshotWithHash,
+};
+use super::manifest::types::CloudSaveGameId;
+use super::manifest::{get_save_rules_for_game, GetSaveRulesForGameInput};
+use super::path_resolution::{resolve_save_rules, ResolveSaveRulesInput};
+use super::save_scanner::{scan_resolved_save_rules, ScannedCloudSaveRule};
+
+pub use types::BuildLocalGameSnapshotPipelineInput;
+
+fn collect_discovered_files(
```
Moyasee, line 24: `collect_discovered_files` has the only real logic in this PR (priority ranking + per-path dedup) and no tests, whereas the rest of the stack unit-tests each module. Worth covering: same absolute_path from two rules with different priority (higher wins), equal-priority tie-break (currently first-rule-in-`scanned_rules`-order wins, which decides the `raw_path` attribution and therefore the aggregate hash, so worth pinning down that it's intended), and the file-exists vs dir vs dynamic-only cases that drive the 2/1/0.

native/hydra-native/src/cloud_save/pipeline/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    scanned_rules: Vec<ScannedCloudSaveRule>,
+) -> Vec<DiscoveredLocalSaveFile> {
+    let mut discovered_by_path = BTreeMap::new();
+
+    for rule in scanned_rules {
+        let priority = if rule
+            .resolved_paths
+            .iter()
+            .any(|path| !path.dynamic && Path::new(&path.path).is_file())
```
Moyasee, line 33: `Path::is_file()` is a blocking fs stat running on the async worker, and it re-stats paths the scanner already walked. For a game with many rules/paths this adds up. Could derive priority from the scan results instead, or move it off the async thread.

Moyasee (commented):
Orchestration reads clean and the `?` propagation is right. Two things: the dedup/priority helper is the only non-trivial logic here and it's untested (every other module in this stack ships unit tests), and it does sync fs I/O inside the async fn. Details inline.

native/hydra-native/src/cloud_save/pipeline/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+use super::path_resolution::{resolve_save_rules, ResolveSaveRulesInput};
+use super::save_scanner::{scan_resolved_save_rules, ScannedCloudSaveRule};
+
+pub use types::BuildLocalGameSnapshotPipelineInput;
+
+const DYNAMIC_PATH_PRIORITY: u8 = 0;
+const STATIC_PATH_PRIORITY: u8 = 1;
+const STATIC_FILE_PRIORITY: u8 = 2;
+
+fn collect_discovered_files(
```
Moyasee, line 24: Still untested. The priority/dedup logic here is the part most likely to regress silently. Cases worth pinning: same absolute_path from two rules where the higher priority wins, equal-priority first-rule-wins (it decides the raw_path attribution and thus the aggregate hash), and the 2/1/0 split across file-exists / static-dir / dynamic-only.

Moyasee (commented):
The `spawn_blocking` move resolves the sync-fs concern, and the named priority constants read nicely. Still no tests on `collect_discovered_files` though, which was the bigger one.

Hachi-R:
Superseded by #2608

PR #2538, opened 2026-07-15
Feature/lbx 949 Add native cloud save upload and restore

Hachi-R:


native/hydra-native/src/cloud_save/restore/replace_targets.rs
```diff
@@ -0,0 +1,427 @@
...
+    })
+}
+
+async fn commit_target(item: &mut PreparedTarget) -> Result<(), String> {
+    let target = Path::new(&item.input.target_path);
+    if target
+        .try_exists()
+        .map_err(|_| "cloud_save_restore_target_inspection_failed".to_string())?
+    {
+        if !tokio::fs::metadata(target)
```
Moyasee, line 199: `fs::metadata` follows symlinks, so this only checks the symlink's *destination*. Two cases worth thinking about:

- If the target itself is a symlink, we move the link to backup and install a regular file, silently breaking a legitimately-symlinked save (and the real file the link pointed at is left stale).
- More importantly, `relative_path` traversal is blocked, but a symlinked *directory* somewhere inside the resolved save tree is still followed on the final `rename`/write, which can land the file outside the resolved root.

Probably fine for the common case, but given this is the write path it'd be worth either `symlink_metadata`/`O_NOFOLLOW` on the components, or an explicit decision that following symlinks in the save tree is acceptable.

native/hydra-native/src/cloud_save/restore/replace_targets.rs
```diff
@@ -0,0 +1,427 @@
...
+fn failure(input: &ReplaceRestoreTarget, reason: &str) -> RestoreFailedFile {
+    RestoreFailedFile {
+        raw_path: input.raw_path.clone(),
+        relative_path: input.relative_path.clone(),
+        target_path: input.target_path.clone(),
+        reason: reason.to_string(),
+    }
+}
+
+fn sibling_path(target: &Path, suffix: &str) -> Result<PathBuf, String> {
```
Moyasee, line 42: The staging/backup siblings live in the target's own directory with a random UUID name. On the happy path they're always cleaned (rollback / remove_backups), but a hard crash between the backup rename and remove_backups leaves `.hydra-restore-<uuid>-{stage,backup}` orphans in the save dir. The scanner globs with `require_literal_leading_dot: false`, so a `<dir>/*` rule will pick those orphans up and fold them into the next snapshot's aggregate hash. Worth either filtering `.hydra-restore-` in the scanner or sweeping them on restore start.

native/hydra-native/src/cloud_save/restore/download_blob.rs
```diff
@@ -0,0 +1,208 @@
...
+        .await
+        .map_err(|error| format!("Failed to create restore temporary directory: {error}"))?;
+    remove_if_exists(partial_path)
+        .await
+        .map_err(|error| format!("Failed to clear partial restore blob: {error}"))?;
+    let mut output = tokio::fs::File::create(partial_path)
+        .await
+        .map_err(|error| format!("Failed to create temporary restore blob: {error}"))?;
+
+    while let Some(chunk) = response
```
Moyasee, line 57: The download streams chunks with no size bound. The content is only hash-verified later (in verify + replace), so a misbehaving or hijacked signed URL can write unbounded bytes to temp before anything rejects it. The manifest already carries `size_bytes`; cheap to pass it through and bail once the written length exceeds it. Low severity given the URL is ours.

Moyasee (commented):
Really solid, careful work. The stage -> hash-verify -> backup -> install -> re-hash flow with ordered rollback is well thought out, the signed URL is stripped from every error path (and tested for), the atomic temp+rename download is right, and the relative-path traversal / drive-letter / duplicate-target guards are all here with good coverage. A few things worth a look before merge, none blocking. Details inline, plus one point on the temp download.

Hachi-R:
Superseded by #2608

Issue #2525, opened 2026-07-15
Feature Request: CLI flag to force Big Picture mode (--big-picture)


lneute:
Hi team,

I use Hydra via AppImage on Linux and I have a specific use case regarding local streaming. I don't want to leave Big Picture mode enabled by default on my desktop, but I need it to start in Big Picture automatically when I launch it remotely via Moonlight/Sunshine.

Currently, there is no way to automate this. I tried creating an isolated profile using --user-data-dir and symlinking the hydra-db directory so I could keep my game library synced. However, since the UI preference (Big Picture on/off) is tied to the same LevelDB database as the game library, the preference ends up syncing back to my normal desktop profile.

It would be incredibly helpful to have a simple CLI flag like --big-picture (or similar). This would allow us to easily automate the launcher via Sunshine, terminal scripts, or custom shortcuts without overwriting the global config.

Thanks for the great work on the launcher!

Moyasee:
Closing this since this behavior has already been implemented and would be available in the next update v4.0.7

PR #2540, opened 2026-07-15
Feature: Added --big-picture CLI flag

Tiger-Tig74:

**Fill in the PR content:**
Added a `--big-picture` flag to start Hydra in big picture mode from the CLI. Issue: #2525


src/main/index.ts
```diff
@@ -281,7 +283,11 @@ app.on("second-instance", (_event, commandLine) => {
 
       WindowManager.mainWindow.focus();
     } else {
-      WindowManager.createMainWindow();
+      WindowManager.createMainWindow({ forceBigPicture });
+    }
+
+    if (forceBigPicture) {
+      void WindowManager.openBigPictureWindow();
```
Hachi-R: `src/main/index.ts:L290` — P2: When Hydra is already running in the tray with `mainWindow === null`, `createMainWindow()` is still awaiting LevelDB when `openBigPictureWindow()` runs, so it falls back to the primary display instead of the saved/main-window display. Only call `openBigPictureWindow()` when the main window already exists; otherwise let `createMainWindow({ forceBigPicture })` open it from `ready-to-show`.

Hachi-R (changes requested):
One blocking issue.

src/main/index.ts
Tiger-Tig74: Thanks for the response. Now it calls `openBigPictureWindow()` inside the `if (WindowManager.mainWindow) {...}` block.

Hachi-R (approved):
Blocker fixed. `openBigPictureWindow()` now runs immediately only when `mainWindow` already exists; otherwise `createMainWindow({ forceBigPicture })` waits for `ready-to-show` and opens Big Picture on the correct display. CI is green.

PR #2300, opened 2026-06-09
Feat/home page

galinhaman22:
**When submitting this pull request, I confirm the following (please check the boxes):**

- [x] I have read the [Hydra documentation](https://docs.hydralauncher.gg/getting-started.html).
- [x] I have checked that there are no duplicate pull requests related to this request.
- [x] I have considered, and confirm that this submission is valuable to others.
- [x] I accept that this submission may not be used and the pull request may be closed at the discretion of the maintainers.

**Fill in the PR content:**

Redesigns the Home tab with an auto-rotating Hero carousel and dynamic catalogue rows (Hot, Top Reviewed on Lerna, genres, tags, classics per-platform/genre, Because You Played, Friends Playing Now, etc.) plus session-seeded shuffling, windowing-based scroll smoothness, and per-row levelDB caching.

PR #2542, opened 2026-07-15
[fix] Redesign of the Home menu

Ahmed-Ouederni:

- [ x] I have read the [Hydra documentation](https://docs.hydralauncher.gg/getting-started.html).
- [ x] I have checked that there are no duplicate pull requests related to this request.
- [ x] I have considered, and confirm that this submission is valuable to others.
- [x ] I accept that this submission may not be used and the pull request may be closed at the discretion of the maintainers.

**Fill in the PR content:**

The current home screen shows only 12 games across three static categories (Hot Now, Top Games This Week, Games to Beat), plus a "Surprise Me" shuffle button. It's the same for every user regardless of their library or play habits, so it doesn't do a great job of surfacing games people would actually be interested in. As a result, I find myself skipping it entirely when I want to discover something new, which feels like a missed opportunity for a launcher's landing page.
A more personalized, dynamic home screen, for example:

image

image

This is my humble contribution to this massive project and I hope it satisfies all the Hydra launcher users.


Moyasee:
Thanks for putting this together, and for including the screenshots. The critique is fair. The home screen is static today and does not do much to surface anything a given user would actually care about.

I have to close this one though. A home redesign has been in flight internally for a while on `feat/home-page` (#2300), paired with curation endpoints built on the API side specifically for it. Rows there arrive pre-curated rather than being assembled in the client, so the `POST /catalogue/search` call this PR adds for the Classics row is the exact pattern that branch removes. Merging both would mean undoing one of them.

That is a timing collision, not a judgement on the work. Sorry you spent the effort without knowing the other branch was there.

Two things worth knowing for next time:

- `game-card.scss` is shared with the catalogue and library pages, so the height change and the hardcoded `#1c1c1c` would have shifted those too. Prefer a theme variable, and scope layout overrides to the page that needs them.
- New translation keys need to land in the other locales, not just `en`. Same for removing the emoji from `home.weekly` and `home.achievements`, which would have left `en` out of sync with every other file.

Thanks again for taking a swing at it.

PR #2543, opened 2026-07-16
fix big picture user profile crash

Hachi-R:


src/big-picture/src/pages/profile/use-profile-data.ts
```diff
@@ -0,0 +1,580 @@
...
+      games,
+      totalCount: response.totalCount ?? games.length,
+    };
+  }
+
+  const games: UserGame[] = [];
+  let totalCount = 0;
+  let hasMore = true;
+
+  while (hasMore) {
```
Moyasee: The old loop checked `isMounted` between pages, so pagination stopped when you left the page. Now it runs to completion even after unmount, and with the `key` remount that happens on every profile navigation, so a large library keeps firing requests in the background. Worth passing a cancellation check into `fetchRemoteLibraryGames`.

src/big-picture/src/pages/profile/profile.tsx
```diff
@@ -521,1511 +423,1546 @@ function getProfileLibraryGameItemId(game: ShopAssets) {
...
-    down: firstContentFocusId
-      ? {
-          type: "item",
-          itemId: firstContentFocusId,
-        }
-      : {
-          type: "block",
-        },
-  };
+  if (isLoading) return null;
```
Moyasee: This blanks the whole page while the `/users/:id` request is in flight. Before, the hero rendered right away and the own profile rendered everything, since local data doesn't depend on this fetch. Was hiding everything intentional, or could the hero stay visible during load?

src/big-picture/src/pages/profile/profile.tsx
```diff
@@ -158,13 +147,15 @@ type ProfileHeroAction = {
...
-const WEEKLY_BAR_PLACEHOLDER = [0.62, 0.14, 0.36, 0.26, 0.34, 0.42, 0.72];
-const WEEKDAY_LABELS = ["M", "S", "T", "W", "T", "F", "S"];
-const PROFILE_RECENT_ACHIEVEMENT_GROUP_LIMIT = 2;
-const PROFILE_RECENT_ACHIEVEMENTS_PER_GAME = 2;
-const PROFILE_RECENT_ACHIEVEMENT_LIBRARY_TAKE = 12;
-const PROFILE_FRIENDS_LIMIT = 5;
-const PROFILE_REMOTE_LIBRARY_PAGE_SIZE = 12;
+const WEEKLY_BAR_PLACEHOLDER = [
+  { day: "monday", label: "M", height: 0.62 },
+  { day: "tuesday", label: "S", height: 0.14 },
```
Moyasee: The day keys don't match the labels here: tuesday gets "S", friday gets "T", etc. If the bars start on Monday the labels should be M T W T F S S.

Moyasee (approved):
Refactor reads faithful to the old behavior, and typecheck + lint pass locally. Nice cleanup.

A few minor things inline. Also `ensureArray`, `normalizeUserProfile`, `PROFILE_FRIENDS_LIMIT` and `ProfileRecentAchievement` are exported but only used inside `use-profile-data.ts`, so they could be module-private.

PR #2546, opened 2026-07-17
Chore/assets modal cleanup

Hachi-R:


src/renderer/src/pages/game-details/modals/game-artwork-picker.tsx
```diff
@@ -138,6 +148,252 @@ interface GameArtworkPickerProps {
...
+    if (event.currentTarget.hasPointerCapture(event.pointerId)) {
+      event.currentTarget.releasePointerCapture(event.pointerId);
+    }
+  };
+
+  const handleWheel = (event: React.WheelEvent<HTMLDivElement>) => {
+    const element = scrollRef.current;
+    if (!element) return;
+
+    event.preventDefault();
```
Moyasee, line 318: Minor, non-blocking: React 18 binds `wheel` as a passive listener on the root, so this `preventDefault()` is a no-op (you'll probably see the "Unable to preventDefault inside passive event listener" warning in the console). The manual `scrollTop += scrollDelta` below still does its job, but the default wheel scroll can't actually be cancelled here, so wheeling over the custom scrollbar gutter can leak scroll to an ancestor. If you want the cancel to stick, a native non-passive listener does it: `el.addEventListener("wheel", handler, { passive: false })` in a useEffect. Low priority since it only affects the narrow scrollbar strip.

Moyasee (approved):
Went through the whole thing, not just the diff: custom scrollbar component, the assets-settings/options-modal/sidebar tweaks, and all 33 locale files. `typecheck:web` and eslint both pass clean on the branch. react-virtual is 3.14.3 so the new `gap`/`paddingEnd` options are supported, the `CloudOfflineIcon` import removals leave no dangling references, and the `custom_assets_not_sync` key lands in the `sidebar` namespace across every locale (matching `useTranslation("sidebar")` in the picker), with real translations rather than English copies. Nice touch parking the 6px scrollbar inside the content's 10px right padding so it never overlaps the tiles.

One minor non-blocking note left inline on the wheel handler. LGTM.

Issue #2045, opened 2026-02-27
Feature: Version tracker ?


steeepaaaan:
It would be nice to have a built in version tracker. Because i when you have a lot of games, it's hard to have any idea about what version a game is.

So could we have an option for managing game versions? Even a manual setting would be better that nothing

philipmoe:
yes pls, maybe and option for auto updating to latest version of your selected source has it

DRKCTRLDEV:
This would be amazing, it'd be even better if it were integrated with steamdb to tell you whether the game has a newer version that is not yet available for download (from your sources)... I'm new to the project but I dont believe this is implemented yet... (ankergames for example has this implemented to see whether or not the source is up to date) 😀

steeepaaaan:
Yeah even just an option to manually add a version code, so you could keep better track of the games you have.

PR #2551, opened 2026-07-17
Feature: track installed repack version and show update-available action

Cauatn:

**Fill in the PR content:**

Closes the first slice of #2045 (version tracker). With a lot of games in the library it's hard to know what version is installed, or whether the download source has since pushed a newer build.

This PR:

- Extracts a best-effort version string (`v1.2.3` / `Build 12345`) from the title of the repack the user picks when starting a download, and stores it on the `Game` record (`version?: string | null`).
- Compares that stored version numerically (not string equality — `1.10` correctly sorts after `1.9`) against every repack currently offered by the download source.
- When a strictly newer version is found, shows an "Update available (vX.Y.Z)" action button in the game's hero panel, styled and placed identically to the other action buttons (Play/Download, favorite, pin, options) rather than replacing the primary action — clicking it opens the repacks modal so the user can grab the new version.

Deliberately out of scope for this first PR (happy to follow up):
- Manual version editing by the user.
- Treating "different" as "newer" via anything beyond numeric segment comparison (no full semver parsing).
- Any integration with external version sources (e.g. SteamDB).

No new dependencies, no DB migration needed (LevelDB is schemaless — new field is optional and simply absent on old records).

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -246,6 +246,7 @@
     "download_options_one": "{{count}} download option",
     "download_options_other": "{{count}} download options",
     "updated_at": "Updated {{updated_at}}",
+    "update_available": "Update available ({{version}})",
     "install": "Install",
     "resume": "Resume",
     "pause": "Pause",
--- a/src/main/events/torrenting/add-game-to-queue.ts
+++ b/src/main/events/torrenting/add-game-to-queue.ts
@@ -31,6 +31,7 @@ const addGameToQueue = async (
     fileSize,
     fileIndices,
     selectedFilesSize,
+    version,
   } = payload;
 
   const parsedFileSize = parseBytes(fileSize ?? null);
@@ -72,7 +73,7 @@ const addGameToQueue = async (
     return handleDownloadError(err, downloader);
   }
 
-  await prepareGameEntry({ gameKey, title, objectId, shop });
+  await prepareGameEntry({ gameKey, title, objectId, shop, version });
 
   try {
     await downloadsSublevel.put(gameKey, download);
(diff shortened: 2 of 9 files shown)


Moyasee:
Thanks for this. The PR is well scoped, the description is clear about what is and is not included, and the no-dependency, no-migration approach is the right instinct.

I tested the extraction against real data rather than reasoning about it, running `extractVersionFromTitle` and `compareGameVersions` over roughly ten thousand real download option titles covering a wide range of version formats. That is where I landed on not taking it.

About a third of titles yield no version at all, and the rate swings from a few percent to over forty depending purely on which naming convention the listing follows. A large part of that gap comes down to single formatting details. As one example, the current pattern matches `Build 12345678` but not `Build-12345678`, and that one character accounts for thousands of misses on its own.

That specific case is easy to fix, but it illustrates the deeper problem: there is no shared convention to target. Across the sources users actually add, several hundred thousand titles, every source has its own ad-hoc format, and each would need tuning that a single regex cannot carry.

The comparison side has the same issue. Taking games that appear in more than one source, roughly half never produce a usable signal at all, either because one side has no parseable version or because the two sides use different schemes, where `compareGameVersions` returns `null` and the update is silently suppressed.

Where it does fire, `reduce` takes the maximum, so a single mis-parsed listing wins outright rather than being averaged away. Real shapes I hit, with the listings for one game on the left and what the UI would show on the right:

```
1.0.14 | 2.4.0 | 1007      ->  "Update available (1007)"
17.2 | 2026.2.241.0        ->  "Update available (2026.2.241.0)"
25.07.01 | 20250701        ->  "Update available (20250701)"
```

That is around 11% of firing cases, and it gets worse with every source added, since each one is another chance to draw a value that parses high and then wins the max.

Worth mentioning too: `DownloadSourcesChecker` already writes `newDownloadOptionsCount` and renders a `+N` badge on the sidebar game item, cleared when the user opens the game's download options, with a settings toggle. It covers the same need server side, with full coverage and no title parsing.

Closing this one, and #2045 stays open. Nice work on the scoping and the writeup.

PR #2555, opened 2026-07-18
feat: add release date and update date sorting to library

murilogiatti:
## Description
Adds two new sorting options to the library:
- **Release Date**: Fetches missing dates from Steam API in the background and allows chronological sorting.
- **New Updates**: Fetches the latest repack upload dates across all installed sources, saving it locally to sort by the newest repacks.

### Technical Details
- Fetches are chunked and batched to prevent rate limiting or freezing the UI thread.
- Fallback sorting uses the exact logic previously available, so if dates are unavailable it will gracefully fallback to the existing unread badges count.
- Bypasses clearing new update badges on manual sort fetches so users don't lose their green dot notifications on sort changes.


src/main/events/library/refresh-library-update-dates.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    // Fetch in background
+    (async () => {
+      try {
+        const BATCH_SIZE = 5;
+        const MAX_DOWNLOADS_PER_GAME = 100;
+        const DOWNLOADS_SKIP_OFFSET = 0;
+        let updatedCount = 0;
+
+        const chunks = chunk(nonCustomGames, BATCH_SIZE);
```
Moyasee, line 102: This sweeps the whole library on every Library mount. The release-date path has a `releaseDateLastCheckedAt` guard, this one has only `isFetching`, which dedupes a run in flight but not repeat runs. The `[sortBy]` effect at `library.tsx:222` fires on mount and Library is routed, so every navigation back with this sort persisted issues one request per non-custom game plus a full `checkForNewUpdates` sweep. Needs the same last-checked guard.

src/main/events/library/refresh-library-release-dates.ts
```diff
@@ -0,0 +1,111 @@
...
+            )) {
+              const details = await Promise.all(
+                gamesBatch.map((game) =>
+                  getSteamAppDetails(game.objectId, "en")
+                )
+              );
+
+              for (const [index, game] of gamesBatch.entries()) {
+                const detailsForGame = details[index];
+                if (!detailsForGame) continue;
```
Moyasee: Rate-limited games retry forever. `getSteamAppDetails` returns null on any error including a 429, this hits `continue`, so `releaseDateLastCheckedAt` is never written and the game stays in `missingSteamGames` for the next run. The 500ms sleep is only between chunks of 50, and inside a chunk it fires 10 back-to-back batches of 5 with nothing between them, so a large library trips the limit and never converges. Stamp the checked-at on failure, or distinguish a 429 from a real miss and back off.

Same file, line 23: a game returning "Coming soon" gets `releaseDateTimestamp: null` and a `releaseDateLastCheckedAt` stamp, so it is excluded permanently and never picks up its date once it actually releases.

src/main/events/library/refresh-library-release-dates.ts
```diff
@@ -0,0 +1,111 @@
...
+                    game.objectId,
+                    "english"
+                  ),
+                  { ...detailsForGame, name: game.title }
+                );
+              }
+            }
+
+            WindowManager.mainWindow?.webContents.send(
+              "on-library-batch-complete"
```
Moyasee: `WindowManager.mainWindow?.webContents.send` where all 10 pre-existing emitters of this event use `WindowManager.sendToAppWindows`. Big picture subscribes at `big-picture/src/pages/library/page.tsx:302` and never sees it. Same at `refresh-library-update-dates.ts:101`.

src/locales/en/translation.json
```diff
@@ -1777,6 +1779,8 @@
     "sort_title": "Title",
     "sort_title_asc": "Title (A-Z)",
     "sort_title_desc": "Title (Z-A)",
```
Moyasee, line 1935: `ru` and `es` are missing both keys. en and pt-BR have them, and `fallbackLng` is `"en"`, so Russian and Spanish users get English labels. Those two need the same key paths.

Separately, the `catalogue` copies at line 238 (and `pt-BR:1601`) are dead: `filter-options.tsx` reads `useTranslation("library")`, and `catalogue.tsx:585-607` still lists only alphabetical and hydraScore. Drop them rather than translating them.

src/main/services/download-sources-checker.ts
```diff
@@ -153,7 +155,9 @@ export class DownloadSourcesChecker {
 
       logger.info(`Using since: ${since} (from last app start)`);
```
Moyasee, line 203: The manual flag skips `clearStaleBadges` but line 194 still advances the baseline with `updateDownloadSourcesCheckBaseline(now)`. So after a restart the normal startup check clears everything and re-adds only what changed since the sort was selected, and the older badges this flag exists to protect are gone anyway. Skipping the baseline update on the manual path too would close it.

src/shared/library-date-sorting.ts
```diff
@@ -0,0 +1,83 @@
...
+  const isoMatch = /^(\d{4})-(\d{2})-(\d{2})$/.exec(dateStr);
+  if (isoMatch) {
+    return Date.UTC(
+      Number.parseInt(isoMatch[1], 10),
+      Number.parseInt(isoMatch[2], 10) - 1,
+      Number.parseInt(isoMatch[3], 10)
+    );
+  }
+
+  const nativeParse = Date.parse(dateStr);
```
Moyasee, line 52: Only a full `YYYY-MM-DD` takes the explicit regex; everything else goes through `Date.parse` (UTC for a bare year or `YYYY-MM`) and then local getters, so it shifts back a day west of UTC:

```
TZ=UTC                 "2025" -> 2025-01-01   order: A, B
TZ=America/Sao_Paulo   "2025" -> 2024-12-31   order: B, A
```

Only visible around Jan 1 boundaries, and the tests pass in both zones so they do not catch it. Using the UTC getters throughout fixes it.

src/main/events/library/check-for-new-updates.ts
```diff
@@ -0,0 +1,14 @@
+import { registerEvent } from "../register-event";
+import { DownloadSourcesChecker } from "@main/services/download-sources-checker";
+import { logger } from "@main/services";
+
+const checkForNewUpdates = async () => {
+  try {
+    // Pass true for isManualRefresh so it doesn't wipe existing badges
```
Moyasee: Comment can come out, along with the ones at `refresh-library-release-dates.ts:34` and `:95` and `refresh-library-update-dates.ts:37`. The one at `:34` says "in chunks of 50" where the inner batch is 5, so it is already out of date. That reasoning belongs in the commit message.

Two leftovers while you are in there: `renderer/src/helpers.ts:14` re-exports `parseSortableDate` and nothing imports it from there, and `refresh-library-release-dates.ts:43` computes `appids` per chunk for an error log only.

Moyasee (dismissed):
The comparator work is solid: memoized, validated against `SORT_OPTIONS` on load so a stale stored value degrades cleanly, and no seconds/millis mixing between the two fields. The refresh jobs behind it are where the problems are. Notes inline.

Branch is behind but merges clean.

src/main/events/library/refresh-library-update-dates.ts
murilogiatti: Added a 5-minute refresh throttle (`DOWNLOAD_SOURCES_REFRESH_INTERVAL_MS`) along with a `sourceSignature` check. The sweep won't re-run on repeat mounts unless 5 minutes have elapsed or the configured download sources list changes.

src/main/events/library/refresh-library-release-dates.ts
murilogiatti: Refactored Steam release date fetching with `getSteamAppDetailsWithStatus` and introduced `releaseDateNextCheckAt` to manage retry policies:
- Rate limits (429) set a global backoff and schedule next check after 1h (or `Retry-After`).
- Transient errors schedule a recheck after 1h.
- 'Coming soon' games recheck after 7 days.
- 'Not found' / confirmed missing dates recheck after 30 days.
Also added a 200ms delay between 5-game sub-batches to prevent rate-limiting spikes.

src/main/events/library/refresh-library-release-dates.ts
murilogiatti: Updated both `refresh-library-release-dates.ts` and `refresh-library-update-dates.ts` to use `WindowManager.sendToAppWindows("on-library-batch-complete")` so events reach Desktop, Big Picture, and all open app windows.

src/locales/en/translation.json
murilogiatti: Added `sort_release_date` and `sort_new_updates` keys to Spanish (`es`) and Russian (`ru`) under the `library` namespace. Removed the unused `sort_release_date` and `sort_new_updates` keys from the `catalogue` namespace in `en`, `pt-BR`, and `fr`.

src/main/services/download-sources-checker.ts
murilogiatti: Updated `DownloadSourcesChecker` so `updateDownloadSourcesCheckBaseline` is skipped when `isManualRefresh` is true, ensuring existing badges aren't lost across app restarts.

src/shared/library-date-sorting.ts
murilogiatti: Added explicit regex matching for partial ISO dates (`YYYY` and `YYYY-MM`) to construct UTC timestamps directly via `Date.UTC` before falling back to `Date.parse`, preventing timezone shifts across dates.

src/main/events/library/check-for-new-updates.ts
murilogiatti: Cleaned up all obsolete/redundant comments, removed the unused `parseSortableDate` re-export from `renderer/src/helpers.ts`, and removed the `appids` string construction in `refresh-library-release-dates.ts`.

src/main/events/library/refresh-library-release-dates.ts
```diff
@@ -0,0 +1,85 @@
...
+        try {
+          const searchParams = new URLSearchParams({
+            appids,
+            l: steamLanguage,
+            cc: "us",
+          });
+          
+          const response = await axios.get(
+            `http://store.steampowered.com/api/appdetails?${searchParams.toString()}`
+          );
```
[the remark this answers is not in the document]
murilogiatti: Resolved in the current branch: Steam app details are fetched through `getSteamAppDetailsWithStatus`, which uses `https://store.steampowered.com/api/appdetails`.

src/main/events/library/refresh-library-release-dates.ts
```diff
@@ -0,0 +1,85 @@
...
+          WindowManager.mainWindow?.webContents.send("on-library-batch-complete");
+        } catch (err) {
+          logger.error(`Failed to fetch release dates chunk for ${appids}`, err);
+        }
+        
+        await delay(500); // Prevent rate limiting between chunks
+      }
+      
+      logger.info(`Finished fetching release dates. Updated ${updatedCount} games.`);
+    })();
```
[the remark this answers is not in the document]
murilogiatti: Resolved: the module-level `isFetching` guard is set before starting the background task and cleared in its `finally` block, preventing concurrent refresh jobs.

src/main/events/library/refresh-library-release-dates.ts
```diff
@@ -0,0 +1,85 @@
...
+              updatedCount++;
+            }
+          }
+          
+          WindowManager.mainWindow?.webContents.send("on-library-batch-complete");
+        } catch (err) {
+          logger.error(`Failed to fetch release dates chunk for ${appids}`, err);
+        }
+        
+        await delay(500); // Prevent rate limiting between chunks
```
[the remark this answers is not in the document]
murilogiatti: Addressed with named timing constants: `SUB_BATCH_DELAY_MS` controls the pause between concurrent sub-batches, while rate-limit responses apply an explicit backoff.

src/main/events/library/refresh-library-update-dates.ts
```diff
@@ -0,0 +1,85 @@
...
+      
+      const promises = batch.map(async (game) => {
+        try {
+          const downloads = await HydraApi.get<GameRepack[]>(
+            `/games/${game.shop}/${game.objectId}/download-sources`,
+            {
+              take: 100,
+              skip: 0,
+              downloadSourceIds,
+            },
```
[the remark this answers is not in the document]
murilogiatti: Addressed: the request pagination values are now `MAX_DOWNLOADS_PER_GAME` and `DOWNLOADS_SKIP_OFFSET` constants.

src/renderer/src/pages/library/library.tsx
```diff
@@ -225,6 +266,12 @@ export default function Library() {
       (importing) => setIsImportingClassics(importing)
     );
 
+    if (sortBy === "new_updates") {
+      window.electron.refreshLibraryUpdateDates?.().catch(() => {});
+    } else if (sortBy === "release_date") {
+      window.electron.refreshLibraryReleaseDates?.(i18n.language).catch(() => {});
+    }
```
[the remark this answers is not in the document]
murilogiatti: Addressed: the refresh effect now depends on `sortBy`; it no longer reads `i18n.language`.

murilogiatti:
All review items have now been addressed. The latest commit also refactors the release-date refresh into smaller helpers to resolve the remaining SonarCloud cognitive-complexity finding, with rate-limit coverage added. Validation completed: `npm test` (77 passing), `npm run typecheck`, Prettier, and `git diff --check`. Requesting a fresh review.

jazir555:
This is my most requested feature, can't wait for this to merge and release. <3 @murilogiatti

src/main/services/download-sources-checker.ts
```diff
@@ -99,7 +108,12 @@ export class DownloadSourcesChecker {
     );
   }
 
-  static async checkForChanges(): Promise<void> {
+  static async checkForChanges(
```
Moyasee, line 133: The new `isChecking` static makes the startup check and the Library-mount manual check mutually exclusive, and the loser is dropped rather than queued.

`main.ts:117` fires `checkForChanges()` fire-and-forget after `await HydraApi.setupApi()`, while the renderer fires `checkForNewUpdates` into `checkForChanges(true)`, which holds the flag for the whole POST over the library. If the manual call wins that race, the session never runs `clearStaleBadges` and never advances the baseline, so stale badges survive and the next session's `since` has not moved.

The precondition is narrow: `library-sort-by` already set to `new_updates` and Library mounting while `setupApi()` is still resolving, both seconds-wide at startup. I could not reproduce it without a running app, so take it as read from the source. A separate guard per path, or queueing the loser, closes it.

src/main/events/library/refresh-library-update-dates.ts
```diff
@@ -0,0 +1,144 @@
...
+import { levelKeys } from "@main/level/sublevels";
+import {
+  getDownloadSourcesSignature,
+  shouldRefreshDownloadSources,
+} from "@main/services/download-sources-refresh-policy";
+import type { GameRepack } from "@types";
+
+import { chunk } from "lodash-es";
+
+let isFetching = false;
```
Moyasee, line 18: Nit. The throttle is process memory, so a restart resets it and a Library mount every 5 minutes or more re-sweeps one `/download-sources` request per non-custom game. The release-date path persists `releaseDateNextCheckAt` per game, so if parity was the intent this wants to live in level too.

src/main/events/library/refresh-library-release-dates.ts
```diff
@@ -0,0 +1,154 @@
...
+import {
+  getReleaseDateNextCheckAt,
+  needsReleaseDateRefresh,
+} from "./release-date-refresh-policy";
+
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+let isFetching = false;
+let nextSteamRequestAt = 0;
+const SUB_BATCH_DELAY_MS = 200;
```
Moyasee: Nit. Five concurrent appdetails requests per 200ms sleep. The 429 handling is right now, but on a large uncached library the first 429 arrives within seconds and stops the sweep for an hour, so dates trickle in over many hours. A larger delay would let the common case finish rather than back off. `nextSteamRequestAt` is also in memory, so a restart drops the global backoff and fires straight back into a throttled endpoint.

Two small ones in the same file: `chunk(Array.from(missingSteamGames), 50)` at line 87 takes a `Game[]` already, and `releaseDateLastCheckedAt` is written at 30 and 49 but never read, since `needsReleaseDateRefresh` keys entirely off `releaseDateTimestamp` and `releaseDateNextCheckAt`.

Moyasee (commented):
All seven are fixed, so this supersedes my changes-requested. You did not patch around them either, the rate-limit handling in particular came back as a proper discriminated union with distinct next-check windows, and the timezone fix holds under every zone I tried. Resolving those threads.

One substantive thing left plus a few nits, inline. Once those are sorted we can approve and merge this.

The ubuntu build is red on GitHub releases returning EOF and 503 for the snap and appimage templates, not on anything here. The same job passes on main at your exact merge base, so it just needs a re-run.

jazir555:
@murilogiatti

src/main/services/download-sources-checker.ts
murilogiatti: Addressed in 621220725. I split the in-flight guards by path (`isAutoChecking` and `isManualChecking`), so the startup check is no longer dropped when the Library-triggered manual check wins the startup race. The automatic path can still clear stale badges and advance the baseline as intended.

src/main/events/library/refresh-library-update-dates.ts
murilogiatti: Addressed in 621220725. The refresh throttle now persists `lastCheckedAt` and the download-source signature in LevelDB. A restart therefore preserves the five-minute window; a source configuration change still triggers a refresh immediately.

src/main/events/library/refresh-library-release-dates.ts
murilogiatti: Addressed in 621220725. Steam sub-batches now pause for 1.5 seconds, and the global rate-limit deadline is persisted in LevelDB. This avoids the initial burst and prevents an app restart from bypassing a 429 backoff. I also removed the redundant `Array.from` and the unused `releaseDateLastCheckedAt` field.

jazir555:
@murilogiatti Does this also feed into the filters for the catalogue page? Currently the release date data received by them is incomplete, this would be perfect there (it has less than 1/10th of the catalogues games returned) . The release date filter functionality is lacking many titles which this would resolve.

murilogiatti:
> **[jazir555](/jazir555) **

@jazir555! No, this PR specifically focuses only on the Library page to allow chronological sorting for the games you have installed/added.

jazir555:
@murilogiatti Okay I'll open a github issue request for an extension of this PR after this lands, the chronological sorting via the Steam API that this PR implements is exactly what the release date filter functionality is lacking. The Release Date filter functionality currently fails at completing its intended function because not all games have that metadata attached, and this PR adds the required metadata.

Hachi-R:
Thanks for the work and for iterating on the review comments — the extra sort options in the existing library picker are a good idea.

The data path is not something we can ship. Release date sorting does not use our API (the catalogue already has `releaseDate`); it scrapes Steam `appdetails` **once per library game**, five at a time with a 1.5s pause, and the UI only refreshes every 50 titles. Until those dates trickle in the list looks unsorted, and a 429 stalls the whole sweep for an hour.

New Updates has the same shape against our API: one `GET /games/:shop/:id/download-sources` **per non-custom game** instead of a single batched call.

Library sort has to be instant from data we already have or can fetch in bulk. Closing this approach.

jazir555:
> Thanks for the work and for iterating on the review comments — the extra sort options in the existing library picker are a good idea.
> 
> The data path is not something we can ship. Release date sorting does not use our API (the catalogue already has `releaseDate`); it scrapes Steam `appdetails` **once per library game**, five at a time with a 1.5s pause, and the UI only refreshes every 50 titles. Until those dates trickle in the list looks unsorted, and a 429 stalls the whole sweep for an hour.
> 
> New Updates has the same shape against our API: one `GET /games/:shop/:id/download-sources` **per non-custom game** instead of a single batched call.
> 
> Library sort has to be instant from data we already have or can fetch in bulk. Closing this approach.

Is this the reason the newest sort in the catalogue page doesn't return the full list of games vs the popular sort? Or is that a seperate issue? This would be great if it can be fixed.

PR #2559, opened 2026-07-19
fix: improve big picture game settings navigation

Moyasee:
Fixes the Big Picture game settings navigation issues from LBX-963, plus some related RB/LB polish.

## What changed

**Sidebar tab states (game settings modal)**

The active tab was hard to make out. The sidebar now switches tabs as focus lands on them, so moving through it changes the tab and its content directly instead of needing a separate press, and the active tab is marked with a clearer background and text highlight that matches how active states look elsewhere in the app.

**Asset type navigation (Customization tab)**

Moving the stick left from a middle asset type (e.g. Logo) jumped straight back to the sidebar instead of moving to the previous type. Tab items now carry per-item left/right focus overrides, so directional input walks between them; only the leftmost tab falls through to the sidebar.

**RB/LB hints and focus follow**

Added a shared `BumperBadge` component (pulled out of the settings page, which now reuses it) and used it to flank the customization asset tabs and the library filter tabs, so RB/LB switching is discoverable. RB/LB now also moves focus, so the focus ring follows the selected tab instead of staying behind. The segmented indicator also recomputes its width on resize, so it renders full width from the first paint instead of only after the first tab change.

**Library grid cover flash**

Switching from an empty collection to one with games mounted the grid fresh, and its column count started at 1 until an effect corrected it after paint, so the first cover flashed at full screen width for a frame. The column count is now seeded from the viewport up front and corrected in a layout effect before paint.

## Screenshots

**Sidebar active state**

image

**RB/LB hint (asset types)**

image

**RB/LB hint (library collections)**

image

## Notes

- The library filter tabs badge and the grid cover flash fix are unrelated to LBX-963, but they touch the same area / reuse the same component, so they are bundled here.
- The Linear ticket also mentions the segmented asset tabs showing two outlines. That is not touched here, can follow up separately if we want it.

Linear: LBX-963

### DIFF
--- a/src/big-picture/src/components/common/bumper-badge/index.tsx
+++ b/src/big-picture/src/components/common/bumper-badge/index.tsx
@@ -1,0 +1,16 @@
+import "./styles.scss";
+
+import cn from "classnames";
+
+interface BumperBadgeProps {
+  label: "LB" | "RB";
+  className?: string;
+}
+
+export function BumperBadge({ label, className }: Readonly<BumperBadgeProps>) {
+  return (
+    <span className={cn("bumper-badge", className)} aria-hidden="true">
+      {label}
+    </span>
+  );
+}
--- a/src/big-picture/src/components/common/bumper-badge/styles.scss
+++ b/src/big-picture/src/components/common/bumper-badge/styles.scss
@@ -1,0 +1,14 @@
+.bumper-badge {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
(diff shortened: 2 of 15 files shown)


Moyasee:
Good call, pulled those out into named SCSS variables at the top of the file.

@greptile review

Hachi-R (approved):
LGTM! The navigation improvements look solid, and removing the save-folder action makes sense for a gamepad-focused Big Picture experience. Nice work!

PR #2561, opened 2026-07-19
fix: keep big picture sidebar open while virtual keyboard is active

Moyasee:

**Fill in the PR content:**

### What

Fixes the Big Picture sidebar flickering (closing and reopening) every time you move around the virtual keyboard while typing in the sidebar search with a gamepad.

### Why it happened

In gamepad mode the sidebar is only kept open by `sidebarFocusWithin`, which tracks focus capture on the search input. Navigating a virtual keyboard key moves DOM focus onto the key button, which lives outside the sidebar, so `onBlurCapture` collapses the sidebar. The virtual keyboard then refocuses the search input on the next frame, reopening it. That blur/refocus on every key press is the flicker.

### The fix

While the virtual keyboard is open and targeting the sidebar search input, the sidebar is now forced open (`virtualKeyboardOnSidebarSearch`), so the transient blur no longer collapses it. The virtual keyboard store already tracks its target, and the search input already carries the `data-sidebar-library-search` marker, so no extra plumbing was needed. The target resets to `null` when the keyboard closes, so nothing stays pinned open afterwards.

### How to test

1. Open Big Picture with a gamepad.
2. Navigate to the sidebar search and open the virtual keyboard.
3. Move between keys / type. The sidebar should stay open with no flicker.

### DIFF
--- a/src/big-picture/src/components/common/virtual-keyboard/index.tsx
+++ b/src/big-picture/src/components/common/virtual-keyboard/index.tsx
@@ -933,6 +933,14 @@ export function VirtualKeyboardProvider() {
     enter();
   }, [enter, pulseKey]);
 
+  const hotkeyBack = useCallback(() => {
+    const keepFocusOnTarget =
+      target instanceof HTMLElement &&
+      target.dataset.sidebarLibrarySearch === "true";
+
+    closeKeyboard(keepFocusOnTarget);
+  }, [closeKeyboard, target]);
+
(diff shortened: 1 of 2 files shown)


Hachi-R (approved):
LGTM! The fix is focused and handles both keeping the sidebar open during virtual keyboard interaction and restoring focus correctly when closing it with the B button. Nice work!

PR #2563, opened 2026-07-19
feat: sidebar and library redesign

galinhaman22:
Redesign of the sidebar and library, plus related fixes.

**Sidebar** — sort/filter flyout (platforms, sort, favorites-first) and an installed-only toggle.

**Library** — Collections / Sort / Consoles as consistent dropdowns that close on scroll; auto-hiding controls bar on scroll; virtualized grid; reworked game-cover fade (subtle bottom band, hidden when a game has no achievement bar); classic games with a custom/SGDB cover now render unblurred with the fade.

**Fixes** — Downloads tab and achievements hero now use manually-uploaded custom assets (hero/logo/cover); respect the "logo + hero" option for classic games; remove unused locale keys and add es/pt-BR/ru translations.


src/renderer/src/helpers.ts
```diff
@@ -329,3 +358,114 @@ export const getClassicsLaunchErrorSystem = (
...
+  return 0;
+};
+
+const compareLibraryGamesByTitle = (
+  a: LibraryGame,
+  b: LibraryGame,
+  ascending = true
+): number =>
+  ascending
+    ? a.title.localeCompare(b.title, undefined, { sensitivity: "base" })
```
chubbygrannychaser: `sortLibraryGames` now runs before the sidebar renders and calls `localeCompare` directly on the title. Library data can be stale or malformed at runtime, so a missing title will crash the sidebar instead of leaving that game searchable/sortable with an empty fallback. Please normalize both operands here (for example `(a.title ?? "").localeCompare(b.title ?? "")`) and apply the same guard to the new sidebar search handler, which also calls `game.title.toLowerCase()` directly. The library page already handles this with `game.title ?? ""`.

chubbygrannychaser (changes requested):
One blocking robustness issue in the new shared sidebar sorting path.

src/renderer/src/context/collection-context-menu/collection-context-menu.context.tsx
```diff
@@ -0,0 +1,353 @@
...
+        `/profile/games/collections/${activeCollection.id}`,
+        {
+          data: { name: nextName },
+          needsAuth: true,
+        }
+      );
+
+      await loadCollections();
+      showSuccessToast(t("collection_renamed"));
+      handleCloseRenameCollectionModal();
```
chubbygrannychaser: This call is a no-op after a successful rename: `isRenamingCollection` is still true until the `finally` block, and `handleCloseRenameCollectionModal` immediately returns while that flag is true. The success toast appears but the modal remains open. The delete path has the same ordering problem. Clear the busy state before closing, or make the success path close the modal state directly.

chubbygrannychaser (commented):
Additional blocking regression found during the full review.

chubbygrannychaser (commented):
One more blocking issue without a resolvable changed diff line: `LibraryGameCard` now prioritizes `customCoverImageUrl`, but its fallback/error state resets only when `game.id` changes. Custom asset updates retain the same id, so a card that already fell back from a failed source can skip a newly selected custom cover until it unmounts. Reset `fallbackIndex` and `imageError` when the ordered candidate URLs change, and add a same-id rerender regression test.

chubbygrannychaser (changes requested):
Re-reviewed the current head. The previously requested title-safety, collection-modal, and custom-artwork reset fixes are present. One blocker remains: the sidebar still defaults `sidebar-favorites-first` to enabled when no saved preference exists, which silently reorders existing users' libraries after upgrade. Please default an absent value to `false` (or provide an explicit migration/opt-in) before merge.

src/renderer/src/components/context-menu/context-menu.tsx
```diff
@@ -41,14 +44,39 @@ export function ContextMenu({
...
     Record<string, React.CSSProperties>
   >({});
+  const [hasModalOpen, setHasModalOpen] = useState(false);
+
+  useEffect(() => {
+    const updateModalState = () => {
+      const modalOpen = document.querySelector("[data-hydra-dialog]") !== null;
+      setHasModalOpen((prev) => (prev === modalOpen ? prev : modalOpen));
+    };
+
```
Moyasee, line 56: Every mounted menu observes the entire document even while invisible, and every mutation batch makes every observer run a document-wide query. Sidebar rows and accumulated profile cards turn this into `O(menus × DOM size)` work. Only observe while visible, or move dialog detection into one shared store.

src/renderer/src/pages/profile/profile-content/user-library-game-card.tsx
```diff
@@ -314,6 +351,15 @@ export function UserLibraryGameCard({
           {renderCoverMedia()}
         </button>
       </li>
+      <GameContextMenu
```
Moyasee: Every profile card now mounts a complete hidden `GameContextMenu`, including an `onGamesRunning` IPC subscription, selectors, modal state, and the document observer. Infinite scroll keeps all loaded cards mounted. Lift one active context menu to `LibraryTab` or `ProfileContent` and pass it the selected game.

src/renderer/src/pages/library/library-game-card-large.tsx
```diff
@@ -108,7 +100,14 @@ export const LibraryGameCardLarge = memo(function LibraryGameCardLarge({
 
   useEffect(() => {
     setHeroIndex(0);
```
Moyasee, line 125: Resetting the source index when these URLs change leaves the previous preload's `onerror` active. An old image failure can then increment the new source index and skip the new first asset. Clear the handler during cleanup and ignore callbacks from stale preload attempts.

src/renderer/src/pages/library/library.tsx
```diff
@@ -151,6 +126,42 @@ export default function Library() {
...
+    },
+    []
+  );
+
+  useEffect(() => {
+    const el = gamesScrollRef.current;
+    if (!el) return;
+
+    const handleWheel = (event: WheelEvent) => {
+      if (event.deltaY > 0) {
```
Moyasee, line 141: Header visibility is driven only by `wheel`, so keyboard, touch, and scrollbar scrolling cannot control it, while `preventDefault()` consumes the first wheel event without scrolling. The collapsed controls also remain keyboard-focusable while invisible. Derive visibility from scroll position and delta, do not cancel wheel input, and make collapsed controls inert.

src/renderer/src/pages/profile/profile-content/user-library-game-card.tsx
```diff
@@ -136,13 +181,23 @@ export function UserLibraryGameCard({
 
       await getUserLibraryGames(sortBy);
 
+      try {
+        window.dispatchEvent(
```
Moyasee: This handler already refreshes the profile library, then this event makes `ProfileContent` fetch it again and also triggers a full local-library read. Keep one owner for the refresh and remove the redundant event-driven requests.

src/renderer/src/components/game-context-menu/use-game-actions.ts
```diff
@@ -292,6 +292,32 @@ export function useGameActions(game: LibraryGame) {
...
+          })
+        );
+      } catch {
+        /* empty */
+      }
+    } catch (error) {
+      showErrorToast(t("failed_update_pinned"));
+      logger.error("Failed to toggle pin", error);
+      throw error;
+    }
```
Moyasee, line 317: The context menu invokes this async action without awaiting or catching it. Rethrowing after the toast therefore produces an unhandled rejection on IPC failure. Handle the returned promise in `ContextMenu` or stop rethrowing here.

src/renderer/src/pages/profile/profile-content/user-library-game-card.tsx
```diff
@@ -47,8 +50,36 @@ export function UserLibraryGameCard({
...
-  const [isPinning, setIsPinning] = useState(false);
   const [imageError, setImageError] = useState(false);
+  const { openCollectionContextMenu } = useCollectionContextMenu();
+  const { library } = useLibrary();
+  const [contextMenu, setContextMenu] = useState<{
+    visible: boolean;
+    position: { x: number; y: number };
+  }>({ visible: false, position: { x: 0, y: 0 } });
+
+  const contextMenuGame = useMemo<GameContextMenuGame>(() => {
```
Moyasee: Every own-profile card subscribes to the full local library and scans it with `find()`, producing `O(cards × library)` work whenever the library updates. Build one game lookup map in `LibraryTab` and pass the matching entry to each card.

src/renderer/src/pages/profile/profile-content/user-library-game-card.tsx
```diff
@@ -201,31 +256,13 @@ export function UserLibraryGameCard({
           type="button"
           className="user-library-game__cover"
           onClick={() => navigate(buildUserGameDetailsPath(game))}
+          onContextMenu={handleContextMenu}
```
Moyasee, line 189: Moving pin/unpin from the native card button into this custom context menu removes the direct keyboard-operable action. The menu does not transfer focus or implement menu keyboard navigation. Keep an explicit keyboard-accessible pin action or implement complete context-menu keyboard behavior.

Moyasee (changes requested):
The new context-menu behavior scales with every mounted sidebar and profile card, and several interaction paths regress keyboard, scrolling, and error handling. Search, category, and platform filters also need a generic no-results state instead of a blank grid. `pt-PT` is missing the new sidebar and library keys, including `filter_sort_tooltip`, `platforms_label`, `consoles_label`, `show_favorites_first`, `sort_title`, `pin_game`, `unpin_game`, and `failed_update_pinned`. The existing `sidebar-favorites-first` behavior was intentionally excluded from this review.

Moyasee:
@galinhaman22 went back over the latest head and resolved the threads that are done. Lifting the context menu out of the profile cards, the local library map, the observer guard, the modal ordering and the title guards all look good.

Still open from the previous rounds:

- `sidebar.tsx:88` still reads `!== "false"`, so favorites-first switches on for every existing user on upgrade without them ever opting in. Should be `=== "true"`.
- `library.tsx:148` header visibility is still wheel-only with `preventDefault`, so keyboard, touch and scrollbar scrolling can't toggle it and the first tick in each direction gets swallowed. Derive it from scroll position and delta instead of cancelling wheel input.
- Still no generic empty state when the search, category or platform filters match nothing, you just get a blank grid. `no_results` already exists in all five locales.
- `pt-PT` is untouched and missing all eight new keys: `filter_sort_tooltip`, `platforms_label`, `consoles_label`, `show_favorites_first`, `sort_title`, `pin_game`, `unpin_game`, `failed_update_pinned`.
- Pin/unpin is still right-click only, and the context menu has no focus transfer or arrow key navigation, so there's no keyboard path to it.

Two smaller things while you're in there:

- `heroSources` at `library-game-card-large.tsx:113` sits outside the `useMemo`, so it gets a new identity every render and the preload effect builds a fresh `Image` each time.
- `handleOnMouseEnterGameCard` / `handleOnMouseLeaveGameCard` at `library.tsx:257` are empty no-ops but still wired to every card.

Moyasee:
@galinhaman22 went over `9ca900d5`. Fair enough on the three you pushed back on, favorites-first as a default makes sense for what favorites are for, and I'll drop the profile pin keyboard path and the header scroll behaviour. One note on the header for later, not blocking: once it's hidden, dragging the scrollbar back to the top doesn't bring it back, you have to wheel up. Fine to leave.

Confirmed fixed on this head: library and sidebar no-results states, pt-PT backfill, `heroSources` memo, and the no-op card handlers are gone. Typecheck and lint are clean.

One new thing, from `a65157cd` rather than the latest commit. The `game_details` namespace now declares `game_added_to_pinned` and `game_removed_from_pinned` twice in `en`, `es`, `pt-BR` and `ru`:

```
343:    "pin_game": "Pin game",
344:    "unpin_game": "Unpin game",
345:    "game_added_to_pinned": "Game pinned",
346:    "game_removed_from_pinned": "Game unpinned",
347:    "failed_update_pinned": "Failed to update pin",
...
611:    "game_removed_from_pinned": "Game removed from pinned",
612:    "game_added_to_pinned": "Game added to pinned",
```

The later pair wins, so the new wording never renders and the toast still says "Game added to pinned". Same shape in `es` (326/443), `pt-BR` (299/480) and `ru` (349/614). `pt-PT` is the only one you got right, you added the three new keys next to the existing pair instead of redeclaring them.

Dropping the two duplicated lines from the new block in those four files is enough, `pin_game` / `unpin_game` / `failed_update_pinned` are the only ones that were actually missing.

Moyasee:
@galinhaman22 checked `80c7fac7`, the duplicate pinned keys are gone in all four files and every namespace parses clean now. `pin_game`, `unpin_game` and `failed_update_pinned` are still there in all five locales, and the toasts fall through to the pre-existing pair as intended. Typecheck passes.

Nothing else open from my side.

src/renderer/src/pages/profile/profile-content/profile-content.tsx
```diff
@@ -166,6 +166,19 @@ export function ProfileContent() {
     shops,
   ]);
 
+  useEffect(() => {
+    const handlePinToggled = () => {
+      if (userProfile) {
+        getUserLibraryGames(sortBy, true, shops);
```
Moyasee: This uses `sortBy`, but every other fetch path uses `effectiveSortBy` (`:142`, `:158`), which falls back to `playedRecently` when the profile has no active subscription.

`sortBy` also survives a user switch, the reset effect below only clears reviews, tab and platform. So: pick "Achievements" on a subscriber's profile, navigate to your own profile without a sub, pin a game, and the list comes back sorted by achievement count while the dropdown still reads "Recently played".

`effectiveSortBy` here and in the dep array.

src/renderer/src/helpers.ts
```diff
@@ -329,3 +358,118 @@ export const getClassicsLaunchErrorSystem = (
...
+      case "title_asc":
+      default:
+        break;
+    }
+
+    return compareLibraryGamesByTitle(a, b);
+  });
+};
+
+export const getGameCollectionIds = (game: LibraryGame): string[] => {
```
Moyasee: This one has no importers. `library.tsx` doesn't import anything from helpers, it still has its own copy at `:68`, plus there's another in `game-context-menu.tsx:45` and a third in `big-picture/library-data.ts:85`.

Same story for the other two extractions: the sidebar picked up `sortLibraryGames` and `filterLibraryGamesByCategory`, but `library.tsx` kept its own inlined sort switch (`:309`) and category filter (`:380`). So the shared helpers landed without the duplication they were meant to remove, and this export is dead.

Happy for it to be a follow-up, just don't want it to quietly stay this way.

src/renderer/src/pages/library/library-game-card.tsx
```diff
@@ -111,14 +93,8 @@ export const LibraryGameCard = memo(function LibraryGameCard({
...
 
-  const classicsSystem =
-    game.shop === "launchbox" ? platformToSystem(game.platform) : null;
-  const classicsPlatformLabel = classicsSystem
-    ? PLATFORM_LABELS[classicsSystem]
-    : null;
-  const classicsEmulatorIcon = classicsSystem
-    ? EMULATOR_ICONS[SYSTEM_TO_BINARY[classicsSystem]]
-    : undefined;
+  const { label: classicsPlatformLabel, emulatorIcon: classicsEmulatorIcon } =
```
Moyasee, line 96: This dropped the shop guard. It used to be `game.shop === "launchbox" ? platformToSystem(game.platform) : null`, now it runs for every game and the badge below renders on `classicsPlatformLabel` alone. Same in `library-game-card-large.tsx:203`.

Any game whose `platform` matches `/playstation|ps1|ps2|ps3/` now gets a PS badge and an emulator icon regardless of shop. `merge-with-remote-games.ts:209` writes `platform` from the API for every shop, so it's reachable, not just theoretical.

Either pass the shop into `getClassicsPlatformDetails` or keep the check at the call site.

Moyasee (commented):
Went back over the full diff on `80c7fac7` rather than just the last commit, and re-checked every open thread from the earlier rounds.

Everything from my previous rounds is either fixed or something you've pushed back on with reasoning I accept. One correction to my own earlier comment on the header: the collapsed controls do leave the tab order, `library.scss:83` sets `visibility: hidden` on the delayed transition, so ignore that half of it.

Three new things below. None of them block, but the first one is a one-word fix and I'd take it before merge.

Still outstanding outside the code: chubby's `CHANGES_REQUESTED` is at the review level, so it needs a re-review from them regardless of what I do, and the Windows build hasn't finished.

Moyasee:
@galinhaman22 checked `770f7dff`. All three are handled:

- `profile-content.tsx:172` now uses `effectiveSortBy`, dep array too.
- `getGameCollectionIds` is down to one copy, `library.tsx` and `game-context-menu.tsx` both import it now. The structural param type is a nicer fit than the old `LibraryGame` one.
- `getClassicsPlatformDetails` takes the shop and gates on `launchbox` internally, both cards pass it.

Typecheck and lint clean locally, Sonar and lint green on CI.

One leftover from that middle point, purely a follow-up: `library.tsx:299` still has its own inlined sort switch rather than calling `sortLibraryGames`. I walked through both and they're behaviourally identical, tie-break on title included, so it's a straight swap whenever you feel like it. The category filter right below it I'd leave alone, the extra `selectedPlatform` branch means it isn't a drop-in for `filterLibraryGamesByCategory`.

Nothing blocking from me.

Moyasee:
@galinhaman22 went over the four new commits on `d7558309`.

`2848eadd` is what I asked for, `sortedLibrary` is down to the helper call and the behaviour is unchanged.

Checked the new achievement sort math since it's easy to get backwards. `bUnlocked * aTotal - aUnlocked * bTotal` is the right cross-multiplied form for descending completion rate, no float involved, and the zero-total branch puts games without achievements last in both argument orders. Also confirmed the data is actually there, `get-library.ts:65-121` populates both `unlockedAchievementCount` and `achievementCount` on every entry, so the sort isn't going to silently rank half the library at 0%. Locales are complete for `sort_achievements` and no duplicate keys anywhere.

The profile card fade fix reads right, the fade now tracks the same condition as the bar instead of `achievementCount` alone. Library card keeps the plain `achievementCount > 0` check, which is correct there since it has no subscription gate.

Two small things, neither blocking:

- Unchecking the last selected console silently no-ops (`sidebar-filter-menu.tsx:52`). The checkbox just refuses to move with nothing explaining why. Marking that last item disabled would at least make the rule visible, and All Consoles is still there as the way out.
- `setup-shell.scss` now puts `white-space: nowrap` on every `.button` inside a setup footer and on every `__ghost-button`, which is a fair bit wider than the classics scan modal. `setup-footer.tsx`, `setup-step-bios.tsx` and `setup-step-firmware.tsx` all use those. Probably fine given how short the labels are, worth an eyeball at a narrow window.

One observation rather than a request: the library page platform filter is still single-select while the sidebar flyout is multi now. Reads deliberate from the commit message, just flagging that someone will notice.

Typecheck and lint clean locally, and CI is fully green now including both builds. Nothing blocking from me.

Moyasee:
@galinhaman22 consolidating everything still open across all three reviews on `d7558309`, so nothing gets lost in the thread history. Nothing here blocks a merge from my side.

## Still actionable

**greptile P2, magic numbers.** The only greptile point nobody has touched. `sideOffset` is `6` in `library-select.tsx:53` and `collections-filter.tsx:84`, `8` in `sidebar-filter-menu.tsx:138`, `collisionPadding` is `16` in all three, plus `ChevronDownIcon size={12}` in the two library dropdowns and the gradient stops `#16b195` / `#3e62c0` in `bottom-panel.tsx:188-189`. Named constants per the style guide.

**Last console can't be unchecked.** `sidebar-filter-menu.tsx:51`. The handler returns early when one console is left, so the checkbox silently refuses to move. Disabling that last item would make the rule visible. All Consoles is still the way out.

**`white-space: nowrap` blast radius.** `setup-shell.scss:74` and `:87` apply to every `.button` inside a setup footer and every `__ghost-button`, which is a lot more than the classics scan modal. `setup-footer.tsx`, `setup-step-bios.tsx`, `setup-step-firmware.tsx`, `setup-step-rom-folder.tsx`, `memory-card-scan-modal.tsx` and `emulator-setup-modal.tsx` all use those classes. Worth an eyeball at a narrow window.

## Settled, listing so they don't get reopened

**greptile P2, DOM coupling.** `use-close-on-library-scroll.ts:9` still does `document.querySelector(".library__games-scroll")`, but it's one hook now instead of two copies and it only attaches while the dropdown is open. Good enough, I'm not pushing for the ref prop.

**greptile P1 and my equivalent, favorites-first defaults on.** `sidebar.tsx:89`. Your call, favorites pinning to the top by default is what favorites are for.

**Header show/hide is wheel-only.** `library.tsx:145-156`. Accepted, changing it would cost the feel you're after. Withdrawing the second half of my original comment as well, `library.scss:83` already drops the collapsed controls out of the tab order.

**No keyboard path to pin/unpin on the profile.** `user-library-game-card.tsx:125`. Accepted, the game page covers it and the card stays uncluttered.

**`filterLibraryGamesByCategory` unused in the library page.** `library.tsx:318-332`. Leave it, the extra `selectedPlatform` branch means it isn't a drop-in.

**Library platform filter is single-select, sidebar is multi.** Deliberate, noting it only so it doesn't come back as a bug report.

## Not code

chubby's review is still sitting at `CHANGES_REQUESTED` even though all their line threads are resolved, so it needs a re-review from them to clear. Mine is too, I'll flip it once you've had a look at the three above. CI is fully green including both builds.

Moyasee (approved):
Re-reviewed at `60648884`. Everything from my punch list is closed out.

The magic numbers are gone from all four spots, and pulling the shared ones into `library-dropdown.ts` so `LibrarySelect` and `CollectionsFilter` can't drift apart is nicer than what I asked for. The `white-space: nowrap` scoping is right too, the `--single-line` modifier keeps it on the classics scan modal and off the other five setup screens, and dropping it from `__ghost-button` is safe since that modal doesn't render one anymore.

On the last-console thing, you're right and I'll drop it. The category and sort groups right above it are both radio groups, so clicking an already-selected item doing nothing is already how this menu behaves. Making the checkbox group jump to "All consoles" would be the inconsistent option, not the other way around.

The three you pushed back on earlier are fine as they are. Favorites first is a product call and it's your call to make, the pin action still lives on the game page, and I'd already walked back half of my header comment.

CI is green on both builds, lint, Sonar and text integrity. Approving.

chubbygrannychaser (approved):
LGTM

PR #2566, opened 2026-07-20
Remove local achievement persistence

chubbygrannychaser:
## Summary

- remove the LevelDB achievement sublevel and replace live achievement state with a process-only memory store
- reconcile achievement state through the authenticated API for both cloud and non-cloud sessions
- clear achievement memory when authentication changes and purge legacy LevelDB records during startup
- update Steam, LaunchBox, RetroAchievements, library count, reset, and notification flows to use the new boundary

## Why

The launcher still persisted unlocked achievements in LevelDB after the API moved non-cloud users to per-session achievement persistence. Those records could be parsed and transferred independently of the authenticated API session, undermining the cloud subscription boundary.

With this change, LevelDB no longer contains individual achievement records. The API is the persistence authority, while the launcher keeps only the temporary working state needed for catalogue rendering, file reconciliation, and notifications.

## User impact

Non-cloud users retain achievements within their authenticated API session. Cloud subscribers continue to reconcile account-wide achievements. Existing local LevelDB achievement records are removed on the first startup after updating.

API reconciliation is attempted whenever a game has a remote ID and remains retryable after transient failures.

## Validation

- `yarn test` — 32 tests passed
- `yarn typecheck`
- `yarn format-check`
- focused ESLint on all changed files — no errors


chubbygrannychaser:
@greptileai please re-review, addressed both P1 (startup PUT flood) and P2 (fire-and-forget session reset ordering).

PR #2567, opened 2026-07-20
feat: make achievement notification windows ephemeral

chubbygrannychaser:
## What changed

- Replace the startup-long Windows achievement notification `BrowserWindow` with an ephemeral presenter-owned window.
- Create the hidden, non-focusable, click-through host only when the queue becomes non-empty.
- Prepare renderer content before calling `showInactive()`, reuse the host for the active burst, and destroy it when the queue drains.
- Move queue ownership and prepare/start/finished state coordination into the main process with sender and request ID validation.
- Add readiness/completion watchdogs, focus/crash handling, bundled image fallback, and native OS fallback behavior.
- Route normal, combined, and test notifications through the presenter.
- Dispose active notification work when preferences or position change, and forward theme changes only to an existing host.
- Keep Linux in-app notifications and macOS behavior unchanged.

## Why

Hydra previously retained a transparent achievement window for the entire application session. In some display or rendering conditions, that dormant native overlay could remain visible as a black rectangle. The persistent host also made focus preservation more fragile while a game was active.

The new lifecycle removes the native overlay entirely while idle and ensures playback starts only after a non-focusable window is ready and visible.

## User impact

Windows users should no longer have a dormant notification window capable of leaving a black rectangle. Achievement bursts still render sequentially, while keyboard and controller focus should remain with the game.

## Validation

- `npm run typecheck`
- `npm test` — 51 tests passed
- Full pre-push ESLint — 0 errors
- Presenter unit coverage for lifecycle, queue reuse, stale acknowledgements, watchdogs, failures, and disposal

## Manual follow-up

Windows validation is still required for focused games, borderless/exclusive fullscreen, taskbar and Alt-Tab visibility, rapid achievements, theme variants, and confirming that no notification window remains after the queue drains.


chubbygrannychaser:
@greptileai please re-review, addressed both P1 and P2 comments.

PR #2573, opened 2026-07-22
Fix Classic game shortcut support

Moyasee:
## What changed

- show Desktop, Start Menu, and Steam shortcut actions for Classic games
- route Classic shortcuts through Hydra's `hydralauncher://run` deep link
- launch the selected or first available Classic disc with its configured emulator
- support checking and deleting Classic Steam shortcuts
- share shortcut argument construction between native and Steam shortcuts

## Why

Classic games do not have an executable path because Hydra launches their discs through a configured emulator. The shortcut UI and Steam shortcut handlers required an executable path, while the shared deep-link handler only supported executable-based games. As a result, desktop options were hidden and Big Picture shortcuts could be created but could not launch Classic games.

## Impact

Classic games can now be launched from Desktop, Windows Start Menu, or Steam through Hydra, including when Hydra is not already running.

## Validation

- `npm run typecheck:node`
- ESLint on all changed files
- `npm test` (53 tests passed)
- `git diff --check`

Linear: LBX-955


Moyasee:
@greptile review

src/main/index.ts
```diff
@@ -207,6 +207,28 @@ const handleRunGame = async (shop: GameShop, objectId: string) => {
...
+    if (!system || !discPath) {
+      logger.error("Classic game has no launchable disc", {
+        shop,
+        objectId,
+        platform: game.platform,
+      });
+      return;
+    }
+
+    await launchClassicsGame({ shop, objectId, discPath, system });
```
chubbygrannychaser: `openClassicsGame` validates that the selected disc still exists and, for PS3, refuses/handles a running RPCS3 instance before launch. This deep-link path calls `launchClassicsGame` directly, so a shortcut can skip both checks: a stale disc is passed to the emulator, and a second RPCS3 spawn is attempted while one is already running. Reuse or extract the existing validation/launch flow here so UI launches and shortcut launches have the same safety behavior.

chubbygrannychaser (dismissed):
The Classic shortcut path bypasses the validation and RPCS3 process guard in `open-classics-game.ts`. Please route this through the same shared launch path (or extract that validation) before calling `launchClassicsGame`.

Hachi-R (changes requested):
## Requested changes

### [P1] Use a stable Hydra path for Steam shortcuts

`src/main/events/library/create-steam-shortcut.ts:214`

Classic shortcuts persist `process.execPath` as their executable. In Hydra's supported AppImage build, that points inside a temporary AppImage mount, so the shortcut cannot start Hydra after that mount disappears. Portable Windows builds have the same stable-launcher concern. Resolve the actual launcher using `APPIMAGE` or `PORTABLE_EXECUTABLE_FILE` before falling back to `process.execPath`.

References:
- https://docs.appimage.org/reference/architecture.html
- https://www.electron.build/nsis/#portable

### [P2] Reject Classic shortcuts when no disc is available

`src/renderer/src/pages/game-details/modals/game-options-modal/general-section.tsx:765`

The new condition exposes every shortcut action for any LaunchBox entry, including games whose `discs` list is empty. Both creation handlers still report success, but launching that shortcut reaches `openClassicsGame` and throws `NO_DISC`. Please gate these actions on an available disc or validate before creating the shortcut.

The extraction of `openClassicsGame` resolves the earlier validation/RPCS3 bypass, and the game-specific Steam identity changes look sound. No tests were added for the new deep-link launch, shortcut matching, or packaging-specific executable resolution.

Hachi-R (changes requested):
## Requested changes

The previous AppImage/portable-path and no-disc-creation findings are fixed. Two issues remain:

### [P2] Keep Steam deletion available after the last disc is removed

`src/renderer/src/pages/game-details/modals/game-options-modal/general-section.tsx:769`

The new guard hides the entire shortcut section, including the Delete Steam Shortcut button. If a user creates a shortcut and subsequently removes the game's last disc, `steamShortcutExists` remains true but Hydra no longer exposes a way to delete it. Big Picture has the same regression at `src/big-picture/src/components/pages/game/game-settings-modal/launch-tab.tsx:555`. Please keep the section visible when a Steam shortcut exists, or gate only the creation controls.

### [P2] Add Snap support to the stable executable resolver

`src/main/helpers/hydra-executable-path.ts:1`

Hydra also ships a Snap target, but Snap sets neither `APPIMAGE` nor `PORTABLE_EXECUTABLE_FILE`, so this falls back to `process.execPath`. That resolves to the payload beneath a revision-specific `/snap/<name>/<revision>` mount and can become stale after refresh; the stable entry point is the command beneath `/snap/bin`. Please account for `SNAP_INSTANCE_NAME`/`SNAP_APP_NAME` when constructing the launcher.

References:
- https://nodejs.org/api/process.html#processexecpath
- https://snapcraft.io/docs/reference/development/yaml-schemas/the-snap-format/

Lint and text-integrity checks pass. No feature tests were added for these shortcut flows.

Hachi-R (changes requested):
## Requested changes

The previous findings are fixed: deleting an existing shortcut remains available without discs, and Snap now uses a stable launcher path on recent snapd. One compatibility issue remains.

### [P2] Do not require `SNAP_APP_NAME` to recognize Snap installs

`src/main/helpers/hydra-executable-path.ts:2`

`SNAP_APP_NAME` is only available starting with snapd 2.77. On older supported installations it is absent, so `getSnapExecutablePath()` returns `null` and falls back to the revision-specific `process.execPath`, recreating the stale-shortcut problem after a Snap refresh. Hydra's generated Snap uses the same application and Snap name, so `SNAP_NAME` can serve as the fallback.

Reference: https://snapcraft.io/docs/reference/development/environment-variables/#snap-app-name

Build, lint, and text-integrity checks pass. No feature tests were added for these shortcut flows.

Hachi-R (approved):
Approved after re-reviewing commit `03bbec42`.

All previously requested changes are resolved:
- Classic deep links reuse the validated launch path.
- AppImage, portable Windows, and Snap use stable Hydra launchers, including the older-snapd fallback.
- Shortcut creation requires an available Classic disc.
- Existing shortcuts remain deletable after all discs are removed.

No blocking findings remain. Build, lint, text-integrity, and SonarCloud checks all pass.

PR #2574, opened 2026-07-22
fix: make download speed chart span the full chart area

Moyasee:

**Fill in the PR content:**

The speed chart on the active download never fills the whole chart area. On wide windows the bars stop somewhere in the middle and stay there, leaving a big empty strip on the right side no matter how long the download runs.

Two things caused it:

- The number of bars is derived from the container width (`floor((width + gap) / (barWidth + gap))`), but the speed history in `download-slice` was hard capped at 120 samples. Anything wider than ~1.7k px asks for more bars than the history can ever provide, so the tail never fills.
- The bars are drawn left anchored, so any shortfall (the cap above, or just the first couple of minutes of a download) shows up as dead space on the right edge, which is the part people actually look at.

What changed:

- Added a shared `MAX_DOWNLOAD_SPEED_HISTORY` constant (300) so the slice and the chart agree on how much history exists, and clamped the bar count to it.
- Spread the bar spacing across the full width so the last bar lands exactly on the right edge instead of leaving the truncation remainder unused.
- Pinned the newest sample to the right edge, so the chart streams right to left and the empty placeholders sit on the left while the history warms up.
- Sized the canvas from its real client height instead of a hardcoded 100px (the CSS box is 80px, so the drawing was being squashed) and skipped drawing when the canvas has no size yet.

### DIFF
--- a/src/renderer/src/constants.ts
+++ b/src/renderer/src/constants.ts
@@ -18,4 +18,6 @@ export const DOWNLOADER_NAME = {
 
 export const MAX_MINUTES_TO_SHOW_IN_PLAYTIME = 120;
 
+export const MAX_DOWNLOAD_SPEED_HISTORY = 300;
+
 export const THEME_WEB_STORE_URL = "https://hydrathemes.shop";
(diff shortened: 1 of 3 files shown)


Hachi-R (approved):
Reviewed the current head. The shared history limit, right-aligned samples, responsive canvas sizing, and full-width bar spacing are consistent, and the checks are green. No blocking issues found.

PR #2575, opened 2026-07-22
fix: support libtorrent 2.1 in the python rpc

Moyasee:

**Fill in the PR content:**

Every download fails in dev on machines whose system Python has libtorrent 2.1:

```
ERROR hydra.rpc: Unhandled RPC error: 'unknown name in torrent params: url'
  File "python_rpc/torrent_downloader.py", line 232, in start_download
    self.torrent_handle = self.session.add_torrent(params)
KeyError: 'unknown name in torrent params: url'
```

Packaged builds are not affected: CI freezes the RPC with libtorrent 2.0 from PyPI, which still exposes the deprecated fields. Dev spawns the system `python3` instead (`src/main/services/python-rpc.ts`), so on distros shipping 2.1 (Arch/CachyOS, among others) it hits a build compiled without the deprecated ABI, where `add_torrent_params::url` no longer exists.

Three more calls in the same file are gone in 2.1 for the same reason. They were latent behind the first failure, except the rate limit one, which was already failing silently inside a bare `except`:

- `torrent_handle.get_torrent_info()` -> `torrent_file()`
- `session.set_download_rate_limit()` -> `apply_settings({"download_rate_limit": ...})`
- `session.abort()` -> removed with no replacement

**What changed**

- Build the params through `lt.parse_magnet_uri()` and set `save_path` / `flags` on the returned `add_torrent_params`, instead of the `url` dict key. The trackers the magnet already carries are kept and ours are appended without duplicates, which the old `trackers` key would have overwritten.
- Read metadata through `torrent_file()` when present, falling back to `get_torrent_info()`, and treat a missing `torrent_info` as `metadata_incomplete` rather than relying on a `RuntimeError`.
- Apply the download limit via `apply_settings`, keeping the legacy setter as a fallback.
- Guard `session.abort()` behind a `getattr` check.

An invalid magnet now surfaces as `invalid_magnet` instead of `internal_error`, since `parse_magnet_uri` raises where the old code deferred to `add_torrent`.

**Compatibility**

Works on 1.2 through 2.1. Every API used here landed in 1.2 at the latest, which is the floor the file already had through `lt.torrent_flags`, `set_flags`/`unset_flags` and `session.delete_partfile`, so the supported range is unchanged apart from gaining 2.1.

**Testing**

- Ran the downloader against libtorrent 2.1.0 directly: `add_torrent` succeeds, the tracker list merges as expected, flags and `save_path` land on the handle, the status payload is unchanged, and `metadata_timeout` / `invalid_magnet` / cancel / abort all behave.
- Started a download from the app in dev on 2.1.0 and it went through with no RPC error.
- Could not exercise 2.0 locally, so that side rests on the API review above plus CI.

### DIFF
--- a/python_rpc/torrent_downloader.py
+++ b/python_rpc/torrent_downloader.py
@@ -141,7 +141,7 @@ class TorrentDownloader:
             raise ValueError("invalid_magnet") from error
 
         params.save_path = save_path
-        params.flags = flags
+        params.flags = params.flags | flags
 
         trackers = list(params.trackers)
         known_trackers = set(trackers)
--- a/python_rpc/torrent_downloader.py
+++ b/python_rpc/torrent_downloader.py
@@ -146,15 +146,21 @@ class TorrentDownloader:
         trackers = list(params.trackers)
         known_trackers = set(trackers)
 
+        tiers = list(params.tracker_tiers)[: len(trackers)]
+        tiers.extend([0] * (len(trackers) - len(tiers)))
+
+        fallback_tier = max(tiers) + 1 if tiers else 0
+
         for tracker in self.trackers:
             if tracker in known_trackers:
                 continue
 
             trackers.append(tracker)
             known_trackers.add(tracker)
+            tiers.append(fallback_tier)
 
         params.trackers = trackers
-        params.tracker_tiers = [0] * len(trackers)
+        params.tracker_tiers = tiers
 
         return params
 
--- a/python_rpc/torrent_downloader.py
+++ b/python_rpc/torrent_downloader.py
@@ -122,9 +122,57 @@ class TorrentDownloader:
             max_download_speed if max_download_speed and max_download_speed > 0 else 0
         )
         try:
-            self.session.set_download_rate_limit(download_limit)
+            self.session.apply_settings({"download_rate_limit": download_limit})
+            return
         except Exception:
             pass
+
+        legacy_setter = getattr(self.session, "set_download_rate_limit", None)
+        if callable(legacy_setter):
+            try:
+                legacy_setter(download_limit)
+            except Exception:
+                pass
+
+    def _build_add_torrent_params(self, magnet: str, save_path: str, flags):
+        try:
+            params = lt.parse_magnet_uri(magnet)
+        except Exception as error:
+            raise ValueError("invalid_magnet") from error
+
+        params.save_path = save_path
+        params.flags = flags
+
+        trackers = list(params.trackers)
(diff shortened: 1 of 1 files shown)


Moyasee:
Good catch. `parse_magnet_uri` gives each `tr=` its own ascending tier (verified on 2.1.0: three trackers come back as tiers 0, 1, 2), so zeroing them collapsed the magnet's whole fallback ladder into tier 0.

Now the parsed tiers are kept as-is and our list appends at `max + 1`, one fallback tier below whatever the magnet declared. Padding covers libtorrent leaving `tracker_tiers` shorter than `trackers`. Magnets with no `tr=` behave exactly as before, everything at tier 0.

Moyasee:
@greptile review

Hachi-R (changes requested):
Please preserve the flags produced by `parse_magnet_uri`. `_build_add_torrent_params` immediately replaces them with `params.flags = flags`, which drops magnet-specific bits such as `default_dont_download` for `so=` selections and `i2p_torrent` for I2P trackers. Without `default_dont_download`, files beyond the explicit priority list fall back to normal priority and may be downloaded even though the magnet excluded them. Merge the requested Hydra flags with the parser-produced flags instead of overwriting them.

Moyasee:
You're right, checked it on 2.1.0 and both bits are really there:

- `&so=0,2-4` -> `0x8020b8`, so `default_dont_download` plus `file_priorities=[4, 0, 4, 4, 4]`. Overwriting the flags kept the priorities but dropped the bit that makes them mean anything, so the excluded files would have downloaded.
- a `.i2p` tracker -> `0x10020b8`, the i2p bit, which isn't even exposed in the python `torrent_flags` enum.

Now merged instead of replaced (`params.flags = params.flags | flags`). Verified the `so=` case for both the plain and the selective download path, the i2p bit, and that the seeding path keeps `upload_mode`.

Heads up on one side effect: the merge also brings in `update_subscribe` and `apply_ip_filter` from the parser defaults, which the old `params["flags"]` dict dropped. Harmless here since we poll `status()` directly and no session IP filter is set, but it's a change from current behavior rather than a restore. Happy to mask it down to just the magnet-specific bits if you'd rather keep that narrow.

Hachi-R (approved):
Re-reviewed the current head. The previous request is resolved: magnet-derived flags are now preserved while Hydra flags are merged in, keeping `so=` selection behavior and I2P mode intact. The remaining compatibility changes are consistent, and the checks are green. No blocking issues found.

PR #2583, opened 2026-07-23
feat: add similar games recommendations

BITO21:
When submitting this pull request, I confirm the following (please check the boxes):

Fill in the PR content:

### What

Adds a “More like this” section to Desktop and Big Picture game pages and improves controller navigation in Big Picture menus.

Desktop displays three games at a time, while Big Picture provides a carousel with up to nine recommendations.

### Why

Game pages currently do not offer a way to discover similar games.

Big Picture controller users also need consistent directional navigation and primary-action support in the sidebar and context menus.

### The implementation

- Fetches server-ranked recommendations from `/catalogue/{shop}/{objectId}/similar`.
- Supplies the configured download sources so availability badges remain accurate.
- Excludes the currently opened game and malformed results.
- Keeps regular and classic recommendations separate by shop.
- Lets platform metadata influence LaunchBox ranking without guaranteeing the same console.
- Reuses the existing Big Picture carousel and the profile’s vertical cover presentation.
- Normalizes LaunchBox genre labels by trimming values, accepting name or description fields, and removing case-insensitive duplicates.
- Activates context-menu items, sidebar routes, the exit action, and library filters through the controller’s primary action.
- Connects every Big Picture sidebar item with explicit up and down navigation.

### How to test

1. Open a regular game and confirm “More like this” appears below its description.
2. Navigate through the Desktop carousel and open a recommendation.
3. Open a classic game and confirm relevant LaunchBox recommendations appear.
4. In Big Picture, verify the recommendations carousel behaves like the Home and Profile carousels with mouse, keyboard, and controller.
5. Navigate through every sidebar route with the controller, including profile, exit, and library filters.
6. Activate a sidebar route, library filter, exit action, and context-menu item using the controller’s primary button.
7. Move from a recommendation to reviews, click a carousel arrow, and confirm the carousel does not snap back unexpectedly.

### DIFF
--- a/src/big-picture/src/locales/en/translation.json
+++ b/src/big-picture/src/locales/en/translation.json
@@ -735,6 +735,8 @@
     "download_sources_removed_other": "{{count}} download sources were removed",
     "download_sources_count_one": "{{count}} download source",
     "download_sources_count_other": "{{count}} download sources",
+    "similar_games": "More like this",
+    "no_similar_games": "No similar games found",
     "game_review_played_for": "after playing for {{time}}",
     "game_review_rating_only": "rating",
     "remove_from_library": "Remove from Library",
--- a/src/big-picture/src/pages/game/game.tsx
+++ b/src/big-picture/src/pages/game/game.tsx
@@ -6,16 +6,29 @@ import {
   getSkuRegionFlag,
   type SkuRegion,
 } from "@renderer/helpers";
-import type { GameShop } from "@types";
+import type { GameShop, ShopAssets } from "@types";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useTranslation } from "react-i18next";
 import { useNavigate, useParams } from "react-router-dom";
-import { buildLibraryToastOptions, getItemFocusTarget } from "../../helpers";
+import {
+  buildLibraryToastOptions,
+  getBigPictureGameDetailsPath,
+  getItemFocusTarget,
+} from "../../helpers";
+import {
+  getSimilarGamesSectionState,
+  SIMILAR_GAMES_LIMIT,
+} from "@renderer/hooks/similar-games";
+import { useSimilarGames } from "@renderer/hooks/use-similar-games";
 import type { LibraryToastSource } from "../../helpers/library-toast";
 import {
   Typography,
   VerticalFocusGroup,
   Divider,
   FocusItem,
+  FocusCarousel,
+  Skeleton,
+  EmptyState,
 } from "../../components";
 import {
   ConfirmationModal,
@@ -55,6 +68,7 @@ import {
   GAME_HERO_ACTIONS_REGION_ID,
   GAME_MEDIA_CAROUSEL_REGION_ID,
   GAME_PAGE_REGION_ID,
+  GAME_SIMILAR_GAMES_REGION_ID,
   GAME_SIDEBAR_ACHIEVEMENTS_ID,
   GAME_SIDEBAR_CONTROLLER_SUPPORT_ID,
   GAME_SIDEBAR_HLTB_ID,
@@ -64,9 +78,11 @@ import {
   GAME_SIDEBAR_REGION_ID,
   GAME_SIDEBAR_REQUIREMENTS_ID,
   GAME_SIDEBAR_STATS_ID,
+  getGameSimilarGameItemId,
 } from "../../components/pages/game/navigation";
 import { NavigationService, type FocusOverrideTarget } from "../../services";
 import { useNavigationStore } from "../../stores";
+import { extractGenreNames } from "./game-metadata";
 import "./game.scss";
 
 const DESCRIPTION_HEADING_SELECTOR = "h1, h2, h3, h4, h5, h6";
@@ -99,7 +115,6 @@ const DESCRIPTION_SCROLL_EDGE_TOLERANCE = 4;
 const DESCRIPTION_FOCUS_ENTRY_MARGIN = 32;
 const DESCRIPTION_SCROLL_ANIMATION_DURATION = 220;
 const DESCRIPTION_RETURN_MIN_VISIBLE_RATIO = 0.5;
-
 type ClassicsLaunchErrorCode = NonNullable<
   ReturnType<typeof getClassicsLaunchErrorCode>
 >;
@@ -415,7 +430,69 @@ function buildDescriptionSections(document: Document | null) {
   return sections;
 }
 
+function SimilarGamesSkeleton({ title }: Readonly<{ title: string }>) {
+  return (
+    <section
+      className="focus-carousel game-page__similar-games-skeleton"
+      data-card-variant="vertical"
+      aria-label={title}
+      aria-busy="true"
+    >
+      <div className="focus-carousel__header">
+        <h2 className="focus-carousel__title">{title}</h2>
+        <div
+          className="game-page__similar-games-skeleton-actions"
+          aria-hidden="true"
+        >
+          <Skeleton />
+          <Skeleton />
+        </div>
+      </div>
+
+      <div className="focus-carousel__viewport-wrapper">
+        <div className="focus-carousel__viewport">
+          <div className="focus-carousel__container">
+            {Array.from({ length: SIMILAR_GAMES_LIMIT }, (_, index) => (
+              <article
+                className="focus-carousel__slide"
+                key={`similar-game-skeleton-${index}`}
+              >
+                <div className="game-page__similar-games-skeleton-card">
+                  <Skeleton className="game-page__similar-games-skeleton-cover" />
+                  <div className="game-page__similar-games-skeleton-body">
+                    <Skeleton className="game-page__similar-games-skeleton-title" />
+                    <Skeleton className="game-page__similar-games-skeleton-subtitle" />
+                  </div>
+                </div>
+              </article>
+            ))}
+          </div>
+        </div>
+      </div>
+    </section>
+  );
+}
+
+function SimilarGamesEmptyState({
+  title,
+  message,
+}: Readonly<{ title: string; message: string }>) {
+  return (
+    <section className="focus-carousel game-page__similar-games-empty">
+      <div className="focus-carousel__header">
+        <h2 className="focus-carousel__title">{title}</h2>
+      </div>
+
+      <EmptyState
+        className="game-page__similar-games-empty-state"
+        title={message}
+      />
+    </section>
+  );
+}
+
 export default function Game() {
+  const { t } = useTranslation("big_picture");
   const { showErrorToast, showSuccessToast } = useBigPictureToast();
   const { shop, objectId } = useParams<{ shop: GameShop; objectId: string }>();
   const navigate = useNavigate();
@@ -548,19 +625,40 @@ export default function Game() {
   const developer = shopDetails?.developers?.[0] ?? "";
   const publisher = shopDetails?.publishers?.[0] ?? "";
   const releaseDate = shopDetails?.release_date?.date ?? "";
-  const launchboxGenres = useMemo(() => {
-    return ((shopDetails?.genres ?? []) as unknown[])
-      .map((genre) => {
-        if (typeof genre === "string") return genre;
-        if (genre && typeof genre === "object" && "name" in genre) {
-          const { name } = genre as { name?: unknown };
-          return typeof name === "string" ? name : "";
-        }
-
-        return "";
-      })
-      .filter((genre) => genre.trim().length > 0);
+  const shopGenres = useMemo(() => {
+    return extractGenreNames(shopDetails?.genres ?? []);
   }, [shopDetails?.genres]);
+  const {
+    games: similarGames,
+    isLoading: areSimilarGamesLoading,
+    isEligible: areSimilarGamesEligible,
+  } = useSimilarGames({
+    objectId: objectId ?? "",
+    shop: shop ?? "custom",
+  });
+  const similarCarouselGames = useMemo<ShopAssets[]>(
+    () =>
+      similarGames.map((similarGame) => ({
+        objectId: similarGame.objectId,
+        shop: similarGame.shop,
+        title: similarGame.title,
+        iconUrl: similarGame.iconUrl,
+        libraryHeroImageUrl: similarGame.libraryHeroImageUrl,
+        libraryImageUrl: similarGame.libraryImageUrl,
+        logoImageUrl: similarGame.logoImageUrl,
+        logoPosition: null,
+        coverImageUrl: similarGame.coverImageUrl,
+        downloadSources: similarGame.downloadSources,
+      })),
+    [similarGames]
+  );
+  const similarGamesSectionState = getSimilarGamesSectionState(
+    areSimilarGamesEligible,
+    areSimilarGamesLoading,
+    similarCarouselGames.length
+  );
+  const hasSimilarGames = similarGamesSectionState === "ready";
+  const shouldRenderSimilarGames = similarGamesSectionState !== "hidden";
   const launchboxRegions = useMemo(
     () =>
       shopDetails?.skus && shopDetails.skus.length > 0
@@ -592,6 +690,16 @@ export default function Game() {
       preferRememberedFocus: true,
     };
   }, [hasNavigableComments]);
+  const similarGamesEntryTarget = useMemo(() => {
+    if (!hasSimilarGames) return undefined;
+
+    return {
+      type: "region" as const,
+      regionId: GAME_SIMILAR_GAMES_REGION_ID,
+      entryDirection: "down" as const,
+      preferRememberedFocus: true,
+    };
+  }, [hasSimilarGames]);
   const bodyUpNavigationTarget = useMemo<FocusOverrideTarget>(() => {
     if (activeMediaItemId) {
       return getItemFocusTarget(activeMediaItemId);
@@ -635,12 +743,15 @@ export default function Game() {
       };
     }
 
-    return descriptionEntryTarget ?? commentsEntryTarget;
+    return (
+      descriptionEntryTarget ?? similarGamesEntryTarget ?? commentsEntryTarget
+    );
   }, [
     activeMediaItemId,
     commentsEntryTarget,
     descriptionEntryTarget,
     hasMedia,
+    similarGamesEntryTarget,
   ]);
   const sidebarEntryTarget = useMemo(
     () => sidebarStatsEntryTarget,
@@ -656,6 +767,13 @@ export default function Game() {
     []
   );
   const commentsTopNavigationTarget = useMemo(() => {
+    if (similarGamesEntryTarget) {
+      return {
+        ...similarGamesEntryTarget,
+        entryDirection: "up" as const,
+      };
+    }
+
     if (descriptionBottomEntryTarget) {
       return descriptionBottomEntryTarget;
     }
@@ -670,7 +788,53 @@ export default function Game() {
     descriptionBottomEntryTarget,
     hasMedia,
     heroActionsLeftNavigationTarget,
+    similarGamesEntryTarget,
   ]);
+
+  const similarGamesTopNavigationTarget = useMemo<FocusOverrideTarget>(() => {
+    if (descriptionBottomEntryTarget) return descriptionBottomEntryTarget;
+    if (activeMediaItemId) return getItemFocusTarget(activeMediaItemId);
+
+    if (hasMedia) {
+      return {
+        type: "region",
+        regionId: GAME_MEDIA_CAROUSEL_REGION_ID,
+        entryDirection: "up",
+        preferRememberedFocus: true,
+      };
+    }
+
+    return heroActionsLeftNavigationTarget;
+  }, [
+    activeMediaItemId,
+    descriptionBottomEntryTarget,
+    hasMedia,
+    heroActionsLeftNavigationTarget,
+  ]);
+
+  const getSimilarGameNavigationOverrides = useCallback(
+    (_similarGame: ShopAssets, index: number, games: ShopAssets[]) => ({
+      up: similarGamesTopNavigationTarget,
+      down: commentsEntryTarget ?? ({ type: "block" } as const),
+      ...(index === 0
+        ? { left: getItemFocusTarget(BIG_PICTURE_SIDEBAR_ITEM_IDS.home) }
+        : {}),
+      ...(index === games.length - 1
+        ? { right: { type: "block" as const } }
+        : {}),
+    }),
+    [commentsEntryTarget, similarGamesTopNavigationTarget]
+  );
+  const getSimilarGameItemId = useCallback(
+    (similarGame: ShopAssets) =>
+      getGameSimilarGameItemId(
+        shop!,
+        objectId!,
+        similarGame.shop,
+        similarGame.objectId
+      ),
+    [objectId, shop]
+  );
   useHeaderTitle(resolvedGameTitle);
 
   const handleOpenDownloadModal = useCallback(() => {
@@ -1046,7 +1210,9 @@ export default function Game() {
                 return;
               }
 
-              focusNavigationTarget(commentsEntryTarget);
+              focusNavigationTarget(
+                similarGamesEntryTarget ?? commentsEntryTarget
+              );
             },
             left: () => {
               navigation.setFocus(BIG_PICTURE_SIDEBAR_ITEM_IDS.home);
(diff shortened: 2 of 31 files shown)


antoinebou12:
Super cool features I want that as a user

I would add a like a better recommendations system using the user data. 

Also, You can use embedding model and the JSON data and kmeas algo more complicated but give good results for recommendations use in sementatic search. You can add a user vector based on click or download other user stats to improve the search 

If you want to keep you can add a Jaccard similarity pour improve current algo weight based on total unique genre

Moyasee (commented):
Checked out the branch and ran the suite: typecheck, eslint and prettier are clean, 84/84 tests pass. The endpoint this depends on is live on prod and staging for both steam and launchbox, so the feature works end to end.

Two things need a decision before merge, everything else is inline.

The description no longer matches the code. It says the implementation ranks by shared genres, excludes the current game and matches classics to the same console. The squashed commit hands all of that to the server, and the console match in particular is not happening:

```
GET /catalogue/launchbox/976/similar   (Crash Bandicoot, PS1)
launchbox 10001 The Last of Us
launchbox 67    Uncharted: Drake's Fortune
launchbox 98824 God of War HD
```

The response has no platform field, so the client could not filter on it either. "How to test" step 3 fails as written. Either restore the console filter or rewrite the description and the test steps.

The other one is the response normalizer, which is all or nothing: a single malformed item drops the whole section. Inline on the file.

Maintainer call, but the profile card extraction and the focus carousel navigation changes are independent of the recommendation feature and could land as their own PR.

src/renderer/src/hooks/similar-games.ts
```diff
@@ -0,0 +1,159 @@
...
+      if (typeof id === "string") return id;
+    }
+
+    return `similar-game-source-${index}`;
+  });
+};
+
+const normalizeSimilarGame = (value: unknown): SimilarGame => {
+  if (!value || typeof value !== "object") {
+    throw new TypeError("Invalid similar games response");
```
Moyasee: `normalizeSimilarGame` throws, so one malformed entry rejects the whole promise and the section vanishes. `downloadSources` is mandatory as well: the test asserts a TypeError for an item that omits it.

The repo already normalizes this payload shape by skipping bad entries:

```ts
// src/big-picture/src/pages/home/home-data.ts:54
export function normalizeShopAssetsList(value: unknown): ShopAssets[] {
  if (!Array.isArray(value)) return [];

  return value.flatMap((game) => {
    const normalizedGame = normalizeShopAssets(game);

    return normalizedGame ? [normalizedGame] : [];
  });
}
```

Return null per item and flatMap here, and keep the throw only for a non-array response.

src/renderer/src/hooks/similar-games.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  return value.map((source, index) => {
+    if (typeof source === "string") return source;
+    if (source && typeof source === "object" && "id" in source) {
+      const id = (source as { id?: unknown }).id;
+      if (typeof id === "string") return id;
+    }
+
+    return `similar-game-source-${index}`;
```
Moyasee: `similar-game-source-${index}` invents an id that matches no real source, and it preserves the array length, which feeds `downloadSourceCount={game.downloadSources.length}` in `focus-carousel/index.tsx`. Drop unrecognized entries the way `normalizeShopAssets` does instead of substituting a placeholder.

src/renderer/src/hooks/similar-games.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  if (candidate.genres != null && !Array.isArray(candidate.genres)) {
+    throw new TypeError("Invalid similar games response");
+  }
+
+  return {
+    objectId: candidate.objectId,
+    shop: candidate.shop,
+    title: candidate.title,
+    genres: (candidate.genres ?? []).filter(
```
Moyasee: Neither the desktop nor the Big Picture section reads `SimilarGame.genres`. It only buys the throw path on line 101, which can reject an otherwise usable response. Remove the field from the interface and the normalizer.

src/renderer/src/hooks/similar-games.ts
```diff
@@ -0,0 +1,159 @@
...
+export const fetchSimilarGames = async (
+  query: SimilarGamesQuery,
+  get: SimilarGamesGet
+) => {
+  if (!isSupportedShop(query.shop) || !query.objectId) return [];
+
+  const response = await get(
+    `/catalogue/${query.shop}/${encodeURIComponent(query.objectId)}/similar`,
+    {
+      params: { take: SIMILAR_GAMES_LIMIT },
```
Moyasee: Only `take` goes out, so `downloadSources` comes back empty for every result:

```
GET /catalogue/steam/1245620/similar?take=9
9 results, downloadSources: [] on all of them
```

Big Picture cards then report zero sources for games that show badges everywhere else in the app. Every other catalogue call passes `downloadSourceIds` (`use-weekly-games.ts:14`, `use-catalogue-data.ts:502`). Worth checking whether the endpoint accepts it.

src/renderer/src/components/game-card/game-card.tsx
```diff
@@ -16,7 +16,10 @@ export interface GameCardProps
     React.ButtonHTMLAttributes<HTMLButtonElement>,
     HTMLButtonElement
   > {
-  game: ShopAssets;
+  game: Pick<
```
Moyasee: Leftover from an earlier iteration: nothing in this PR passes a partial `ShopAssets` to `GameCard`. Reverted it to `game: ShopAssets` locally and `yarn typecheck` still passes, so this can come out.

src/big-picture/src/components/common/focus-carousel/index.tsx
```diff
@@ -726,13 +743,9 @@ export function FocusCarousel({
     const targetFocusId = currentFocusedIndex === -1 ? rememberedFocusId : null;
     const focusedIndex = getFocusedIndex(targetFocusId);
 
-    if (currentFocusedIndex !== -1) {
-      lastAlignedFocusIdRef.current = null;
-      return;
-    }
+    if (currentFocusedIndex !== -1) return;
```
Moyasee: Dropping `lastAlignedFocusIdRef` leaves this effect unguarded, and `getItemId` is an inline arrow at `game.tsx:1370`, so the effect reruns on every render of the page.

Repro: focus a card in the similar games carousel, move focus down into the comments, then click the header arrows with the mouse. The scroll fires `select`, `syncControls` flips `canScrollPrev`, the rerender runs this effect with `currentFocusedIndex === -1`, and it scrolls back to the remembered card's alignment. `tabIndex={-1}` covers the focused-card case but not this one.

src/big-picture/src/components/common/focus-carousel/index.tsx
```diff
@@ -829,7 +841,9 @@ export function FocusCarousel({
                 type="button"
                 className="focus-carousel__header-button"
                 aria-label="Previous"
+                tabIndex={-1}
```
Moyasee: `tabIndex={-1}` and the mousedown `preventDefault` apply to every `FocusCarousel`, home included, not only the new section, so these controls stop being keyboard reachable app wide. Fine if Big Picture is controller and mouse only by design, but it is a global behavior change bundled into a feature PR and should at least be called out in the description.

src/renderer/src/pages/game-details/similar-games/similar-games.tsx
```diff
@@ -0,0 +1,119 @@
...
+                      }
+                    />
+                  </SimilarGamesSlide>
+                ))}
+          </div>
+
+          <button
+            type="button"
+            className="similar-games__control similar-games__control--previous"
+            aria-label="Previous"
```
Moyasee: Hardcoded English. `.cursorrules` routes user-facing strings through i18next, and screen reader users on other locales get these labels untranslated. Same for the two labels in `focus-carousel/index.tsx`.

Line 64 also puts `aria-label={t("similar_games")}` on the section while the `<h2>` right under it carries the same text, so it is announced twice. Drop the aria-label there.

src/renderer/src/components/vertical-cover-card/vertical-cover-card.scss
```diff
@@ -0,0 +1,114 @@
...
+  box-shadow: 0 8px 10px -2px rgba(0, 0, 0, 0.5);
+  cursor: pointer;
+  transition: all ease 0.2s;
+
+  &::before {
+    content: "";
+    position: absolute;
+    top: 0;
+    left: 0;
+    z-index: 3;
```
Moyasee: `z-index: 3` is new. The old `.user-library-game__cover::before` had no z-index, so `.user-library-game__overlay` (z-index 2) painted above the glare. On profile cards the playtime badge and the achievement bar now sit under it. Drop the z-index, or raise the overlay above it.

src/locales/en/translation.json
```diff
@@ -288,6 +288,7 @@
     "last_time_played": "Last played {{period}}",
     "not_played_yet": "You haven't played {{title}} yet",
     "next_suggestion": "Next suggestion",
+    "similar_games": "More like this",
```
Moyasee, line 299: Missing `es` and `ru` here, and `es`, `fr` and `ru` in `src/big-picture/src/locales` (the `format` block). Big Picture `fr` is currently one key behind `en`; this takes it to two.

Hachi-R (changes requested):
Re-reviewed the current head 829b01b5 and the existing discussion. I am requesting changes for five user-visible correctness and regression issues.

1. [P2] Same-console classic recommendations are not enforced (src/renderer/src/hooks/similar-games.ts:144). The request sends only take and the normalized response contains no platform field. The live LaunchBox response documented in the review mixes PS1 and PS3 titles, so the PR description and test step 3 do not hold. Either implement the platform constraint or explicitly change the feature scope and test plan.

2. [P2] One malformed recommendation removes the entire section (src/renderer/src/hooks/similar-games.ts:120). response.map(normalizeSimilarGame) throws for any invalid item, and the hook then replaces all results with an empty array. Normalize entries independently and skip malformed items while reserving a throw for a non-array response.

3. [P2] The shared Big Picture carousel can snap back unexpectedly (src/big-picture/src/components/common/focus-carousel/index.tsx:735). Removing the previous lastAlignedFocusIdRef guard lets the alignment effect rerun whenever an inline getItemId changes identity. After focus leaves the carousel, clicking its arrows can rerender and realign it to the remembered card. Restore an equivalent guard or stabilize the effect inputs.

4. [P2] Download-source badges are inaccurate (src/renderer/src/hooks/similar-games.ts:70). The request omits configured downloadSourceIds, yielding empty source arrays, while malformed entries are converted into invented IDs that inflate the displayed count. Pass the active source IDs and discard invalid source entries.

5. [P2] The extracted cover glare overlays profile statistics (src/renderer/src/components/vertical-cover-card/vertical-cover-card.scss:18). Its z-index 3 paints above the profile overlay at z-index 2, obscuring playtime and achievement information. Keep the glare below card overlays.

Lint and text-integrity checks are green; the build was still running at the time of review.

src/renderer/src/main.tsx
```diff
@@ -88,6 +88,7 @@ await i18n
   .init({
     resources,
     fallbackLng: "en",
+    returnEmptyString: false,
```
Moyasee: This flips i18next behavior for every string in the app, and it is here only so the 32 empty `similar_games` placeholders fall back to English.

Nothing else needs it. There are currently zero empty translation values across `src/locales` and `src/big-picture/src/locales`, and the crowd locales already rely on absent-key fallback: against `en`, `es` is missing 25 `game_details` keys, `ru` 8, `fr` 21, `de` 150.

Drop this line, drop the empty entries, and ship only the languages you can actually translate. Missing keys already fall back to English on their own.

Moyasee (commented):
Everything from the last round is addressed: the normalizer skips bad items instead of rejecting the batch, no more invented source ids, `genres` is gone, `downloadSourceIds` goes out with the request, the `GameCard` prop change is reverted, the focus carousel edits were dropped in favor of memoizing `getItemId`, the arrow labels are translated, and the glare `z-index` is out. The description and test plan now match what the code actually does.

One thing left, on the i18n approach. Inline.

Two notes that are not code:

@Hachi-R's changes-requested review was written against the previous head and is still on the PR, so it keeps blocking merge even after this round.

I could not confirm server side that `/similar` honors `downloadSourceIds`. The endpoint accepts the parameter without erroring, but with no sources configured there is no way to tell whether it filters on it. Worth checking with sources set up that the badges actually show.

Once the locale bit is sorted this is good to merge from my side.

src/big-picture/src/i18n.tsx
```diff
@@ -129,6 +129,7 @@ export async function initializeBigPictureI18n() {
       .init({
         resources: localeResources,
         fallbackLng: "en",
+        returnEmptyString: false,
```
Moyasee: Same as the renderer side: this only exists to prop up the empty placeholders, so it should come out with them.

src/locales/de/translation.json
```diff
@@ -176,6 +176,7 @@
     "deck_verified_only": "Verified only"
   },
   "game_details": {
+    "similar_games": "",
```
Moyasee: These placeholders across 32 locales are what force `returnEmptyString: false`. Leaving the crowd-maintained files untouched gives exactly the same UI, since i18next falls back to `en` for an absent key, and it keeps the translation state honest: an empty value reads as translated-to-nothing rather than not yet translated.

Recent precedent for a multi-locale string is real translations everywhere (9b324e6b1) or plain English text in the untranslated files (b0f3a0919), not empty values.

src/locales/es/translation.json
```diff
@@ -237,6 +237,7 @@
     "mode_modern_games": "Juegos modernos"
   },
   "game_details": {
+    "similar_games": "",
```
Moyasee: `es` is one of the locales maintained in-repo alongside `en`, `ru` and `pt-BR`, so this one deserves a real string rather than a placeholder.

If it helps: `es` "Juegos similares", `ru` "Похожие игры", and for Big Picture `fr` "Jeux similaires".

Hachi-R (dismissed):
[P2] Reject empty recommendation identities in src/renderer/src/hooks/similar-games.ts (lines 86-88). These checks only verify that objectId and title are strings, so a malformed entry with empty or whitespace-only values still passes normalization. It then survives the shop/current-game filter, consumes one of the nine result slots, renders a blank card, and navigates to an invalid game-details route. Require both trimmed values to be non-empty, preferably return the normalized values, and add a whitespace-only case to the malformed-response tests. The rest of the latest head looks good, and the current Build, Lint, Text Integrity, and Sonar checks are green.

antoinebou12:
Addressed the latest review and added controller navigation improvements.

Changes:
- Reject empty or whitespace-only recommendation IDs/titles and trim valid values.
- Added regression coverage for blank identities.
- Added deterministic Up/Down controller navigation across the Big Picture sidebar menu.
- Added explicit controller primary actions for sidebar routes, Exit, library filters, and context-menu items.
- Kept the latest `main` sync and previous recommendation/i18n fixes.

Validation passed: typecheck, ESLint, Prettier, and 11 recommendation tests.

Local commit: `186408ec4` (`fix similar games review and controller navigation`). The commit is ready, but this account cannot push to the PR head repository `BITO21/hydra` (read-only permission). Please push this commit from the fork owner account or grant write access so the PR head updates.

antoinebou12:
I pushed the updated work to my fork and opened fresh draft PRs against `main` because this original PR is owned by the `BITO21` fork and its head cannot be replaced:

- Similar games updates: https://github.com/hydralauncher/hydra/pull/2689
- Controller menu navigation: https://github.com/hydralauncher/hydra/pull/2690

Both branches include the latest review fixes and are ready for review.

src/renderer/src/hooks/similar-games.ts
```diff
@@ -0,0 +1,155 @@
...
+  options: SimilarGamesRequestOptions
+) => Promise<unknown>;
+
+const isSupportedShop = (
+  shop: GameShop
+): shop is Exclude<GameShop, "custom"> => {
+  return shop === "steam" || shop === "launchbox";
+};
+
+export const extractSimilarGameGenres = (genres: readonly unknown[]) => {
```
Moyasee: Once genres came out of `SimilarGame` in the last round, nothing in the similar-games path reads this. Its only consumer is now the LaunchBox Genres row at `big-picture/src/pages/game/game.tsx:509`, so a helper named for this feature exists solely to serve unrelated UI, and swapping it in for the old inline version changed what that row renders:

```
dupes differing in case:   before "Action,  action , Action"   after "Action"
object with description:   before ""                           after "Shooter"
untrimmed string:          before "  Platform  "               after "Platform"
```

The new output is better. The problem is it is an unadvertised change to an element this PR is not about. Move it next to the metadata row as something like `extractGenreNames` and mention the behavior change in the PR body.

src/big-picture/src/pages/game/game.tsx
```diff
@@ -494,19 +505,32 @@ export default function Game() {
...
+        libraryHeroImageUrl: similarGame.libraryHeroImageUrl,
+        libraryImageUrl: similarGame.libraryImageUrl,
+        logoImageUrl: similarGame.logoImageUrl,
+        logoPosition: null,
+        coverImageUrl: similarGame.coverImageUrl,
+        downloadSources: similarGame.downloadSources,
+      })),
+    [similarGames]
+  );
+  const hasSimilarGames =
```
Moyasee: `hasSimilarGames` gates both the divider and the carousel on the fetch having resolved, so in big picture they pop in after the fact and shove the reviews section down while the user is scrolling. Desktop renders skeletons for exactly this reason, worth matching.

src/renderer/src/pages/game-details/similar-games/similar-games.tsx
```diff
@@ -0,0 +1,119 @@
...
+      emblaApi.off("select", syncControls);
+      emblaApi.off("reInit", syncControls);
+    };
+  }, [emblaApi, syncControls]);
+
+  useEffect(() => {
+    emblaApi?.scrollTo(0, true);
+  }, [emblaApi, objectId, shop]);
+
+  if (!isEligible || (!isLoading && games.length === 0)) return null;
```
Moyasee: For a game with no recommendations this shows the heading and three skeleton cards, then deletes the whole section once the response lands and everything below jumps up. Either hold the section until the result is known, or keep a rendered empty state.

src/renderer/src/components/vertical-cover-card/vertical-cover-card.tsx
```diff
@@ -0,0 +1,122 @@
+import { ImageIcon } from "@primer/octicons-react";
+import type { ButtonHTMLAttributes, ReactNode } from "react";
+import { useEffect, useRef, useState } from "react";
+
+import { isAnimatedCoverCandidate, useCoverPoster } from "@renderer/hooks";
+import { getVerticalCoverCardImageSources } from "./vertical-cover-card-image-sources";
```
Moyasee, line 6: `.cursorrules:46` asks for `T[]` over `Array<T>`. Also at `vertical-cover-card-image-sources.ts:2` and `similar-games.test.ts:137`. `similar-games.ts:41` in this same PR already uses `readonly unknown[]`.

Moyasee (commented):
No live asks left from either round, and the tests cover the custom-game path. Four small things inline, none blocking.

Two non-code notes. Hachi-R's changes-requested was submitted against `3a4351e4` and you force-pushed `d7d07e14` nine minutes later with the fix in it, so it is stale but will keep blocking merge until it is dismissed or re-reviewed. I will sort that out. And the branch needs an update from main before it lands, it merges clean, just behind.

src/big-picture/src/pages/game/game.tsx
```diff
@@ -548,19 +607,33 @@ export default function Game() {
...
+    !areSimilarGamesLoading && similarCarouselGames.length > 0;
+  const shouldRenderSimilarGames = areSimilarGamesLoading || hasSimilarGames;
```
Moyasee: The two platforms now shift in opposite directions. Big picture reserves the skeleton's space and then collapses roughly 550px if the fetch resolves empty, which both reachable paths do (an API error logs and clears, an empty response returns nothing). Desktop went the other way and renders nothing at all while loading, so the section pops in above the reviews and pushes them down.

Either behavior is fine, they just want to be the same one on both. Reserving space and keeping an empty state is the less jarring of the two.

Nit while you are here: `SIMILAR_GAMES_SKELETON_COUNT = 9` at line 113 duplicates `SIMILAR_GAMES_LIMIT` in `similar-games.ts:3`.

Moyasee (commented):
All four are done, and the genre helper landing in its own module with the behavior change written into the body is exactly right. One thing left, inline.

Worth knowing the block on this is just the absence of an approval now. Dismissing the stale review removed the block but did not add one, so it needs a green tick rather than anything from you.

src/renderer/src/hooks/use-similar-games.ts
```diff
@@ -0,0 +1,83 @@
...
+      return;
+    }
+
+    setIsLoading(true);
+
+    const get: SimilarGamesGet = (path, options) =>
+      globalThis.window.electron.hydraApi.get<unknown>(path, options);
+
+    const loadSimilarGames = async () => {
+      const sources = (await globalThis.window.electron.leveldb.values(
```
Moyasee: The other readers of this sublevel go through the service wrapper and order with lodash: `levelDBService.values("downloadSources")` then `orderBy(sources, "createdAt", "desc")` at `game-details.context.tsx:416` and `home.tsx:45`. Same result as the hand-rolled sort, so this is purely about matching the form already in use.

src/renderer/src/pages/game-details/similar-games/similar-games.tsx
```diff
@@ -0,0 +1,110 @@
...
+      emblaApi.off("select", syncControls);
+      emblaApi.off("reInit", syncControls);
+    };
+  }, [emblaApi, syncControls]);
+
+  useEffect(() => {
+    emblaApi?.scrollTo(0, true);
+  }, [emblaApi, objectId, shop]);
+
+  if (!isEligible || isLoading || games.length === 0) return null;
```
Moyasee: Returning null while loading holds the whole section back, so the heading and three cards land roughly 250ms in and push the reviews block down. Navigation resets scroll to the top (`app.tsx:477`) so that shift is below the fold in practice, but Big Picture solves the same thing in this PR with `SimilarGamesSkeleton` (`game.tsx:1529`) and desktop is left without one. Worth rendering a placeholder here so the two surfaces behave the same.

Moyasee (commented):
Both asks from last round landed: `extractGenreNames` sits next to the metadata row now with the behavior change written into the description, and Big Picture renders a skeleton instead of popping the carousel in. The `Array<T>` spots are gone too.

The question left open on 08-07 is settled. The similar endpoint does honor `downloadSourceIds`, and results come back with their source lists filled in, so the Big Picture badges are accurate.

Hachi-R's changes-requested is dismissed and the branch is current with main, so nothing stale is holding this up.

Two nits inline, neither is a defect. Once they are sorted we can approve and merge this.

Hachi-R (approved):
No blocking issues found. The previous service-wrapper and desktop-skeleton notes are addressed in the current head, and all required checks pass.

src/big-picture/src/locales/en/translation.json
```diff
@@ -736,6 +736,8 @@
     "download_sources_removed_other": "{{count}} download sources were removed",
     "download_sources_count_one": "{{count}} download source",
     "download_sources_count_other": "{{count}} download sources",
+    "similar_games": "More like this",
```
chubbygrannychaser, line 739: [P1] Put these Big Picture keys at the namespace root, or change the lookup to use the format key prefix. src/big-picture/src/pages/game/game.tsx calls t("similar_games") and t("no_similar_games") with the big_picture namespace, but these entries are nested under format, so the new heading and empty-state text resolve to missing-key output. Apply the same fix to the added non-English Big Picture locale entries.

chubbygrannychaser (changes requested):
## Review summary

The current head still has one blocking localization issue. The Big Picture similar-games heading and empty state use root-key lookups, but the new Big Picture locale entries are nested under format. See the inline comment for the fix.

Focused tests, TypeScript validation, formatting, lint, and git diff check passed.

src/big-picture/src/locales/en/translation.json
BITO21: Fixed in a2ec36b9c. The format object is only the source-file organization: formatResources extracts it and the loader registers that object directly as the big_picture namespace root. I extracted that registration into a testable helper and added an integration test that reads the real locale JSON files and verifies the similar_games and no_similar_games root-key lookups for en, es, fr, pt-BR, and ru. I also added localized empty-state copy for the four non-English locales. The format prefix remains intentionally unused because that level is not registered at runtime. Focused tests (12/12), full typecheck, ESLint, Prettier, and git diff check pass.

chubbygrannychaser (approved):
## Review summary

The current head resolves the previous Big Picture localization issue. The resource registration now exposes the format resources at the namespace root, and the added test verifies both similar-games strings across all five locales. No blocking correctness issues remain.

Focused tests, TypeScript validation, formatting, lint, and git diff check passed.

PR #2544, opened 2026-07-16
Feature: QoL frontend updates for torrent downloads

Tiger-Tig74:
**When submitting this pull request, I confirm the following (please check the boxes):**

- [x] I have read the [Hydra documentation](https://docs.hydralauncher.gg/getting-started.html).
- [x] I have checked that there are no duplicate pull requests related to this request.
- [x] I have considered, and confirm that this submission is valuable to others.
- [x] I accept that this submission may not be used and the pull request may be closed at the discretion of the maintainers.

**Fill in the PR content:**
Added many UI based QoL updates regarding downloading torrents, including custom tracker list implementation via the settings, and a modal on the downloading game hero view that includes:
- A Details page that shows all the trackers and their status (working/failed).
- A Statistics page with download/upload speed, total uploaded/download, upload ratio and bytes remaining to be downloaded.

This also includes the implementations from: #2510

PR #2585, opened 2026-07-24
feat: global trackers system

Sociable-Fish:
feat: global trackers system

Adds settings UI to configure global trackers auto-appended to every new torrent download. Manual trackers (textarea) + remote URL list, auto-saves with 1s debounce.

Key changes:
- Settings section under Downloads tab (two toggle+input groups)
- global-trackers.ts helper — reads prefs from LevelDB, fetches+caches remote lists (1h TTL), deduped concurrent fetch guard, merges manual+URL trackers
- tracker-list.ts — validates tracker URLs (http/https/udp/ws/wss) and list URLs (http/https only), parses newline-separated text, strips comments, dedupes
- saveGlobalTrackers IPC — validates+persists to LevelDB, invalidates URL+memo caches
- startGameDownload + addGameToQueue call getGlobalTrackers() and attach customTrackers[]
- getTorrentFiles passes global trackers to metadata resolution
- Python RPC — validate_trackers() server-side, _build_add_torrent_params merges magnet+user+default trackers with tiering (tier 0 user, tier 1 defaults), _apply_trackers() adds to running torrents
- resumeSeeding + RPC action("start") pass customTrackers
- TorrentInvalidTrackers error type mapped end-to-end
- Startup refresh of URL cache
- Download Sources settings extracted to own sidebar tab; Global Trackers replaces its spot in Downloads tab
- Locale keys for en and pt-BR


python_rpc/torrent_downloader.py
```diff
@@ -143,8 +149,22 @@ def _build_add_torrent_params(self, magnet: str, save_path: str, flags):
...
+                continue
+
+            magnet_trackers.append(tracker)
+            known_trackers.add(tracker)
+            tiers.append(0)
+
+        fallback_tier = max(tiers) + 1 if tiers else 0
 
         tiers = list(params.tracker_tiers)[: len(trackers)]
         tiers.extend([0] * (len(trackers) - len(tiers)))
```
Moyasee: These two lines are a leftover from the merge and need to go, along with the duplicate `fallback_tier` on the line below them.

`trackers` here is now the function parameter, not the accumulated tracker list. Lines 156-157 already built `tiers` correctly against `magnet_trackers`, and line 167 already computed `fallback_tier` from it. This block then throws both away and rebuilds `tiers` sized by the length of the caller's list.

Two consequences:

1. `len(trackers)` raises `TypeError: object of type 'NoneType' has no len()` when no global trackers are configured, which is the default. That is every existing user, on every new download.
2. Even with a list, `params.tracker_tiers` ends up shorter than `params.trackers` by the number of magnet trackers, so the tiering this function exists to set up is lost. With an empty list every tracker collapses to tier 0, where `main` produces `[0, 1, 2, 2, ...]`.

Deleting lines 169 through 172 should be the whole fix, since the first block already produces both arrays.

src/shared/tracker-list.ts
```diff
@@ -0,0 +1,27 @@
+const VALID_TRACKER_PROTOCOLS = ["http:", "https:", "udp:", "ws:", "wss:"];
+const VALID_TRACKER_LIST_PROTOCOLS = ["http:", "https:"];
+
+export const isValidTrackerUrl = (url: string): boolean => {
+  try {
+    return VALID_TRACKER_PROTOCOLS.includes(new URL(url).protocol);
```
Moyasee: This accepts a tracker with no host, which the Python side rejects. `new URL("udp://").protocol` is `"udp:"`, so `udp://` and `udp:///announce` both pass here, get saved to preferences, and then fail `validate_trackers` because it requires `parsed.netloc`.

The user-visible result is that every torrent download fails with the invalid-trackers error, with nothing flagged at save time.

Checking the host would line this up with the Python rule:

```ts
export const isValidTrackerUrl = (url: string): boolean => {
  try {
    const parsed = new URL(url);
    return VALID_TRACKER_PROTOCOLS.includes(parsed.protocol) && !!parsed.host;
  } catch {
    return false;
  }
};
```

Same applies to `isValidTrackerListUrl`. Worth a case in the test file too, since the current tests only cover bad protocols and unparseable input, not a valid protocol with an empty host.

src/main/events/torrenting/start-game-download.ts
```diff
@@ -37,33 +38,38 @@ const startGameDownload = async (
...
+      timestamp: Date.now(),
+      queued: true,
+      pinnedToHero: false,
+      extracting: false,
+      automaticallyExtract,
+      automaticallyDeleteArchiveFiles,
+      fileIndices,
+      selectedFilesSize,
+      fileSize: selectedFilesSize ?? null,
+      ...(globalTrackers.length ? { customTrackers: globalTrackers } : {}),
```
Moyasee: This conditional spread is what makes the Python crash the default path rather than an edge case. With no global trackers configured, `customTrackers` is absent from the record, so `download.customTrackers` is `undefined` downstream and the key disappears at `JSON.stringify`.

Always setting the field, `customTrackers: globalTrackers`, would keep the wire shape stable and mean Python receives `[]` instead of `None`. That is worth doing regardless of the Python fix, since it also keeps stored records consistent.

src/main/events/torrenting/get-torrent-files.ts
```diff
@@ -43,11 +47,14 @@ const getTorrentFiles = async (
...
 
   try {
+    const trackers = await getGlobalTrackers();
+
     const response = await PythonRPC.rpc.call<TorrentFilesResponse>(
       "torrent_files",
       {
         magnet,
         timeout_ms: 45_000,
+        trackers: trackers.length ? trackers : undefined,
```
Moyasee: This sends `undefined` explicitly whenever the user has no global trackers, so it hits the `len(None)` crash on the metadata path. That makes the file-selection modal fail for anyone in the default configuration.

Passing `trackers` unconditionally would avoid it.

src/main/services/download/download-manager.ts
```diff
@@ -1033,6 +1033,7 @@ export class DownloadManager {
       game_id: levelKeys.game(download.shop, download.objectId),
       url: download.uri,
       save_path: download.downloadPath,
+      trackers: download.customTrackers,
```
Moyasee, line 1074: This is the path I would worry about most. Seeding is restored at startup, and every download record already in a user's LevelDB predates this change, so `customTrackers` is `undefined` for all of them. Combined with the `len(None)` issue, resuming seeding would fail on launch for existing users even if they never touch the new setting.

src/main/helpers/global-trackers.ts
```diff
@@ -0,0 +1,197 @@
...
+  if (appendUrl && url && isValidTrackerListUrl(url)) {
+    const cache = await getGlobalTrackersUrlCache();
+    if (cache?.url === url) {
+      urlCache = cache.trackers;
+    }
+
+    if (isGlobalTrackersUrlCacheStale(cache, url)) {
+      if (!cache || cache.url !== url) {
+        try {
+          urlCache = await dedupedFetchAndCache(url);
```
Moyasee: This await sits in the download-start path. `AXIOS_TIMEOUT_MS` is 15s, and `saveGlobalTrackers` deletes the cache whenever the URL changes, so the first download after editing the URL can block here for up to 15 seconds before `validateDownloadUrl` even runs.

The background-refresh branch below is correctly fire-and-forget. It is only this cold-cache branch that blocks. Falling through with whatever is cached, or an empty list, and letting the fetch happen in the background would keep the start path responsive.

python_rpc/torrent_downloader.py
```diff
@@ -261,6 +301,7 @@ def start_download(
                 if not selective_download:
                     self.torrent_handle.set_flags(lt.torrent_flags.auto_managed)
                     self.torrent_handle.resume()
+                    self._apply_trackers(trackers)
```
Moyasee, line 299: Minor: `_apply_trackers` is only reached in the non-selective branch, so resuming a selective download against an existing handle will not pick up trackers. Low impact, but the asymmetry looks unintentional.

Moyasee (changes requested):
Reviewed this closely given it touches the download path. Two blockers, one of which takes out torrent downloads for everyone in the default configuration.

## Blocker 1: leftover block in `_build_add_torrent_params`

`main` has a single tiers block. This branch has the new one plus the old one still present, and `trackers` in the leftover now binds to the new function parameter instead of the local tracker list. Details inline, but the effect:

I ran the branch's actual method with a stubbed libtorrent, magnet carrying 2 trackers at tiers `[0, 1]`:

```
trackers=None      -> TypeError: object of type 'NoneType' has no len()
trackers=[u1, u2]  -> 98 trackers vs 96 tiers      LENGTH MISMATCH
trackers=[]        -> 96 trackers vs 94 tiers      LENGTH MISMATCH
                      all tiers 0; main produces [0, 1, 2, 2, ...]
```

`None` is the default path rather than an edge case. `appendGlobalTrackers ?? false` means `getGlobalTrackers()` returns `[]` for anyone who has not configured this, `customTrackers` is then omitted from the `Download` record, `trackers: download.customTrackers` is `undefined`, and `JSON.stringify` in `python-rpc.ts:263` drops the key. Python sees `None`, `validate_trackers` passes it straight through, and `len(None)` throws.

Three paths reach it: starting a torrent download, resuming seeding at startup, and the file-selection metadata fetch.

Worth flagging that CI cannot catch this. Lint, Sonar and both builds pass because nothing exercises `_build_add_torrent_params` in Python.

## Blocker 2: settings can persist a tracker that Python rejects

`isValidTrackerUrl` never checks the host, while `validate_trackers` requires `netloc`:

| input | TS, saves it | Python |
| --- | --- | --- |
| `udp:///announce` | accepts | rejects |
| `udp://` | accepts | rejects |
| `udp://:6969/announce` | rejects | accepts |

Typing `udp://` into the textarea saves without complaint, gets attached to every subsequent download, and Python then rejects the entire list. Every torrent download fails with "The tracker list contains invalid trackers", with no error at save time and no indication of which line is at fault.

## Locales

pt-BR is missing `settings.global_trackers`, the section heading. The other seven landed.

pt-PT, ru and es are missing all eight new keys. Those three already carry pre-existing drift of 443, 17 and 70 keys respectively, so this is additive rather than something introduced here, but the new keys are worth adding while this is open.

## Smaller things

- `save-global-trackers.ts` gates the URL cache delete on `userPreferences` being truthy, so a first-run save would not clear a stale cache. Practically unreachable, noting it for completeness.
- The description says "Old per-download tracker management files (set-download-trackers.ts, manage-trackers-modal.tsx, validate-global-trackers-url.ts) removed". None of those exist in `main` or on this branch. Looks like text carried over from #2544. Worth correcting so reviewers do not go looking for a regression that is not there.
- Title is still "Trackers feature test" and there is no label on the PR.

## What looks good

The settings tab addition is purely additive with no index-based routing, and `download_sources` already exists in all five locales. The startup cache refresh is guarded and fire-and-forget. The new error type is mapped end to end. Deduplicating concurrent URL fetches is a nice detail. The committed test fits existing convention since `yarn test` already globs `src/**/*.test.ts`.

Blocking on 1 and 2. Blocker 1 is a four-line deletion, but because it breaks downloads and seeding in the default state it needs a real end-to-end check with global trackers both off and on before this moves.

src/main/events/torrenting/add-game-to-queue.ts
```diff
@@ -28,38 +28,43 @@ const addGameToQueue = async (
...
+      shop,
+      objectId,
+      status: "paused",
+      progress: 0,
+      bytesDownloaded: 0,
+      downloadPath,
+      downloader,
+      uri,
+      folderName: null,
+      fileSize: selectedFilesSize ?? null,
```
Moyasee: This one needs reverting. The `parseBytes` fallback that `c542918` removed was not dead.

`main` has:

```ts
const parsedFileSize = parseBytes(fileSize ?? null);
...
fileSize: selectedFilesSize ?? parsedFileSize,
```

and the renderer still sends `fileSize: repack.fileSize` at `game-details.tsx:120`, with `StartGameDownloadPayload.fileSize` still declared. So for anyone queuing without picking individual files, the repack's declared size used to be parsed and stored, and now `null` goes in instead.

Where that shows up:

- `download-group.tsx:741` gates on `download.fileSize != null && download.fileSize > 0`, so queued entries render no size at all
- `download-manager.ts:552` falls back to `download.selectedFilesSize ?? download.fileSize ?? 0`, which now resolves to `0`

I suspect the confusion came from `start-game-download.ts`, which genuinely does use `?? null` in `main`. The two files differed deliberately: the queue path has a repack size available up front and no metadata yet, the start path gets its size from metadata shortly after.

Worth noting this is also outside the scope of a trackers PR, so reverting keeps the diff honest.

src/shared/tracker-list.ts
```diff
@@ -0,0 +1,34 @@
+const VALID_TRACKER_PROTOCOLS = ["http:", "https:", "udp:", "ws:", "wss:"];
+const VALID_TRACKER_LIST_PROTOCOLS = ["http:", "https:"];
+
+export const isValidTrackerUrl = (url: string): boolean => {
+  try {
+    const parsed = new URL(url);
+    return (
+      VALID_TRACKER_PROTOCOLS.includes(parsed.protocol) && !!parsed.hostname
```
Moyasee: The `udp://` cases are properly closed now. There is a narrower version of the same split still left, for the schemes the WHATWG parser treats as special:

```
                     TS hostname   Python netloc
https:///x           "x"           ""
http:///announce     "announce"    ""
wss:///a             "a"           ""
udp:///announce      ""            ""            <- correctly rejected by both
```

Node collapses the extra slash for http/https/ws/wss and reads the first path segment as the host, while `urlparse` leaves `netloc` empty. So a triple-slash typo on one of those saves fine here and then gets the whole list rejected by `validate_trackers`, which is the same all-downloads-fail mode as before.

`udp://` is safe precisely because `udp:` is not a special scheme, which is why the fix works there and not here.

Requires a typo to hit, so not a blocker on its own. Something like `parsed.href.startsWith(parsed.protocol + "//" + parsed.host)`, or just rejecting a raw `:///` substring before parsing, would close it. A case in the test file alongside the protocol-only ones would be worth having either way.

Moyasee (dismissed):
Re-reviewed at `829a257`. Both blockers are genuinely fixed and I verified them rather than going off the commit messages. One new issue came in with the cleanup commit, which is why this is still changes-requested rather than an approval.

## Verified fixed

**The `_build_add_torrent_params` leftover.** I re-ran the same harness against the updated file, stubbed libtorrent, magnet carrying 2 trackers at tiers `[0, 1]`:

```
trackers=None      -> OK   96 trackers / 96 tiers   tiers [0,1,2,2,...]
trackers=[u1,u2]   -> OK   98 trackers / 98 tiers   tiers [0,1,0,0,2,2,...]
trackers=[]        -> OK   96 trackers / 96 tiers   tiers [0,1,2,2,...]
```

No crash, lengths agree, and the tiering matches what `main` produces, with user trackers landing at tier 0. You also added two further layers, `validate_trackers` returning `[]` and both call sites sending `customTrackers` unconditionally. Any one of the three would have covered it, so this is well defended now.

**Host validation.** Both validators check `hostname`, and the new tests cover exactly the cases that were failing.

**The blocking fetch.** Always fire-and-forget now. Worth being aware that a cold cache returns `[]`, so the first download right after configuring a URL will not carry the URL trackers and the one after it will. That seems like the right trade against stalling the start path.

**`_apply_trackers`.** Selective resume is covered, since the handle gets removed and re-added through `_build_add_torrent_params`. The `if trackers: self._apply_trackers(trackers)` after `add_torrent` is always redundant, because `params` already carries the trackers and `add_torrent` always runs at that point, but it dedups against the existing list so it does no harm. Fine to leave.

**pt-BR heading.** Leaving it to the English fallback makes sense given the value is "Trackers", which pt-BR keeps as-is anyway.

## Still open

The `fileSize` change is the blocking one, details inline. Short version: the `parseBytes` fallback that commit `c542918` removed was not dead, and dropping it means queued downloads lose their displayed size.

Smaller items:

- pt-PT, ru and es are still missing all eight new keys. Those files already carry pre-existing drift, so this is additive rather than something this PR caused, but it is a cheap thing to fold in while this is open.
- Title is still "Trackers feature test" and the PR has no label.
- The description still lists three "removed" files that do not exist in `main` or on this branch.

CI is green across the board now, both builds included.

Everything dangerous is resolved. What is left is a two-line revert plus some tidying.

Hachi-R (changes requested):
Re-reviewed at dbc6dfd. The previously reported fileSize fallback and malformed triple-slash URL issues are fixed, and the implementation otherwise looks good. Local typechecking and all 64 tests pass, changed-file ESLint has no errors, and the GitHub Build, Lint, and Text Integrity workflows are green.

One blocking failure-path issue remains at src/main/helpers/global-trackers.ts:27. finally() returns a new promise, and that derived promise is currently ignored. If fetchAndCacheGlobalTrackersUrl() rejects, the caller's catch handles the original promise, but the promise created here still rejects unhandled. Because network failures and timeouts are expected, this can terminate the Electron main process. Please store the finally()-chained promise in pendingUrlFetches and return that same promise, so the caller observes the rejection while cleanup still runs.

Hachi-R (approved):
Re-reviewed at 736b5c6. The previously blocking unhandled-rejection issue is fixed correctly by storing and returning the finally-chained promise. I verified the failure path no longer emits an unhandled rejection and that the pending entry is cleaned up. Typechecking, all 64 tests, focused ESLint, and diff checks pass locally, and the GitHub Build, Lint, and Text Integrity workflows are all green. Approved.

PR #2599, opened 2026-07-27
fix: Steam client not available inside the wine prefix

Moyasee:
Closes LBX-706.

**Fill in the PR content:**

On Linux, games that need the Steam client reachable from inside the wine prefix don't start when launched from Hydra. They fail with `Could not Initialize Steam`, `Steam is not installed, please install it first`, or nothing at all. The same executable works when added directly to Steam, and no combination of `WINEDLLOVERRIDES` in the launch options helps, which is what made this awkward to track down.

Games installed through Steam were hitting the same wall from the other direction, plus a second problem: Hydra was building them a brand new prefix instead of using the one Steam already has, so even the ones that did start had no saves.

Windows is untouched by all of this.

### What's going on

The bundled umu-launcher declares `STEAM_COMPAT_CLIENT_INSTALL_PATH` in its env dict as an empty string, never fills it in, and then does `os.environ[key] = val` over the whole dict before handing off to Proton. The variable gets blanked no matter what we pass in.

Proton needs that path. It reads it in `setup_prefix()` and uses it to copy `steamclient.dll`, `steamclient64.dll`, `Steam.dll`, `GameOverlayRenderer64.dll` and `SteamService.exe` out of `<steam>/legacycompat` into `C:\Program Files (x86)\Steam` inside the prefix. With an empty path every copy silently no-ops, and the prefix is left with `SteamClientDll64` in the registry pointing at a `steamclient64.dll` that isn't there. Proton's own dist doesn't carry that file either, only `lsteamclient.dll`, so nothing fills the gap.

Confirmed empirically on Supermarket Together: the looping `Sys_LoadModule failed` in the Proton log drops to zero once the files are in place.

### Installed Steam games now go through Steam

If the executable sits under `steamapps/common` and there's an `appmanifest_*.acf` next to it whose `installdir` matches the folder, Hydra opens `steam://rungameid/<appid>` instead of launching it itself. Steam's own prefix, so cloud saves, the overlay, achievements and any launch options set in Steam all work, and there's no way for us to damage a prefix a user cares about.

Reusing Steam's `compatdata` prefix directly was the other option and I dropped it: on a Proton downgrade the script hits `Removing newer prefix`, calls `remove_tracked_files()` and drops the guard, which would take the user's real prefix and saves with it.

The manifest match matters. A game Hydra downloaded into that folder has no manifest claiming it, so it still launches the normal way. If Steam isn't actually installed, which can happen when a library folder outlives the uninstall, we fall back to launching directly rather than opening a URL nothing will answer.

Two things had to move for this to work:

- `hasLinuxCompatibilityProcessMatch` rejected any process whose `STEAM_COMPAT_DATA_PATH` didn't equal the prefix Hydra expected. For a game Steam started that's never true, so playtime never recorded. The check is now skipped for executables inside a Steam library. `closeGame`'s fallback lookup had the same problem and the same fix.
- Launch options set in Hydra don't reach a game Steam starts. Rather than dropping them quietly, the field is disabled with a note pointing at Steam, and the main process logs when a stored value is being ignored. The Clear button stays available so a leftover value can still be removed.

### Everything else that needed the Steam client

New `steam-client-compat` service. When a game needs the Steam client it does the copy Proton would have done if the variable had survived, and passes `STEAM_COMPAT_CLIENT_INSTALL_PATH` along too, so this starts working on its own if umu ever stops blanking it.

"Needs the Steam client" means one of: marker files next to the executable, the game living in a Steam library, or a Steamworks DLL somewhere in the folder. That last one is a bounded scan, 8 levels and 8000 entries, which is what it takes to reach the ones buried inside an Electron bundle. Anything that doesn't match resolves to `null` and takes exactly the path it does today.

Copies are skipped when size and mtime already match, so we're not moving 21 MB on every launch.

For games shipping their own proxy loaders, `=n,b` overrides are added for the ones actually present in the folder (`winmm`, `dinput8`, `dsound`, `version`). `Umu.launchExecutable` gained an `extraEnv` option, and `WINEDLLOVERRIDES` and `LD_PRELOAD` are merged rather than overwritten so anything in a user's launch options still wins.

The overlay is enabled the same way Steam does it, `gameoverlayrenderer.so` preloaded and `SteamOverlayGameId` set, reading the app id out of the game's own config when it declares one. That part stays gated on the marker files, since games launched by Steam already get the real thing.

### Removed the forced Proton logging

`PROTON_LOG=1` was being set on every umu launch. That was writing multi-megabyte `steam-<id>.log` files into `$HOME` and, more importantly, was the cause of a freeze a tester reported. The stalled region of the log was wall-to-wall `RtlCaptureStackBackTrace`. Proton's own source is explicit that logging is off by default for performance.

It's gone now. Putting `PROTON_LOG=1` in the launch options still turns it on, and `PROTON_LOG_DIR` points at Hydra's logs directory so it no longer lands in the home folder.

### A correction

An earlier revision of this branch set `UMU_USE_STEAM=1`. That was wrong and has been reverted. It makes Proton append `lsteamclient=d`, disabling the module that actually backs `steamclient64.dll` under wine, and a tester hit it directly:

```
err:steam:setup_steam_registry Failed to load lsteamclient module, skipping initialization
err:steam:setup_steam_files Could not convert "" to win path.
```

With the bridge disabled the load falls through to the genuine Windows `steamclient64.dll`, which cannot initialize under wine, and the game reports `Failed to load original steamclient. Error code: 998` (`ERROR_NOACCESS`).

I also want to retract the verification claim from the first version of this description. I said `ActiveProcess\PID` and `steamapps/libraryfolders.vdf` appearing in the prefix proved the Steam client had come up. Diffing a working prefix against a broken one shows those are identical in both cases. They're written by Proton's `steam.exe` stub either way and prove nothing.

### What's verified and what isn't

Verified locally: the copy lands the files, is idempotent, re-copies when the source changes, and leaves games that don't need it untouched. On Supermarket Together the repeating `Sys_LoadModule failed` in the Proton log goes to zero once the files are seeded. A tester confirmed games open that previously didn't, that the overlay comes up on one of them, and that the log files stopped appearing in the home directory.

Not yet confirmed end to end on the last few commits: that a Steam-library game keeps its saves across a Hydra launch, that playtime records for it, and that the Stop button closes it. The freeze fix is also diagnosis-driven rather than reproduced on demand, since it came out of reading the stalled region of the log. Those are the parts that need eyes before this merges.


Hachi-R (changes requested):
Requesting changes based on four functional issues in the Linux Steam handoff path:

1. src/main/services/process-watcher.ts:199 and the equivalent close-game.ts condition treat every executable under steamapps/common as Steam-launched. Games without a matching manifest, or launches where Steam is unavailable, fall back to Hydra but still skip prefix isolation. This can attribute playtime to or stop a Wine process using another prefix. Please gate the relaxation on an actual/resolved Steam handoff.

2. src/renderer/src/pages/game-details/modals/game-options-modal.tsx:749 marks settings as Steam-managed from the manifest alone. launchThroughSteam can still fail and fall back to a direct Hydra launch, but the Wine prefix, Proton, and launch-option controls remain disabled. Please keep the fallback settings usable or derive this state from the same handoff-availability decision.

3. src/renderer/src/pages/game-details/modals/game-options-modal/general-section.tsx:366 hides the Clear button whenever Steam manages the game, so stored Hydra launch options cannot be removed. This also contradicts the PR description. Please show Clear whenever a stored value exists, even if the input is disabled.

4. src/main/services/steam-client-compat.ts:250 always prefers objectId because the caller always supplies it. Custom-game object IDs are UUIDs, not Steam App IDs, so layouts without FakeAppId receive an invalid SteamOverlayGameId and never reach the 480 fallback. Please only use objectId when it is numeric.

The Steam-client prefix seeding itself looks directionally sound, but these regressions should be addressed before merge.

Moyasee:
All four addressed in c507056cd.

1. Added a `steamLaunchedGames` set that `launchGame` writes to only when `launchThroughSteam` actually returns true, and clears on the fallback path and on game close. The process watcher and `closeGame` check that instead of the path, so a game under `steamapps/common` with no matching manifest, or one where Steam is missing, keeps prefix isolation.

2. `resolveSteamLaunchInfo` now returns null unless `isSteamAvailable()` passes, which is the same gate `launchThroughSteam` uses. If Steam is gone the modal shows the normal controls again.

3. Clear is back whenever a stored value exists, disabled input or not. That was my mistake, the description was right.

4. `objectId` only feeds `SteamOverlayGameId` when it matches `/^\d+$/`, so custom game UUIDs fall through to 480.

One thing worth flagging: the set lives in memory, so if Hydra restarts while a handed-off game is running it loses the mark and prefix isolation kicks back in for that session. Playtime would stop recording for that run. Happy to persist it if you would rather not have that gap.

Hachi-R (changes requested):
Thanks for addressing the previous review. The Clear-button and numeric overlay-ID issues are fixed, and the prefix-isolation change now works for successful handoffs within the same Hydra process. I’m still requesting changes for three remaining issues:

1. **[P1] Flatpak-only Steam installations cannot seed the Steam client files** (`src/main/services/steam-client-compat.ts:286`). `isSteamAvailable()` recognizes `~/.var/app/com.valvesoftware.Steam`, but `getSteamLocation()` only searches `~/.steam/steam` and `~/.local/share/Steam`. On Flatpak-only systems this therefore looks available to the handoff/UI while `requiredFilePath` points at a nonexistent native path and the compatibility bridge returns null. Please resolve the Flatpak Steam root as well (and cover it with a test).

2. **[P2] Steam handoff state is lost when Hydra restarts** (`src/main/services/process-watcher.ts:199`). `steamLaunchedGames` is memory-only, so restarting Hydra while a Steam-started game remains open makes the watcher enforce Hydra’s expected prefix again and stop recognizing the Steam compatdata process. That loses running/playtime state for the rest of the session. Prefer deriving this from the live process prefix and the manifest-resolved Steam prefix, or persist and revalidate the handoff.

3. **[P2] The settings modal still disables fallback settings before a handoff succeeds** (`src/renderer/src/pages/game-details/modals/game-options-modal.tsx:749`). `resolveSteamLaunchInfo()` only establishes that a manifest and an apparently available Steam install exist. `shell.openExternal()` can still reject, in which case `launchThroughSteam()` returns false and Hydra launches directly, but `steamLaunchInfo !== null` keeps Wine/Proton/launch-option controls disabled. Please keep the fallback controls usable or base this state on an actual handoff result.

Validation at `c507056`: node and renderer type-checks pass. The full suite is 76/77 with only the existing Windows `EBUSY` cleanup failure in `notification-icon.test.ts`.

PR #2601, opened 2026-07-30
feat: support archive.org direct downloads

Moyasee:

**Fill in the PR content:**

Adds `archive.org` as a downloader, so download sources can point at Internet Archive file links. Closes LBX-991.

```
https://archive.org/download/megabonk.zip
```

A link like this behaves exactly like any other direct download link. The URL is resolved locally into a canonical archive.org URL and handed to the HTTP downloader, which already follows the redirect to the node server and resumes through range requests, so there is no metadata lookup and no extra request. The file is saved in the download folder under its own name and extracted like any other archive. Node server links (`ia800101.us.archive.org/25/items/<item>/<file>`) are normalized to the same canonical form, and nested paths keep only the file name.

Links that do not point at a file (`/download/<item>`, `/download/<item>/subfolder/`, `/details/<item>`) are deliberately not accepted, so `getDownloadersForUri` does not offer archive.org for them and the option stays unavailable instead of failing at start. No API key or account is involved.

The whole resolver lives in `src/shared/archive-org.ts`, so the renderer and the main process apply the same rule.

`sanitizeRelativePath` now drops `.` and `..` segments as well. Unrelated to archive.org strictly speaking, but it is the function that turns a remote listing into an on-disk path, and it should not be able to write outside the download folder.

### Tested

Against live archive.org files:

- `Megabonk.zip` (217,483,656 bytes) and `Alleyway (USA).gb` (32,768 bytes) resolve and download with the expected byte counts, valid content, and a `206` on resume.
- Nested and node-host links resolve to the right canonical URL and file name.
- Item, folder, details, metadata, traversal (`/download/../etc/passwd`) and non-archive links are all rejected by the resolver.
- `yarn typecheck` (node + web), lint and the existing test suite pass.

### DIFF
--- a/src/big-picture/src/constants.ts
+++ b/src/big-picture/src/constants.ts
@@ -23,4 +23,5 @@ export const DOWNLOADER_NAME: Record<Downloader, string> = {
   [Downloader.Rootz]: "Rootz",
   [Downloader.Premiumize]: "Premiumize",
   [Downloader.AllDebrid]: "AllDebrid",
+  [Downloader.ArchiveOrg]: "Archive.org",
 };
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -590,6 +590,7 @@
     "download_error_torrent_too_many_files": "This torrent has too many files to display safely.",
     "download_error_torrent_files_unavailable": "Could not fetch torrent files. Please try again.",
     "download_error_torrent_invalid_trackers": "The tracker list contains invalid trackers. Please check the list and try again.",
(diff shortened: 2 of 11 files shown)


Moyasee:
Scope reduced: item and folder downloads are out, only direct file links are supported now. The metadata API client, the file-selection step for archive.org and the batch generalization are gone, so the diff is down to ~130 added lines. @greptile review

chubbygrannychaser (approved):
LGTM

PR #2603, opened 2026-07-30
feat: unlock datanodes and vikingfile through the hydra api

Moyasee:

**Fill in the PR content:**

Moves both remaining scraped/Nimbus hosters over to the launcher API.

- Datanodes no longer posts the fake `download2` form to `datanodes.to` from the client. It calls `POST /hosters/datanodes/unlock` instead, which tries premium credentials first and falls back to the free flow server-side.
- VikingFile used to hit `MAIN_VITE_NIMBUS_API_URL/hosters/unlock`. Nimbus is dead, so it now calls `POST /hosters/vikingfile/unlock`. The redirect-following `HEAD` request is gone too, since the endpoint already returns the direct provider URL.
- `MAIN_VITE_NIMBUS_API_URL` is removed from `.env.example` and the env typings.

VikingFile is faster now but requires an active Hydra Cloud subscription and is capped at 10 uncached unlocks per UTC day, so the UI needs to say so:

- The desktop download modal marks VikingFile as unconfigured without a subscription, shows a "Hydra Cloud" badge next to it, and clicking it opens the Hydra Cloud modal instead of selecting the downloader.
- Big picture filters VikingFile out of the downloader list when there is no active subscription (there's nowhere to subscribe from there anyway).
- The main process passes `needsSubscription` to `HydraApi`, so it fails fast locally before spending a request.

Error mapping was reworked for both hosters: quota (429), subscription required (400 with a plain-text message), sign-in required (401), file gone (404) and provider unavailable (502) each get their own message. The old `download_error_vikingfile_nimbus_quota_exceeded` key mentioned Nimbus by name, so it was replaced. New strings are in en/pt-BR/pt-PT/ru/es, and the stale key was dropped from the remaining locales so they fall back to en.


src/main/helpers/download-error-handler.ts
```diff
@@ -1,9 +1,56 @@
...
+]);
+
+const handleHydraUnlockError = (
+  err: AxiosError,
+  downloader: Downloader
+): DownloadErrorResult | null => {
+  if (!HYDRA_UNLOCK_DOWNLOADERS.has(downloader)) return null;
+
+  const status = err.response?.status;
+  const message = (err.response?.data as { message?: unknown } | undefined)
```
chubbygrannychaser: The API contract described in this PR returns the VikingFile subscription failure as a plain-text `400` response. For a string body, this cast does not convert it to `{ message }`; `.message` is `undefined`, so the condition below misses the subscription error and we fall through to the generic Axios message. Normalize `err.response?.data` as either a string or an object message (and add a regression test) so subscribed-gating failures consistently show `VikingFileSubscriptionRequired`.

chubbygrannychaser (changes requested):
The documented subscription error is still not mapped correctly.

chubbygrannychaser (approved):
LGTM

chubbygrannychaser (changes requested):
Please restore a focused regression test for the current plain-text VikingFile 400 mapping. The fix correctly normalizes string and object response bodies, and the subscription refresh flow is present, but commit ac146cb8 removes src/main/helpers/download-error-handler.test.ts and its test resolver. The current head therefore has no automated coverage for the exact regression fixed after the prior review. Keep a focused test (using the project’s supported test setup if needed) that proves a plain-text 400 becomes VikingFileSubscriptionRequired before this is merged.

Moyasee:
Restored in 07bd3d27e. The test asserts that a plain-text 400 maps to `VikingFileSubscriptionRequired`, and covers the json body variant plus the rest of the unlock status mapping.

It had to bring `scripts/ts-path-aliases.mjs` back with it. `ts-node/esm` resolves neither the tsconfig path aliases nor extensionless relative imports, so `download-error-handler.ts` fails on its `@shared` import before any assertion runs. Every test currently in the repo imports alias-free leaf modules with explicit extensions, which is why this has not come up before. The resolve hook reads the aliases straight out of `tsconfig.node.json` rather than repeating them.

I did check the test actually catches the regression rather than passing by construction: stashing the normalization in `handleHydraUnlockError` turns the plain-text case red and leaves the other four green.

heliokroger (dismissed):
LGTM

chubbygrannychaser (approved):
LGTM

Moyasee:
@greptile review

PR #2606, opened 2026-07-30
feat(cloud-save): add native save discovery

Hachi-R:
## Summary

- add native cloud-save manifest downloading, caching, and indexing
- resolve cross-platform save paths and custom paths
- scan resolved rules and expose the native APIs through N-API

## Context

This is the first independent slice extracted from the final Cloud Saves V2 implementation and rebased onto the latest `main`. Follow-up PRs will build the snapshot, upload, restore, and UI layers on top of it.

## Validation

- `cargo fmt -- --check`
- `cargo check`
- `cargo test --lib` (42 passed)

### DIFF
--- a/native/hydra-native/Cargo.toml
+++ b/native/hydra-native/Cargo.toml
@@ -7,13 +7,31 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
+anyhow = "1.0"
+blake3 = "=1.8.5"
+globetter = "=0.2.0"
+globset = "=0.4.15"
 image = { version = "0.25.8", default-features = false, features = ["gif", "jpeg", "png", "webp"] }
+indexmap = { version = "2.14", features = ["serde"] }
 mime_guess = "2.0.5"
 napi = { version = "3.5.2", default-features = false, features = ["napi8", "tokio_rt"] }
 napi-derive = "3.3.2"
+reqwest = { version = "0.13", features = ["stream"] }
+serde = { version = "1.0.228", features = ["derive"] }
+serde_json = "1.0.150"
+serde_yaml_ng = "0.10"
+sha2 = "0.10"
 sysinfo = "0.37.2"
-tokio = { version = "1.50.0", features = ["rt"] }
+tokio = { version = "1.52", features = ["fs", "rt", "sync", "macros"] }
+unicode-normalization = "0.1.25"
 uuid = { version = "1.11.0", features = ["v4"] }
+walkdir = "=2.5.0"
+
+[target.'cfg(windows)'.dependencies]
+known-folders = "=1.2.0"
+
+[dev-dependencies]
+tempfile = "3.27"
 
 [build-dependencies]
 napi-build = "2.3.1"
--- a/native/hydra-native/src/cloud_save/manifest/cache.rs
+++ b/native/hydra-native/src/cloud_save/manifest/cache.rs
@@ -1,0 +1,209 @@
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, OnceLock};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use anyhow::{Context, Result};
+use tokio::fs;
+use tokio::sync::Mutex;
+
+use super::indexer::build_manifest_index;
+use super::types::ManifestIndex;
+use crate::constants::MANIFEST_INDEX_VERSION;
+
+const MANIFEST_CACHE_TTL_MS: i64 = 24 * 60 * 60 * 1000;
+const MANIFEST_HTTP_TIMEOUT_SECS: u64 = 30;
+const RAW_MANIFEST_FILE_NAME: &str = "cloud-save-manifest.yaml";
+const INDEX_FILE_NAME: &str = "cloud-save-manifest-index.json";
+
+type ManifestCache = Arc<Mutex<Option<ManifestIndex>>>;
+type ManifestCaches = Mutex<HashMap<PathBuf, ManifestCache>>;
+
+static CACHE_LOCKS: OnceLock<ManifestCaches> = OnceLock::new();
+static HTTP_CLIENT: OnceLock<reqwest::Client> = OnceLock::new();
+
+fn http_client() -> &'static reqwest::Client {
+    HTTP_CLIENT.get_or_init(|| {
+        reqwest::Client::builder()
+            .timeout(Duration::from_secs(MANIFEST_HTTP_TIMEOUT_SECS))
+            .build()
+            .expect("Failed to build cloud save manifest HTTP client")
+    })
+}
+
+fn now_ms() -> i64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_millis() as i64
+}
+
+fn is_index_expired(index: &ManifestIndex, current_time: i64) -> bool {
+    index.fetched_at + MANIFEST_CACHE_TTL_MS <= current_time
+}
+
+async fn cache_for(key: PathBuf) -> ManifestCache {
+    let caches = CACHE_LOCKS.get_or_init(|| Mutex::new(HashMap::new()));
+    let mut caches = caches.lock().await;
+    caches
+        .entry(key)
+        .or_insert_with(|| Arc::new(Mutex::new(None)))
+        .clone()
+}
+
+fn raw_manifest_path(user_data_path: &Path) -> PathBuf {
+    user_data_path.join(RAW_MANIFEST_FILE_NAME)
+}
+
+fn index_path(user_data_path: &Path) -> PathBuf {
+    user_data_path.join(INDEX_FILE_NAME)
+}
+
+async fn read_index(path: &Path) -> Option<ManifestIndex> {
+    let content = fs::read_to_string(path).await.ok()?;
+    let index: ManifestIndex = serde_json::from_str(&content).ok()?;
+    (index.version == MANIFEST_INDEX_VERSION).then_some(index)
+}
+
+async fn raw_manifest_fetched_at(path: &Path) -> Option<i64> {
+    let modified = fs::metadata(path).await.ok()?.modified().ok()?;
+    Some(
+        modified
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_millis() as i64,
+    )
+}
+
+async fn remove_cache_file(path: &Path) -> Result<()> {
+    match fs::remove_file(path).await {
+        Ok(()) => Ok(()),
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => Ok(()),
+        Err(error) => Err(error)
+            .with_context(|| format!("Failed to remove stale cache file {}", path.display())),
+    }
+}
+
+async fn write_atomically(path: &Path, content: &[u8]) -> Result<()> {
+    let parent = path
+        .parent()
+        .with_context(|| format!("Invalid manifest cache path: {}", path.display()))?;
+    fs::create_dir_all(parent)
+        .await
+        .with_context(|| format!("Failed to create cache directory {}", parent.display()))?;
+    let temp_path = path.with_extension(format!("{}.{}.tmp", std::process::id(), now_ms()));
+    if let Err(error) = fs::write(&temp_path, content).await {
+        return Err(error).with_context(|| {
+            format!(
+                "Failed to write temporary cache file {}",
+                temp_path.display()
+            )
+        });
+    }
+    if let Err(error) = fs::rename(&temp_path, path).await {
+        let _ = fs::remove_file(&temp_path).await;
+        return Err(error).with_context(|| {
+            format!(
+                "Failed to replace cache file {} with {}",
+                path.display(),
+                temp_path.display()
+            )
+        });
+    }
+    Ok(())
+}
+
+async fn write_index(path: &Path, index: &ManifestIndex) -> Result<()> {
+    let content = serde_json::to_vec_pretty(index)
+        .with_context(|| format!("Failed to serialize manifest index for {}", path.display()))?;
+    write_atomically(path, &content).await
+}
+
+async fn download_manifest(source_url: &str) -> Result<String> {
+    let response = http_client()
+        .get(source_url)
+        .send()
+        .await
+        .with_context(|| format!("Failed to download cloud save manifest from {source_url}"))?
+        .error_for_status()
+        .with_context(|| format!("Cloud save manifest request failed for {source_url}"))?;
+
+    response
+        .text()
+        .await
+        .with_context(|| format!("Failed to read cloud save manifest body from {source_url}"))
+}
+
+async fn load_index(user_data_path: &Path, source_url: &str) -> Result<ManifestIndex> {
+    let current_time = now_ms();
+    let index_file = index_path(user_data_path);
+    let raw_file = raw_manifest_path(user_data_path);
+    let disk_index = read_index(&index_file).await;
+    let has_matching_disk_index = disk_index
+        .as_ref()
+        .is_some_and(|index| index.source_url == source_url);
+    let disk_index = disk_index.filter(|index| index.source_url == source_url);
+    let mut fallback = disk_index.clone();
+
+    if let Some(index) = &disk_index {
+        if !is_index_expired(index, current_time) {
+            return Ok(index.clone());
+        }
+    }
+
+    let raw_yaml = if disk_index.is_some() {
+        fs::read_to_string(&raw_file).await.ok()
+    } else {
+        None
+    };
+
+    if let Some(raw_yaml) = raw_yaml {
+        let fetched_at = raw_manifest_fetched_at(&raw_file)
+            .await
+            .unwrap_or(current_time);
+        if let Ok(rebuilt) = build_manifest_index(&raw_yaml, source_url, fetched_at) {
+            // A valid rebuilt index remains usable in memory if refreshing disk fails.
+            let _ = write_index(&index_file, &rebuilt).await;
+            if !is_index_expired(&rebuilt, current_time) {
+                return Ok(rebuilt);
+            }
+            fallback = Some(rebuilt);
+        }
+    }
+
+    let fresh_result = async {
+        let raw_yaml = download_manifest(source_url).await?;
+        let fresh = build_manifest_index(&raw_yaml, source_url, now_ms())?;
+        if !has_matching_disk_index {
+            remove_cache_file(&raw_file).await?;
+            remove_cache_file(&index_file).await?;
+        }
+        write_atomically(&raw_file, raw_yaml.as_bytes()).await?;
+        write_index(&index_file, &fresh).await?;
+        Ok::<ManifestIndex, anyhow::Error>(fresh)
+    }
+    .await;
+
+    fresh_result.or_else(|error| fallback.ok_or(error))
+}
+
+pub async fn get_manifest_index(user_data_path: &Path, source_url: &str) -> Result<ManifestIndex> {
+    let cache = cache_for(user_data_path.to_path_buf()).await;
+    let mut cached = cache.lock().await;
+
+    if let Some(index) = cached.as_ref() {
+        if index.source_url == source_url && !is_index_expired(index, now_ms()) {
+            return Ok(index.clone());
+        }
+    }
+
+    let fallback = cached
+        .as_ref()
+        .filter(|index| index.source_url == source_url)
+        .cloned();
+    let index = load_index(user_data_path, source_url)
+        .await
+        .or_else(|error| fallback.ok_or(error))?;
+    *cached = Some(index.clone());
+    Ok(index)
+}
--- a/native/hydra-native/src/cloud_save/manifest/cache.rs
+++ b/native/hydra-native/src/cloud_save/manifest/cache.rs
@@ -8,6 +8,7 @@ use tokio::fs;
 use tokio::sync::Mutex;
 
 use super::indexer::build_manifest_index;
+use super::lookup::ManifestLookupIndex;
 use super::types::ManifestIndex;
 use crate::constants::MANIFEST_INDEX_VERSION;
 
@@ -16,7 +17,7 @@ const MANIFEST_HTTP_TIMEOUT_SECS: u64 = 30;
 const RAW_MANIFEST_FILE_NAME: &str = "cloud-save-manifest.yaml";
 const INDEX_FILE_NAME: &str = "cloud-save-manifest-index.json";
 
-type ManifestCache = Arc<Mutex<Option<ManifestIndex>>>;
+type ManifestCache = Arc<Mutex<Option<Arc<ManifestLookupIndex>>>>;
 type ManifestCaches = Mutex<HashMap<PathBuf, ManifestCache>>;
 
 static CACHE_LOCKS: OnceLock<ManifestCaches> = OnceLock::new();
@@ -187,23 +188,28 @@ async fn load_index(user_data_path: &Path, source_url: &str) -> Result<ManifestI
     fresh_result.or_else(|error| fallback.ok_or(error))
 }
 
-pub async fn get_manifest_index(user_data_path: &Path, source_url: &str) -> Result<ManifestIndex> {
+pub async fn get_manifest_index(
+    user_data_path: &Path,
+    source_url: &str,
+) -> Result<Arc<ManifestLookupIndex>> {
     let cache = cache_for(user_data_path.to_path_buf()).await;
     let mut cached = cache.lock().await;
 
     if let Some(index) = cached.as_ref() {
-        if index.source_url == source_url && !is_index_expired(index, now_ms()) {
-            return Ok(index.clone());
+        if index.manifest.source_url == source_url && !is_index_expired(&index.manifest, now_ms()) {
+            return Ok(Arc::clone(index));
         }
     }
 
     let fallback = cached
         .as_ref()
-        .filter(|index| index.source_url == source_url)
-        .cloned();
+        .filter(|index| index.manifest.source_url == source_url)
+        .map(Arc::clone);
     let index = load_index(user_data_path, source_url)
         .await
+        .map(ManifestLookupIndex::new)
+        .map(Arc::new)
         .or_else(|error| fallback.ok_or(error))?;
-    *cached = Some(index.clone());
+    *cached = Some(Arc::clone(&index));
     Ok(index)
 }
--- a/native/hydra-native/src/cloud_save/manifest/indexer.rs
+++ b/native/hydra-native/src/cloud_save/manifest/indexer.rs
@@ -1,0 +1,135 @@
+use anyhow::{bail, Context, Result};
+use indexmap::IndexMap;
+use serde_yaml_ng::{Mapping, Value};
+
+use super::types::{ManifestFileRule, ManifestGameEntry, ManifestIndex, ManifestRuleCondition};
+use crate::constants::MANIFEST_INDEX_VERSION;
+
+fn string_value(value: &Value) -> Option<String> {
+    value.as_str().map(ToString::to_string)
+}
+
+fn mapping_value<'a>(mapping: &'a Mapping, key: &str) -> Option<&'a Value> {
+    mapping.get(Value::String(key.to_string()))
+}
+
+fn read_tags(value: Option<&Value>) -> Vec<String> {
+    value
+        .and_then(Value::as_sequence)
+        .map(|items| items.iter().filter_map(string_value).collect())
+        .unwrap_or_default()
+}
+
+fn read_conditions(value: Option<&Value>) -> Vec<ManifestRuleCondition> {
+    value
+        .and_then(Value::as_sequence)
+        .map(|items| {
+            items
+                .iter()
+                .filter_map(|item| {
+                    let mapping = item.as_mapping()?;
+                    let os = mapping_value(mapping, "os").and_then(string_value);
+                    let store = mapping_value(mapping, "store").and_then(string_value);
+                    (os.is_some() || store.is_some()).then_some(ManifestRuleCondition { os, store })
+                })
+                .collect()
+        })
+        .unwrap_or_default()
+}
+
+fn map_game_entry(manifest_key: &str, value: &Value) -> Option<ManifestGameEntry> {
+    let game = value.as_mapping()?;
+    let files = mapping_value(game, "files")?.as_mapping()?;
+    let files = files
+        .iter()
+        .filter_map(|(raw_path, metadata)| {
+            let raw_path = string_value(raw_path)?;
+            let metadata = metadata.as_mapping();
+            Some(ManifestFileRule {
+                raw_path,
+                tags: metadata
+                    .map(|m| read_tags(mapping_value(m, "tags")))
+                    .unwrap_or_default(),
+                when: metadata
+                    .map(|m| read_conditions(mapping_value(m, "when")))
+                    .unwrap_or_default(),
+            })
+        })
+        .collect::<Vec<_>>();
+
+    (!files.is_empty()).then_some(ManifestGameEntry {
+        manifest_key: manifest_key.to_string(),
+        files,
+    })
+}
+
+pub fn build_manifest_index(
+    raw_yaml: &str,
+    source_url: &str,
+    fetched_at: i64,
+) -> Result<ManifestIndex> {
+    let root: Value = serde_yaml_ng::from_str(raw_yaml)
+        .with_context(|| format!("Failed to parse cloud save manifest from {source_url}"))?;
+    let Some(root) = root.as_mapping() else {
+        bail!("Cloud save manifest root from {source_url} must be a YAML map");
+    };
+    let mut games = IndexMap::new();
+
+    for (key, value) in root {
+        let Some(manifest_key) = string_value(key) else {
+            continue;
+        };
+        if let Some(entry) = map_game_entry(&manifest_key, value) {
+            games.insert(manifest_key, entry);
+        }
+    }
+
+    Ok(ManifestIndex {
+        version: MANIFEST_INDEX_VERSION,
+        fetched_at,
+        source_url: source_url.to_string(),
+        games,
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::cloud_save::manifest::source::resolve_source_url;
+    use std::time::{SystemTime, UNIX_EPOCH};
+
+    #[tokio::test]
+    async fn builds_index_from_real_manifest() {
+        let source_url = resolve_source_url(None);
+
+        let raw_yaml = reqwest::get(&source_url)
+            .await
+            .unwrap()
+            .error_for_status()
+            .unwrap()
+            .text()
+            .await
+            .unwrap();
+
+        let fetched_at = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis() as i64;
+
+        let index = build_manifest_index(&raw_yaml, &source_url, fetched_at).unwrap();
+
+        let example = index
+            .games
+            .values()
+            .flat_map(|game| game.files.iter().map(move |file| (game, file)))
+            .find(|(_, file)| !file.tags.is_empty() || !file.when.is_empty())
+            .expect("no file with tags or conditions found");
+
+        println!("{}", serde_json::to_string_pretty(&example).unwrap());
+
+        assert_eq!(index.version, 1);
+        assert_eq!(index.source_url, source_url);
+        assert_eq!(index.fetched_at, fetched_at);
+        assert!(!index.games.is_empty());
+    }
+}
--- a/native/hydra-native/src/cloud_save/manifest/lookup.rs
+++ b/native/hydra-native/src/cloud_save/manifest/lookup.rs
@@ -1,0 +1,67 @@
+use unicode_normalization::{char::is_combining_mark, UnicodeNormalization};
+
+use super::types::{ManifestGameEntry, ManifestIndex};
+
+fn normalize_manifest_key(value: &str) -> String {
+    value
+        .nfkd()
+        .filter(|character| !is_combining_mark(*character) && character.is_alphanumeric())
+        .flat_map(char::to_lowercase)
+        .collect()
+}
+
+pub fn find_manifest_entry<'a>(
+    index: &'a ManifestIndex,
+    object_id: &str,
+    remote_id: Option<&str>,
+    title: Option<&str>,
+) -> Option<&'a ManifestGameEntry> {
+    let candidates = [Some(object_id), remote_id, title]
+        .into_iter()
+        .flatten()
+        .filter(|candidate| !candidate.is_empty())
+        .collect::<Vec<_>>();
+
+    for candidate in &candidates {
+        if let Some(entry) = index.games.get(*candidate) {
+            return Some(entry);
+        }
+    }
+
+    for candidate in candidates {
+        let normalized_candidate = normalize_manifest_key(candidate);
+        if let Some(entry) = index.games.iter().find_map(|(key, entry)| {
+            (normalize_manifest_key(key) == normalized_candidate).then_some(entry)
+        }) {
+            return Some(entry);
+        }
+    }
+
+    None
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::cloud_save::manifest::cache::get_manifest_index;
+    use crate::cloud_save::manifest::source::resolve_source_url;
+    use tempfile::tempdir;
+
+    #[tokio::test]
+    async fn finds_balatro_from_real_manifest() {
+        let source_url = resolve_source_url(None);
+        let cache_directory = tempdir().unwrap();
+
+        let index = get_manifest_index(cache_directory.path(), &source_url)
+            .await
+            .unwrap();
+
+        let result = find_manifest_entry(&index, "2379780", None, Some("Balatro"))
+            .expect("Balatro should be found in the real manifest");
+
+        println!("{}", serde_json::to_string_pretty(result).unwrap());
+
+        assert_eq!(result.manifest_key, "2379780");
+        assert!(!result.files.is_empty());
+    }
+}
--- a/native/hydra-native/src/cloud_save/manifest/lookup.rs
+++ b/native/hydra-native/src/cloud_save/manifest/lookup.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use unicode_normalization::{char::is_combining_mark, UnicodeNormalization};
 
 use super::types::{ManifestGameEntry, ManifestIndex};
@@ -10,8 +12,29 @@ fn normalize_manifest_key(value: &str) -> String {
         .collect()
 }
 
+pub struct ManifestLookupIndex {
+    pub manifest: ManifestIndex,
+    normalized_games: HashMap<String, String>,
+}
+
+impl ManifestLookupIndex {
+    pub fn new(manifest: ManifestIndex) -> Self {
+        let mut normalized_games = HashMap::new();
+        for key in manifest.games.keys() {
+            normalized_games
+                .entry(normalize_manifest_key(key))
+                .or_insert_with(|| key.clone());
+        }
+
+        Self {
+            manifest,
+            normalized_games,
+        }
+    }
+}
+
 pub fn find_manifest_entry<'a>(
-    index: &'a ManifestIndex,
+    index: &'a ManifestLookupIndex,
     object_id: &str,
     remote_id: Option<&str>,
     title: Option<&str>,
@@ -23,16 +46,18 @@ pub fn find_manifest_entry<'a>(
         .collect::<Vec<_>>();
 
     for candidate in &candidates {
-        if let Some(entry) = index.games.get(*candidate) {
+        if let Some(entry) = index.manifest.games.get(*candidate) {
             return Some(entry);
         }
     }
 
     for candidate in candidates {
         let normalized_candidate = normalize_manifest_key(candidate);
-        if let Some(entry) = index.games.iter().find_map(|(key, entry)| {
-            (normalize_manifest_key(key) == normalized_candidate).then_some(entry)
-        }) {
+        if let Some(entry) = index
+            .normalized_games
+            .get(&normalized_candidate)
+            .and_then(|key| index.manifest.games.get(key))
+        {
             return Some(entry);
         }
     }
@@ -45,7 +70,39 @@ mod tests {
     use super::*;
     use crate::cloud_save::manifest::cache::get_manifest_index;
     use crate::cloud_save::manifest::source::resolve_source_url;
+    use crate::cloud_save::manifest::types::ManifestGameEntry;
+    use indexmap::IndexMap;
     use tempfile::tempdir;
+
+    #[test]
+    fn normalized_lookup_preserves_exact_matches_and_manifest_order() {
+        let entry = |manifest_key: &str| ManifestGameEntry {
+            manifest_key: manifest_key.to_string(),
+            files: vec![],
+        };
+        let mut games = IndexMap::new();
+        games.insert("Pokémon".to_string(), entry("Pokémon"));
+        games.insert("Pokemon".to_string(), entry("Pokemon"));
+        let index = ManifestLookupIndex::new(ManifestIndex {
+            version: 1,
+            fetched_at: 0,
+            source_url: "test".to_string(),
+            games,
+        });
+
+        assert_eq!(
+            find_manifest_entry(&index, "missing", None, Some("Pokemon"))
+                .unwrap()
+                .manifest_key,
+            "Pokemon"
+        );
+        assert_eq!(
+            find_manifest_entry(&index, "missing", None, Some("POKEMON"))
+                .unwrap()
+                .manifest_key,
+            "Pokémon"
+        );
+    }
 
     #[tokio::test]
     async fn finds_balatro_from_real_manifest() {
--- a/native/hydra-native/src/cloud_save/manifest/rules.rs
+++ b/native/hydra-native/src/cloud_save/manifest/rules.rs
@@ -1,0 +1,121 @@
+use super::types::{
+    CloudSaveGameId, CloudSaveRule, CloudSaveRuleCondition, GameSaveRules, ManifestFileRule,
+    ManifestGameEntry,
+};
+use crate::cloud_save::identity::build_rule_id;
+
+const FILE_RULE_KIND: &str = "file";
+const DIRECTORY_RULE_KIND: &str = "dir";
+const LUDUSAVI_RULE_SOURCE: &str = "ludusavi";
+
+pub(crate) fn infer_rule_kind(raw_path: &str) -> &'static str {
+    if raw_path
+        .chars()
+        .any(|character| matches!(character, '*' | '?' | '[' | '{' | ']'))
+    {
+        return FILE_RULE_KIND;
+    }
+    if raw_path.ends_with('/') {
+        return DIRECTORY_RULE_KIND;
+    }
+    let base_name = raw_path.rsplit('/').next().unwrap_or(raw_path);
+    if base_name.contains('.') {
+        FILE_RULE_KIND
+    } else {
+        DIRECTORY_RULE_KIND
+    }
+}
+
+fn build_rule(file: &ManifestFileRule) -> CloudSaveRule {
+    let mut rule = CloudSaveRule {
+        rule_id: String::new(),
+        kind: infer_rule_kind(&file.raw_path).to_string(),
+        raw_path: file.raw_path.clone(),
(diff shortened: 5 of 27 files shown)


Moyasee (commented):
Read through the whole slice. The architecture is good: clean module split, well factored path resolution, the Wine/Proton/Steam root layering is thought through and has realistic tests, and the `<storeUserId>` capture-and-group design is genuinely clever (the `is_shared_scan_root` trick correctly keeps `GraphicsConfig.xml` out of the Elden Ring rule). Custom paths are validated properly and the scanner is correctly moved off the async runtime with `spawn_blocking`.

One data-loss bug though, plus a few things worth doing before this ships.

**`save_scanner/scan_rules.rs:112` — a wildcard rule matching several directories keeps only one of them**

For non-user-bound rules the selection is `selected_root = is_shared_scan_root || !selected_concrete_root`. The first concrete root sets `selected_concrete_root = true`, so every later matched directory in the same loop gets `selected_root = false` and never reaches `candidate_selected_paths`. `scan_path.rs:180` produces one `ScannedCloudSavePath` per matched directory (`scanned_by_root` is keyed by resolved path), so a rule like `<base>/XCOM/game_*` matching `game_1`, `game_2`, `game_3` discovers only the lexicographically first one and silently drops the rest.

I pulled the live manifest to size this: 310 rules have that shape (`<base>/Profiles/Profile*`, `<base>/Saves/Replay-*`, `<winDocuments>/My Games/Overlord II/Save - *`, and so on). File-globs like `<base>/Profiles/*/*.cfg` are fine because they group under the shared scan root, so the fix is narrow: treat every concrete root a single candidate produces as selected. The current tests only exercise multi-root selection through the `<storeUserId>` path, which takes the other branch, so this needs its own test.

This is adjacent to greptile's `scan_rules.rs:143` comment but a separate mechanism, and it bites earlier: the dedup only matters for roots that survive selection, and here they do not.

**No containment check on manifest-supplied paths, and symlinks are followed**

Nothing between `manifest/rules.rs::build_rule` and `WalkDir` constrains where a rule may point. A manifest entry of `files: {"<home>/../..": ...}` resolves to `/home/user/../..`, `canonicalize` makes that `/`, and `scan_directory` enumerates to depth 100 and hands every absolute path back to JS, which a later slice uploads. Banning `..` outright is not an option since the real manifest uses it legitimately (25 entries, e.g. `<winLocalAppData>/../LocalLow/...`), so this wants canonicalize-then-verify against an allow-list of roots (home, appdata, install dir, prefix, steam).

Separately `follow_links` is `true` for every non-`<custom>` rule (`scan_rules.rs:26`, `scan_path.rs:79`), so a symlink dropped into a save directory by a mod, a shared save archive, or a previously restored cloud save makes the scanner walk and report files anywhere on disk. The custom-path branch already handles this correctly (`can_refuse_symlinks_that_leave_a_custom_root`); manifest rules do not.

**`manifest/mod.rs:20` — 22 MB of YAML downloaded unbounded and parsed on the napi runtime thread**

`response.text()` has no size cap, only the 30s timeout. Then `serde_yaml_ng::from_str` builds a full `Value` tree for 38,890 rules and `serde_json::to_vec_pretty` re-serializes the index, all inline in `get_save_rules_for_game`'s future. That blocks a tokio worker for seconds and spikes a few hundred MB. `save_scanner/mod.rs:17` already does this right; the manifest path should use `spawn_blocking` too, plus a byte limit on the download.

**`save_scanner/glob.rs:99` — `expand_braces` is exponential and recursive on untrusted input**

Each brace group doubles the output, so `{a,b}{a,b}...` gives 2^n patterns and each one gets run through a filesystem glob. Recursion depth equals the number of groups, and a stack overflow is not caught by `spawn_blocking` — it takes the process down. `is_brace_syntax` also rescans from index 0 per character, making each level O(n^2). No cap on expansion count, pattern length, or depth. The longest real manifest path is 196 chars so this is a hostile-manifest or hostile-`sourceUrl` vector rather than a live bug, but it is cheap to bound.

**`path_resolution/capture.rs:81` and `:103` — two byte-index panics in `capture_segment`**

`&value[..prefix.len()]` panics when `value` is non-ASCII and the byte index lands mid-codepoint, which is reachable because `**` in a template lets `match_segments` test the capture against arbitrary path components: `capture_segment("_steam___HYDRA..__", "日本語です", true)` gives "byte index 7 is not a char boundary". The second one is worse in a quiet way: `haystack` is `value[capture_start..].to_lowercase()` but `haystack.find(needle)` is used as an index into the original `value`, so any length-changing lowercase corrupts the offset. `capture_segment("__HYDRA...___data", "ẞé_data", false)` panics, and the non-panicking cases silently produce a wrong capture. `spawn_blocking` keeps it from crashing the process but it fails the whole scan batch.

Smaller things:

- `save_scanner/glob.rs:117` — brace expansion changes manifest semantics. Ludusavi and `globetter` treat `{}` literally; a group without a comma returns `Err("brace must contain alternatives")` and aborts the candidate with `cloud_save_invalid_glob`. Six live entries hit this today, including `{home}/unity3d/HOF Studios/Depth of Extinction` and paths containing `{random}`, `{id}`, `{64bitSteamID}`. They should be literal components, not scan failures.
- `save_scanner/scan_path.rs:83` — `entry.map_err(...)?` inside the `WalkDir` loop turns any per-entry error (EACCES on one subdirectory, or a symlink cycle now that `follow_links` is on) into `outcome: "failed"` with zero files for the whole candidate, instead of returning what was readable. These belong in `warning_codes` with the walk continuing.
- `save_scanner/scan_path.rs:37` — `Path::new(pattern)` with no absolute-path requirement, so relative manifest paths glob against the Electron process CWD. 140 live entries have no leading token, including `Saves are located within a profile subfolder`, `HKEY_CURRENT_USER/Software/...`, and `C:/DIG/SAVEGAME.*` (relative when running on Linux).
- `path_resolution/resolve_rules.rs:14` — when `raw_path` starts with `<custom>`, the resolved path becomes `preferred_path.replace('\\', "/")` verbatim and `unresolved_tokens` is cleared. None of `custom.rs`'s validation is re-applied and the value is not glob-escaped, so a stored path containing `*` or `[` silently becomes a glob.
- `manifest/source.rs:3` — any `sourceUrl` string is accepted, so an `http://` manifest is fetched in the clear and the MITM-able body is written to the on-disk cache where it is trusted for 24h. Worth restricting to https, ideally to a known host.
- `manifest/cache.rs:30` — `.expect()` on the HTTP client build inside an async N-API function. napi is built without `catch-unwind` (`Cargo.toml:17`), so a panic there leaves the JS promise unsettled rather than rejecting. (I traced the sync `resolve_save_rules` export separately for the same class of problem and found nothing reachable; the slicing there is all guarded by `starts_with`/ASCII checks and `to_ascii_lowercase`, which preserves offsets.)
- `indexer.rs:101`, `lookup.rs:50`, `rules.rs:86`, `resolve_path.rs:342` all hit `cdn.losbroxas.org` inside `#[tokio::test]`, so `cargo test` fails offline and `resolves_real_balatro_save_path_with_wine` asserts on an exact upstream rule string. A checked-in fixture would fix both. Related: no CI workflow runs `cargo test` or `cargo clippy` today, so none of these tests run anywhere.

Nits:

- `resolve_path.rs:94` — `glob_base_path("/a*/b")` returns `""`, and `assign_scan_roots` then treats that as a prefix of every absolute candidate. Unreachable today since no live entry globs the first segment.
- `scan_rules.rs:10` — returns `Result<_, String>` but the closure can never produce `Err`; the error channel and the `map_err` at `save_scanner/mod.rs:20` are dead.
- `cache.rs:142` — `has_matching_disk_index` is always equal to `disk_index.is_some()` after the filter on the next line.
- `path_resolution/types.rs:46` — `dynamic` is written everywhere and read nowhere; same for `variant_id`/`relative_path`/`authority` in `UserLocationCoverage`, which are hardcoded.
- `manifest/rules.rs:14` vs `identity/mod.rs:61` — the glob-character sets disagree, `]` is in one and not the other.
- `context.rs:137` — on macOS the `<xdgData>` fallback is `<home>/Library`, but Ludusavi defines it as `<home>/Library/Application Support`. Only hit when `appDataDir` is absent. `<xdgData>` on Linux also ignores `$XDG_DATA_HOME`.

Hachi-R:
This feature is being submitted as a stacked series of pull requests. Each PR intentionally contains only one layer of the final implementation and must be reviewed against its direct base branch, not as a standalone complete feature.

1. **#2606 — Native discovery**  
   Only manifest loading, path resolution, and local save discovery belong here. Missing snapshots, uploads, restores, sync lifecycle, custom-path UI, and user-facing consumers are expected.

2. **#2607 — Native snapshots**  
   Adds snapshot construction on top of discovery. Network transfer and synchronization policy are intentionally absent.

3. **#2608 — Native transfer**  
   Adds upload, download, and restore primitives. Production lifecycle integration is introduced later.

4. **#2610 — Sync core and lifecycle**  
   Connects discovery, snapshots, and transfers to manual and automatic synchronization. V1/V2 exclusivity and launch lifecycle behavior belong here.

5. **#2611 — Desktop UI**  
   Adds the desktop-facing controls and presentation. Big Picture behavior is intentionally not included.

6. **#2612 — Variants, validation, and custom paths**  
   Adds account variants, custom-path validation, approval, rebinding, overlap handling, and related safeguards. Complaints about custom paths being incomplete in earlier PRs should be evaluated here instead.

7. **#2613 — Cloud-save deletion**  
   Adds remote and local cloud-save deletion behavior without changing the previous synchronization architecture.

8. **#2614 — Big Picture**  
   Adds the Big Picture presentation and integration while reusing the functionality introduced by the earlier PRs.

When reviewing an individual PR, please report bugs introduced by that slice or broken contracts with its existing base. Do not report functionality intentionally implemented by a later PR as missing from an earlier one. However, concrete data-loss, crash, or security issues inside the current slice remain valid even if the full feature is stacked.

Moyasee (commented):
Reviewed against `main` per your stacking comment, so nothing here is "missing feature from a later slice". Built the crate and ran the suite locally; the findings I could reproduce have their output inline.

The Wine/Proton resolution matrix is the strongest part of this. `resolve_path.rs` handles the launcher-prefix vs compatdata split carefully, and those tests read as real scenarios rather than coverage padding. `local_id` length-prefixing its parts and the atomic index write with stale-data fallback are both the right calls.

Three things I think need to land before merge, plus one cheap perf fix.

### 1. Two tests fail on a clean run

```
finds_balatro_from_real_manifest ............... FAILED
builds_balatro_save_rules_from_real_manifest ... FAILED
test result: FAILED. 48 passed; 2 failed
```

Both fail with `operation timed out` on the manifest fetch. Four tests download the full manifest inside `cargo test`. That file is 22.7 MB and takes about 17s to pull on a good connection, against a 30s client timeout, so running them in parallel tips two over. They also assert on third-party data holding still (Balatro at 2379780 keeping a `<winAppData>/Balatro` rule). `#[ignore]` or a trimmed fixture would fix both problems.

### 2. `cargo test --lib` does not link on macOS

No `.cargo/config.toml`, so the test harness fails with undefined `_napi_delete_reference` and `_napi_reference_unref`. I needed `RUSTFLAGS="-C link-arg=-undefined -C link-arg=dynamic_lookup"` to run anything. This crate had no tests before, so it is newly relevant. Worth adding the config so the suite is runnable on all three dev platforms.

### 3. CI never runs any of it

`build.yml` and `release.yml` install Rust and build the addon, but there is no `cargo test` or `cargo clippy` step. 50 new tests land ungated.

### Measured numbers

Release build against the real manifest, 48,639 games:

```
parse+index:                     465ms
one ManifestIndex clone:          14ms
exact-hit lookup:                 17us
miss lookup (normalized scan):     7ms
```

Details in the inline comments on `cache.rs` and `lookup.rs`.

Blocking in my view: the `when` filtering question, the cross-root dedup, and the failing tests. The `Arc<ManifestIndex>` change is two lines for a large payoff and worth folding in now.

native/hydra-native/src/cloud_save/path_resolution/resolve_rules.rs
```diff
@@ -0,0 +1,82 @@
...
+use super::types::{PathResolutionContext, ResolvedCloudSaveRule};
+
+pub fn resolve_rules(
+    rules: Vec<CloudSaveRule>,
+    context: &PathResolutionContext,
+) -> Vec<ResolvedCloudSaveRule> {
+    rules
+        .into_iter()
+        .map(|rule| {
+            let mut resolved = resolve_path(&rule.raw_path, context);
```
Moyasee, line 13: Manifest `when` conditions are never evaluated. Every rule goes straight through `resolve_path` and `rule.when` is only copied onto the output. Ran it against a Linux/Steam context:

```
epic-only  when=[(None, Some("epic"))] -> resolved=["/games/Game/EpicSaves"]
mac-only   when=[(Some("mac"), None)]  -> resolved=["/home/hydra/MacSaves"]
```

OS conditions are masked by accident a lot of the time, since `<winAppData>` and friends collapse to `<skip>` off Windows and get dropped by `has_unresolved_placeholder`. But `store` conditions have no backstop at all, and neither do `<base>`, `<root>`, or `<home>` paths, which resolve on every platform.

Since `when` is forwarded onto both `ResolvedCloudSaveRule` and `ScannedCloudSaveRule`, this may be a deliberate "the consumer filters" contract. If it is, worth saying so somewhere, because as it stands `scanResolvedSaveRules` reports save data for storefronts the user does not own.

native/hydra-native/src/cloud_save/path_resolution/resolve_rules.rs
```diff
@@ -0,0 +1,82 @@
...
+            if rule.raw_path.starts_with("<custom>") {
+                if let Some(preferred_path) = &rule.preferred_path {
```
Moyasee, line 15: `preferred_path` is taken verbatim and becomes the sole resolved path, without going through `decode_custom_path`, which is right next door doing exactly this validation.

I get that approval and validation are #2612's job, so not asking you to move it forward. A note here would keep the gap deliberate rather than something that gets forgotten between slices.

Minor: this hardcodes `"<custom>"` rather than using `CUSTOM_PATH_PREFIX`, and `scan_rules.rs:26` does the same.

native/hydra-native/src/cloud_save/save_scanner/scan_rules.rs
```diff
@@ -0,0 +1,608 @@
...
+                    }
+                    selected_paths.extend(candidate_selected_paths);
+
+                    if !user_bound && candidate_has_files {
+                        let mut seen_absolute_paths = HashSet::new();
+                        let mut seen_relative_paths = HashSet::new();
+                        for scanned in &mut selected_paths {
+                            scanned.files.retain(|file| {
+                                seen_absolute_paths.insert(file.absolute_path.clone())
+                                    && seen_relative_paths.insert(file.relative_path.clone())
```
Moyasee, line 142: `seen_relative_paths` is shared across roots, but relative paths are only unique *within* a root, so a file gets dropped when a different selected root already used that name. Two roots each holding their own `save.dat`:

```
root=.../a  files=[("save.dat", ".../a/save.dat")]
root=.../b  files=[]
TOTAL FILES KEPT: 1
```

Keying on `(resolved_path, relative_path)` fixes it and still preserves what the dedup is for.

Scope note so this is not overstated: I could only reach it through the public `scanResolvedSaveRules` boundary, which takes arbitrary `ResolvedCloudSaveRule` from JS. I could not get there from `resolveSaveRules` output, because `glob_base_path` always yields a scan root covering every match. So today it is an API-surface bug rather than live data loss, but the napi function is public and #2610 feeds it.

Separately, `seen_absolute_paths.insert(..) && seen_relative_paths.insert(..)` short-circuits, so a repeated absolute path skips the relative-set update.

native/hydra-native/src/cloud_save/save_scanner/scan_rules.rs
```diff
@@ -0,0 +1,608 @@
...
+
+            if rule.unresolved_tokens.is_empty() {
+                for candidate in &rule.resolved_paths {
+                    let template = capture_template(&rule.raw_path, &candidate.path);
+                    let mut scanned_paths = match scan_resolved_path_with_capture(
+                        &candidate.path,
+                        candidate.case_sensitive,
+                        candidate.scan_root.as_deref(),
+                        template.as_deref(),
+                        !rule.raw_path.starts_with("<custom>"),
```
Moyasee, line 26: `follow_links` is `!raw_path.starts_with("<custom>")`, so community-maintained manifest rules follow symlinks while the user's own custom paths do not. Given #2608 uploads whatever comes out of here, I would expect the opposite bias. Is this deliberate?

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,442 @@
...
+        case_sensitive,
+        require_literal_separator: true,
+        require_literal_leading_dot: false,
+        follow_links,
+    }
+}
+
+fn glob_matches(pattern: &str, options: MatchOptions) -> Result<Vec<PathBuf>, String> {
+    let direct_path = Path::new(pattern);
+    if direct_path.exists() {
```
Moyasee, line 197: `case_sensitive: false` is a no-op for any literal path. This `exists()` early return and the `!has_glob_pattern` branch below both bypass `MatchOptions` entirely, and `case_sensitive` only ever reaches `globetter::glob_with`.

Wine partly survives because of the `drive_*/users/*` variant, which does go through the glob path. But every `steam_proton_paths` candidate is fully literal, so a Proton prefix with different on-disk casing silently finds nothing on ext4 or btrfs, which is the exact case the flag exists for.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,442 @@
...
+
+    Some((canonical_root, normalize_path(&relative.to_string_lossy())))
+}
+
+fn scan_directory(root: &Path, follow_links: bool) -> Result<ScannedCloudSavePath, String> {
+    let mut files = Vec::new();
+    let traversal_root = std::fs::canonicalize(root)
+        .map_err(|error| format!("cloud_save_filesystem_error: {error}"))?;
+
+    for entry in WalkDir::new(&traversal_root)
```
Moyasee, line 258: Nothing bounds what this enumerates. Depth is capped at 100 but there is no file-count or byte cap, and every entry is collected into memory. A manifest rule resolving to `<home>`, or `<base>` on a 100 GB install, hangs or balloons.

The manifest is community-editable and #2608 uploads what this returns, so the discovery layer seems like the right place for a cap and a root denylist rather than pushing it downstream.

(Small thing: `MAX_SCAN_DEPTH = 100` actually yields 99 levels of files, which your own test asserts. Just a naming mismatch.)

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,442 @@
...
+        let scanned =
+            scan_resolved_path(&pattern, true, Some(&temp.path().display().to_string())).unwrap();
+
+        assert_eq!(scanned[0].files.len(), 1);
+        assert_eq!(scanned[0].files[0].relative_path, "{Deluxe}/save.dat");
+    }
+
+    #[test]
+    fn preserves_every_directory_matched_by_one_glob() {
+        let temp = tempfile::tempdir_in(".").unwrap();
```
Moyasee, line 565: `tempdir_in(".")` drops temp directories into `native/hydra-native/`, so a panic before cleanup leaves them in the working tree. If this is dodging the macOS `/tmp` symlink, a one-line comment would save the next person from "fixing" it back to `tempdir()`.

native/hydra-native/src/cloud_save/save_scanner/glob.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    let mut expanded = Vec::new();
+    for alternative in alternatives {
+        let candidate = format!(
+            "{}{}{}",
+            &pattern[..opening],
+            alternative,
+            &pattern[closing + 1..]
+        );
+        expanded.extend(expand_braces(&candidate)?);
```
Moyasee, line 131: No cap on expansion. Roughly 20 nested two-way groups in a manifest path expands to a million patterns, each of which then gets globbed. The manifest is community-editable, so a cheap upper bound here seems worth it.

native/hydra-native/src/cloud_save/manifest/cache.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    let fallback = cached
+        .as_ref()
+        .filter(|index| index.source_url == source_url)
+        .cloned();
+    let index = load_index(user_data_path, source_url)
+        .await
+        .or_else(|error| fallback.ok_or(error))?;
+    *cached = Some(index.clone());
```
Moyasee: The whole `ManifestIndex` gets deep-cloned per lookup, here and again at line 150 on the fresh-disk-index path. Measured on a release build against the real manifest, 48,639 games:

```
parse+index:                     465ms
one ManifestIndex clone:          14ms
exact-hit lookup:                 17us
```

So a cache hit costs 14 to 28ms of pure memcpy before any actual lookup work. Resolving a 500-game library is 7 to 14 seconds of copying. Returning `Arc<ManifestIndex>` removes it outright and is about a two-line change.

native/hydra-native/src/cloud_save/manifest/cache.rs
```diff
@@ -0,0 +1,209 @@
...
+
+static CACHE_LOCKS: OnceLock<ManifestCaches> = OnceLock::new();
+static HTTP_CLIENT: OnceLock<reqwest::Client> = OnceLock::new();
+
+fn http_client() -> &'static reqwest::Client {
+    HTTP_CLIENT.get_or_init(|| {
+        reqwest::Client::builder()
+            .timeout(Duration::from_secs(MANIFEST_HTTP_TIMEOUT_SECS))
+            .build()
+            .expect("Failed to build cloud save manifest HTTP client")
```
Moyasee, line 31: `.expect()` in a napi code path. A panic here takes the Electron main process with it rather than surfacing as a JS error. `hash_json` in `identity/mod.rs:43` has the same shape.

native/hydra-native/src/cloud_save/manifest/lookup.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    for candidate in &candidates {
+        if let Some(entry) = index.games.get(*candidate) {
+            return Some(entry);
+        }
+    }
+
+    for candidate in candidates {
+        let normalized_candidate = normalize_manifest_key(candidate);
+        if let Some(entry) = index.games.iter().find_map(|(key, entry)| {
```
Moyasee: This runs NFKD plus lowercase plus an allocation across all 48k games for every candidate. Measured 7ms per miss on a release build, versus 17us for an exact hit.

The normalized keys could be computed once when the index is built. The index is already persisted to disk, so it costs nothing after the first fetch.

native/hydra-native/src/cloud_save/manifest/indexer.rs
```diff
@@ -0,0 +1,135 @@
...
+mod tests {
+    use super::*;
+    use crate::cloud_save::manifest::source::resolve_source_url;
+    use std::time::{SystemTime, UNIX_EPOCH};
+
+    #[tokio::test]
+    async fn builds_index_from_real_manifest() {
+        let source_url = resolve_source_url(None);
+
+        let raw_yaml = reqwest::get(&source_url)
```
Moyasee, line 105: This is one of four tests that download the full manifest during `cargo test`. It is 22.7 MB and takes about 17s to fetch on a good connection, against the 30s timeout in `cache.rs`, so running them in parallel makes them flake. Two failed for me on a clean run.

They also couple the suite to third-party data holding still: Balatro staying at 2379780 with a `<winAppData>/Balatro` rule. `#[ignore]`, or a checked-in trimmed fixture, would fix both.

native/hydra-native/src/cloud_save/manifest/source.rs
```diff
@@ -0,0 +1,33 @@
+pub const DEFAULT_CLOUD_SAVE_MANIFEST_URL: &str = "https://cdn.losbroxas.org/manifest.yaml";
+
+pub fn resolve_source_url(configured_url: Option<String>) -> String {
```
Moyasee, line 3: No scheme check on the caller-supplied URL, so `http://` or `file://` both go through. The caller is the main process so severity is low, but an https allowlist here is one line and the result drives filesystem scanning.

native/hydra-native/src/cloud_save/manifest/rules.rs
```diff
@@ -0,0 +1,121 @@
...
+    ManifestGameEntry,
+};
+use crate::cloud_save::identity::build_rule_id;
+
+const FILE_RULE_KIND: &str = "file";
+const DIRECTORY_RULE_KIND: &str = "dir";
+const LUDUSAVI_RULE_SOURCE: &str = "ludusavi";
+
+pub(crate) fn infer_rule_kind(raw_path: &str) -> &'static str {
+    if raw_path
```
Moyasee, line 12: The glob-char check runs before the trailing-slash check, so `Profiles/*/` classifies as `"file"`. That feeds `target_semantics` into the rule id hash, and the id is meant to be stable and referenced by later slices, so it seems worth getting right now rather than after ids are in the wild.

native/hydra-native/src/cloud_save/path_resolution/context.rs
```diff
@@ -0,0 +1,186 @@
...
+            input.platform
+        ));
+    }
+
+    let home_dir = normalize_separators(&input.home_dir);
+    let install_dir = install_directory(input.executable_path.as_deref());
+    let wine_prefix_path = input.wine_prefix_path.as_deref().map(normalize_separators);
+    let documents_dir = input.documents_dir.as_deref().map(normalize_separators);
+    let app_data_dir = input.app_data_dir.as_deref().map(normalize_separators);
+    let os_username = basename(&home_dir).unwrap_or_else(|| "*".to_string());
```
Moyasee, line 126: The `"*"` fallback never actually wildcards. Every consumer runs `os_username` through `literal()` or `globset::escape`, so it ends up as a path segment literally named `*` that matches nothing. Net effect is a silent empty scan instead of a reported unresolved token.

native/hydra-native/Cargo.toml
```diff
@@ -7,13 +7,31 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
+anyhow = "1.0"
+blake3 = "=1.8.5"
```
Moyasee, line 11: `blake3`, `globetter`, `globset`, `walkdir`, and `known-folders` are pinned with `=` while everything else uses caret ranges. Intentional, or leftover from pinning down a build issue?

(Checked `globetter` since it is an unusual name: it is mtkennerly's glob fork, same author as the ludusavi manifest. No concern there.)

Hachi-R:
Status of the review points:

### Done

- Preserved every directory matched by wildcard rules when they share the same scan root, including the profile directory in the relative path.
- Made store-user capture Unicode-safe.
- Kept literal braces literal while still expanding real brace alternatives such as `{sav,dat}`.
- Rejected relative manifest paths.
- Fixed literal paths with `case_sensitive: false`: exact matches remain the fast path, with a case-insensitive fallback when the exact casing does not exist.
- Removed the repeated normalized lookup scan: normalized keys are indexed once, cached through `Arc`, exact matches still take precedence, and normalized collisions retain manifest insertion order.

### Not done

- `when` filtering: intentionally unchanged. The Ludusavi manifest specification allows tools to scan entries even when the constraints are not applicable, so treating `when` as a mandatory filter could discard valid Wine/Proton and cross-platform saves.
- `preferred_path` / custom-path validation: belongs to the custom-path PR (#2612), not this native-discovery PR.
- Direct-NAPI cross-root deduplication edge case: unchanged. The production resolver groups these matches under a shared scan root, so the normal resolve-to-scan pipeline is not affected.
- Symlink policy: unchanged. Manifest rules follow symlinks; custom-path restrictions are handled by the later custom-path flow.
- Scan file/byte caps and root denylist: not added; these are separate product/security policy changes.
- Brace-expansion cap: not added; this was considered theoretical hardening outside the current scope.
- Source URL HTTPS allowlist: not added; changing the internal manifest-source override policy is outside this PR.
- Serialization `.expect()` / panic hardening: not changed; this is defensive hardening rather than a demonstrated discovery bug.
- Wildcard directory-kind inference and `os_username = *` behavior: not changed in this pass.
- Hardcoded custom placeholder constant and dependency pinning comments: not changed.
- Test-only `tempdir_in(.)` comment: not changed.
- Live-manifest test isolation, macOS linker configuration, and CI coverage comments: not changed; no test or CI infrastructure changes were included.

Moyasee (commented):
Re-reviewed at c199642. Rebuilt and re-ran the suite, and measured the manifest changes against the same 48,639-game manifest as last time.

### The two manifest fixes are good

| | before | after |
|---|---|---|
| cache-hit clone | 14ms | ~4ns (`Arc::clone`) |
| exact-hit lookup | 17us | 1.9us |
| miss lookup | 7ms | 1.6us |
| one-time index build | n/a | +5.8ms |

Exactly the right shape, and paying 5.8ms once to delete 7ms per miss is a good trade. Keeping manifest insertion order for normalized collisions is a nice touch, and the `Pokémon` / `Pokemon` test pins the behavior properly.

### The case-sensitivity fix does not do anything

I built a case-sensitive APFS volume to test on a filesystem that behaves like ext4 rather than trusting default macOS behavior. `MatchOptions.case_sensitive` only applies to pattern components that actually contain a wildcard. Literal components get concatenated and matched by the OS, whether they sit before or after a wildcard. Details and measurements inline on `scan_path.rs`.

I also got something wrong in my first review and want to correct it: I said Wine partly survived through the `drive_*/users/*` variant. It does not. The gap is wider than I described, not narrower.

### The new test cannot pass on any platform

```
test result: FAILED. 51 passed; 1 failed
panicked at src/cloud_save/save_scanner/scan_path.rs:309:9:
assertion failed: sensitive.is_empty()
```

Two different failure modes depending on the filesystem, no platform where it is green. Inline.

That also sharpens the CI point from last time: there is now a failing test in the tree and nothing catches it.

### Points I am withdrawing

Your `when` rationale is right and I should have checked the ludusavi spec before raising it. Ludusavi does scan entries whose constraints do not apply and leans on the path simply not existing, so mandatory filtering would drop legitimate Proton and cross-platform saves. Dropping that one entirely.

The cross-root dedup scoping is fair as well. The production resolver does group those under a shared scan root, which matches what I found when I tried to reach it from `resolveSaveRules`. The remaining deferrals to #2612 and to product policy all read as reasonable calls to me.

One live datapoint on the network tests, since it is easy to dismiss as my connection: across three runs this session they failed twice and passed once, same code, no changes in between.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,465 @@
...
+        follow_links,
+    }
+}
+
+fn glob_matches(pattern: &str, options: MatchOptions) -> Result<Vec<PathBuf>, String> {
+    let direct_path = Path::new(pattern);
+    if direct_path.exists() {
+        return Ok(vec![direct_path.to_path_buf()]);
+    }
+    if !has_glob_pattern(pattern) && options.case_sensitive {
```
Moyasee, line 200: This does not have the intended effect. Falling through to `globetter` does not buy case-insensitive matching for a literal path, because `MatchOptions.case_sensitive` only applies to components that contain a wildcard. Everything else is concatenated and handed to the OS.

Measured on a case-sensitive APFS volume, disk holds `Prefix/drive_c/users/steamuser/AppData/Roaming/Balatro/Save.DAT`:

```
   0 matches  <- all literal, wrong case
   1 matches  <- all literal, right case
   0 matches  <- correct-case prefix + wildcard + wrong-case literals after
   1 matches  <- correct-case prefix + wildcard + correct-case literals after
```

Third row is the important one: the wrong-cased literals sit *after* `drive_*` and `users/*` and still fail, so this is not just about the leading prefix.

End to end on the real candidate shape `resolve_save_rules` emits:

```
prefix/drive_*/users/*/AppData/Roaming/Balatro
  1 root, 1 file   <- user whose dir is AppData/Roaming
  1 root, 1 file   <- after adding a second user with appdata/roaming, still missed
```

I also owe you a correction on my first review: I said Wine partly survived through the `drive_*/users/*` variant. That was wrong. Only `drive_*` and `users/*` case-fold, and the `AppData/Roaming/Game` tail stays case-sensitive, so no candidate shape has ever matched case-insensitively. The gap is wider than I originally described.

A fix that actually works means walking the path component by component and doing a case-insensitive directory scan at each level when the exact casing misses, which is roughly what Wine does internally. That is more than a one-line change, so it may be worth splitting out rather than holding this PR.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,465 @@
...
+            .join("users")
+            .join("steamuser")
+            .join("save.dat")
+            .display()
+            .to_string();
+        let insensitive = scan_resolved_path(&requested, false, None).unwrap();
+        let sensitive = scan_resolved_path(&requested, true, None).unwrap();
+
+        assert_eq!(insensitive[0].files.len(), 1);
+        assert!(sensitive.is_empty());
```
Moyasee, line 483: This test cannot pass on either platform it is gated to.

On macOS, APFS is case-insensitive by default, so the `direct_path.exists()` fast path returns a hit even with `case_sensitive: true` and this assert fails:

```
test result: FAILED. 51 passed; 1 failed
panicked at src/cloud_save/save_scanner/scan_path.rs:309:9:
assertion failed: sensitive.is_empty()
```

On a case-sensitive filesystem the failure moves to the line above instead. I measured `scan_resolved_path(requested, false, None)` returning 0 roots there, which makes `insensitive[0]` an index-out-of-bounds panic.

Once the underlying matching is fixed, this test needs a case-sensitivity guard rather than a `target_os` gate. `target_os = "macos"` is not a proxy for a case-sensitive filesystem, and on Linux it is not guaranteed either. Probing it at runtime (write `X`, check whether `x` resolves) and skipping otherwise would be more honest than either gate.

native/hydra-native/src/cloud_save/manifest/lookup.rs
```diff
@@ -0,0 +1,124 @@
...
+        .collect()
+}
+
+pub struct ManifestLookupIndex {
+    pub manifest: ManifestIndex,
+    normalized_games: HashMap<String, String>,
+}
+
+impl ManifestLookupIndex {
+    pub fn new(manifest: ManifestIndex) -> Self {
```
Moyasee, line 21: This is the right fix. Measured against the real manifest, 48,639 games, release build:

```
ManifestLookupIndex::new:  5.8ms   (one time)
exact-hit lookup:          1.9us   (was 17us)
miss lookup:               1.6us   (was 7ms)
```

Paying 5.8ms once to delete 7ms per miss is a good trade. Preserving manifest insertion order on normalized collisions via `or_insert_with` is the right call too, and the `Pokémon` / `Pokemon` test pins it.

Only micro-note, take it or leave it: `normalized_games` stores a cloned original key, so every key is held twice. `IndexMap` supports index-based access, so a `HashMap<String, usize>` plus `get_index` would drop one copy. Marginal against a 22.7 MB manifest, so entirely reasonable to ignore.

native/hydra-native/src/cloud_save/manifest/cache.rs
```diff
@@ -0,0 +1,215 @@
...
+pub async fn get_manifest_index(
+    user_data_path: &Path,
+    source_url: &str,
+) -> Result<Arc<ManifestLookupIndex>> {
+    let cache = cache_for(user_data_path.to_path_buf()).await;
+    let mut cached = cache.lock().await;
+
+    if let Some(index) = cached.as_ref() {
+        if index.manifest.source_url == source_url && !is_index_expired(&index.manifest, now_ms()) {
+            return Ok(Arc::clone(index));
```
Moyasee, line 200: Confirmed this removes the per-lookup clone. 14ms down to about 4ns for a cache hit.

Small leftover, not a blocker: `load_index` still returns `ManifestIndex` by value and does two full clones internally (line 147, and the fresh-disk-index return just below it), so a cold start still pays roughly 28ms. That is once per process or per TTL expiry rather than per lookup, so the hot path is genuinely fixed. Mentioning it only so it is a known cost rather than a surprise later.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
Moyasee: Correction on this one. I said Wine partly survived through the `drive_*/users/*` variant. That is wrong: only `drive_*` and `users/*` case-fold, and the `AppData/Roaming/Game` tail stays literal and case-sensitive, so no candidate shape has ever matched case-insensitively. Verified on a case-sensitive volume. Wrote it up with the measurements on line 42 of the current head.

native/hydra-native/src/cloud_save/manifest/cache.rs
Moyasee: Confirmed fixed at c199642. Cache-hit clone went from 14ms to about 4ns. Left a note on the current line about the two clones still inside `load_index`, which are cold-path only.

native/hydra-native/src/cloud_save/manifest/lookup.rs
Moyasee: Confirmed fixed at c199642. Miss lookup went from 7ms to 1.6us, exact hit from 17us to 1.9us, for 5.8ms of one-time index build. Good trade.

Moyasee (commented):
Re-reviewed at 2fda30e. Built a case-sensitive APFS volume again and ran the suite both there and on ordinary case-insensitive macOS.

### The matching fix works

```
case-insensitive macOS:   53 passed; 0 failed
case-sensitive volume:    25 passed; 0 failed   (save_scanner subset)
```

Both new tests actually execute on the case-sensitive volume rather than skipping, so this is verified rather than vacuous. Walking components with an exact-match fast path before falling back to a folded `read_dir` is the right shape, and it means correctly-cased paths still cost one `exists()` call. The runtime probe replacing the `target_os` gate is exactly right too.

### One new regression, and I think it blocks

A dangling symlink whose name matches a path component now aborts the entire rule scan on the `case_sensitive: false` path. Same directory, same broken link, both engines:

```
Ok(1 roots, 1 files)   <- exact casing, case_sensitive=true   (globetter)
Ok(1 roots, 1 files)   <- exact casing, case_sensitive=false  (exact fast path)
Ok(1 roots, 1 files)   <- Game/*/Saves, case_sensitive=true   (globetter)
Err(cloud_save_filesystem_error: No such file or directory)   <- Game/*/Saves, case_sensitive=false
Err(cloud_save_filesystem_error: No such file or directory)   <- wrong casing, case_sensitive=false
```

The real `Game/AppData/Saves/s.dat` exists the whole time and is found by every path except the new one. Details and the one-line fix inline on `can_descend`.

This matters because Wine prefixes are full of dangling symlinks by construction: `drive_c/users/<user>/My Documents`, `My Music`, `My Pictures` and friends point at host XDG directories, and dangle whenever the host directory does not exist. A rule resolving through `users/*/My Documents/Game` hits exactly this.

### Two engine divergences

`case_sensitive` now selects between two different matchers, so a flag about letter casing also changes matching semantics. Two cases where the same pattern and the same tree give different answers, inline on `recursive_components`.

### Everything else from the last round still stands

The macOS linker config, the live-manifest tests, and the missing `cargo test` in CI are unchanged. Not re-litigating them, just noting they are still open.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,647 @@
...
+fn match_options(case_sensitive: bool, follow_links: bool) -> MatchOptions {
+    MatchOptions {
+        case_sensitive,
+        require_literal_separator: true,
+        require_literal_leading_dot: false,
+        follow_links,
+    }
+}
+
+fn can_descend(path: &Path, follow_links: bool) -> Result<bool, String> {
```
Moyasee, line 59: This propagates on a dangling symlink and kills the whole rule scan.

`symlink_metadata` succeeds for a broken link, and when `follow_links` is true you fall through to `std::fs::metadata`, which fails `NotFound`. The `?` at both call sites in `component_matches` carries that all the way out of `scan_resolved_path_with_capture`.

Measured on a case-sensitive volume. Tree is `Game/AppData/Saves/s.dat` plus a dangling `Game/appdata`:

```
Ok(1 roots, 1 files)   <- exact casing, case_sensitive=true   (globetter)
Ok(1 roots, 1 files)   <- exact casing, case_sensitive=false  (exact fast path)
Ok(1 roots, 1 files)   <- Game/*/Saves, case_sensitive=true   (globetter)
Err(cloud_save_filesystem_error: No such file or directory)   <- Game/*/Saves, case_sensitive=false
Err(cloud_save_filesystem_error: No such file or directory)   <- wrong casing, case_sensitive=false
```

The real save is found by every path except the new one. `scan_rules` catches the `Err` and records `outcome: "failed"` with empty `scanned_paths`, so the user's saves go invisible because an unrelated broken link happened to share a directory.

This is not a contrived tree. Wine creates `My Documents`, `My Music`, `My Pictures`, `My Videos` under `drive_c/users/<user>/` as symlinks into host XDG directories, and they dangle whenever the host directory is absent. A rule resolving through `users/*/My Documents/Game` walks straight into it, and `case_sensitive: false` is the Wine path specifically.

A metadata failure here means "cannot descend", not "abort the scan":

```rust
Ok(std::fs::metadata(path).map(|metadata| metadata.is_dir()).unwrap_or(false))
```

`follow_links: false` already behaves correctly, since the symlink check returns early before the `metadata` call.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,647 @@
...
+        let path = entry.path();
+        if is_last || can_descend(&path, follow_links)? {
+            matches.push(path);
+        }
+    }
+
+    Ok(matches)
+}
+
+fn recursive_components(
```
Moyasee, line 129: Two places where this diverges from globetter on the same input, which is worth knowing about since `case_sensitive` now picks the engine and callers set it for reasons that have nothing to do with matching semantics.

Symlink loop under `**`. WalkDir reports the loop as an error and the `?` propagates, where globetter absorbs it:

```
a/**/*.dat  with a/b/loop -> a
  case_sensitive=true  -> Ok(0)
  case_sensitive=false -> Err(cloud_save_filesystem_error: File system loop found: ...)
```

Trailing `**` matches a different set. Tree is `a/top.dat` and `a/b/deep.dat`:

```
a/**
  globetter     roots=1  files=["deep.dat"]
  hand-rolled   roots=2  files=["deep.dat", "top.dat"]
```

Neither is obviously the wrong answer, but a rule returning different files depending on a casing flag is going to be confusing to debug later. Worth either picking one semantic deliberately or pinning both in a test so the difference is intentional.

Unrelated to the divergence: this materializes every path under a `**` into a `Vec` before the next component filters it. For `<base>/**/*.sav` over a large install that is the full directory listing held at once, which compounds the unbounded-enumeration point from my first review rather than being separate from it.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,647 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::fs;
+
+    use tempfile::tempdir;
+
+    use super::*;
+
+    fn filesystem_is_case_sensitive(root: &Path) -> bool {
```
Moyasee, line 429: This is the right call, and it does what it says. Verified both ways: on ordinary macOS the two case tests skip and the suite is 53/53, and with `TMPDIR` pointed at a case-sensitive APFS volume they actually run and the scanner subset is 25/25.

Small thought for later: a skipped test looks identical to a passing one in the output, so on the CI runners these two will silently do nothing if the workspace filesystem is case-insensitive. Not worth solving now, but if `cargo test` does get wired into CI it would be worth having the skip print a line.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
```diff
@@ -0,0 +1,647 @@
...
+        return match std::fs::metadata(direct_path) {
+            Ok(_) => Ok(vec![direct_path.to_path_buf()]),
+            Err(error) if error.kind() == ErrorKind::NotFound => Ok(Vec::new()),
+            Err(error) => Err(format!("cloud_save_filesystem_error: {error}")),
+        };
+    }
+
+    let mut matches = Vec::new();
+    for expanded in expand_braces(pattern)? {
+        if options.case_sensitive {
```
Moyasee, line 210: Confirmed this now resolves what it could not before. On a case-sensitive volume, disk holding `Prefix/drive_c/users/steamuser/AppData/Roaming/Balatro/Save.DAT`, requesting `prefix/drive_*/users/*/appdata/roaming/balatro/*.dat` finds the file, where the previous approach returned nothing.

The exact-match fast path inside `component_matches` is what makes this affordable: a correctly-cased path still costs one `exists()` and never lists a directory, and only the level where casing actually diverges pays for a `read_dir`. Good design.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
Moyasee: Confirmed fixed at 2fda30e. Verified on a case-sensitive APFS volume: the wine candidate shape now resolves the file it was missing, and both new tests actually execute there rather than skipping. Left notes on the current head about a broken-symlink regression in the new walker.

Hachi-R:
Updated status after commits `c4fcf61ca`, `c199642f4`, and `2fda30e73`:

### Implemented

- **Wildcard rules matching multiple directories:** all matched directories are now retained under their shared scan root, and profile/root segments remain part of each file's relative path.
- **Unicode capture panics and incorrect offsets:** capture boundaries are now calculated on UTF-8 character boundaries, including length-changing case folds.
- **Literal braces:** values such as `{random}` and `{64bitSteamID}` remain literal, while real alternatives such as `{sav,dat}` are still expanded.
- **Relative manifest paths:** candidates that remain relative are rejected instead of being resolved against Electron's working directory.
- **Manifest cache hot-path cloning:** cached indexes are shared through `Arc`; cache hits no longer deep-clone the full manifest.
- **Normalized manifest lookup:** normalized keys are indexed once. Exact matches still win, and normalized collisions preserve manifest insertion order.
- **Case-insensitive Wine/Proton discovery:** the scanner now walks path components and matches both wildcard and literal components case-insensitively when the exact path is absent. This includes literal tails after `drive_*` and `users/*`, such as `AppData/Roaming/Game`.
- **Case-sensitivity regression tests:** tests now probe the actual filesystem behavior at runtime and include the real Wine-shaped wildcard-plus-literal-tail case.

Validation for the final matching change: `cargo check` passed and all 7 focused `save_scanner::scan_path` tests passed.

### Reviewed and intentionally unchanged

- **Manifest `when` filtering:** no change because Ludusavi intentionally permits scanning entries whose constraints do not apply; mandatory filtering can remove valid Proton and cross-platform locations. The reviewer subsequently withdrew this finding.
- **Cross-root relative-path dedup at the raw N-API boundary:** no change because production `resolveSaveRules` output groups those matches under a shared scan root. The reviewer confirmed the normal pipeline is unaffected and withdrew the finding. The unrestricted low-level N-API edge remains documented.
- **Custom `preferred_path` validation:** deferred to #2612, which owns custom-path approval, decoding, and validation. Duplicating only part of that policy in this discovery slice would create two validation paths.
- **Manifest containment rules, symlink restrictions, scan byte/file caps, and root denylisting:** not added because these require a product/security policy defining which legitimate community-manifest locations may be rejected. They are not mechanical fixes and can change which saves users are allowed to select or discover.
- **Manifest download byte limit and moving YAML/index processing to `spawn_blocking`:** not included because this is separate resource/performance hardening rather than a correctness regression in the selected fixes. It should be designed together with download and memory limits.
- **Brace expansion depth/count limits:** not added because this only addresses a hostile-manifest scenario and requires choosing explicit limits; it is separate hardening.
- **Continuing after individual `WalkDir` errors:** not changed because returning partial scans plus warnings changes the current all-or-error result semantics and needs an explicit API decision.
- **`sourceUrl` HTTPS/host allowlist:** not added because the override is supplied by the trusted main process. Restricting it is a source-policy change, not a scanner correction.
- **N-API `.expect()` / serialization panic hardening:** not changed because it requires broader error-propagation changes and no reachable normal-input failure was demonstrated.
- **Wildcard directory `target_semantics`:** not changed because it changes generated rule IDs. That should be handled deliberately with compatibility/migration consideration rather than as a review-only mechanical edit.
- **`os_username = *` fallback:** not changed because there is no agreed cross-user fallback behavior; guessing or scanning every user would expand discovery scope.
- **Cold-path `ManifestIndex` clones:** not changed because they occur only on process start or TTL refresh; the reviewer explicitly classified the remaining ~28 ms cost as non-blocking.
- **Duplicated normalized lookup keys:** not changed because the reviewer classified the memory saving as marginal relative to the manifest size.
- **Root-level glob-base edge, dead error channel/fields, redundant cache boolean, glob-character-set mismatch, placeholder constant, dependency pinning, and XDG fallback notes:** not changed because they are unreachable cases, cleanup/nits, stack-facing metadata, or unrelated platform/version-policy work rather than demonstrated regressions in this slice.
- **Live-manifest fixtures, network-test flakiness, macOS test linker configuration, CI `cargo test`/`clippy`, and the `tempdir_in(.)` note:** not changed because this update intentionally did not expand into test or CI infrastructure work.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
Moyasee: Your status list files this under "continuing after individual `WalkDir` errors", and I think that is a different issue from the one I raised. No `WalkDir` is involved here. The failure is the plain `std::fs::metadata` call on line 45, on a `DirEntry` that `read_dir` just handed back.

The fix does not return partial scans and does not add warnings, so it needs no API decision. It makes `case_sensitive: false` behave the way `case_sensitive: true` already behaves on the identical tree, which is the divergence I was reporting.

I applied it locally to confirm rather than just asserting it:

```rust
Ok(std::fs::metadata(path)
    .map(|metadata| metadata.is_dir())
    .unwrap_or(false))
```

Results on the current head with only that change:

```
case-insensitive macOS:    54 passed; 0 failed
case-sensitive volume:     26 passed; 0 failed   (save_scanner subset)
```

And the behavior on the broken-symlink tree, all three now agreeing:

```
Ok(1 roots, 1 files)   <- Game/*/Saves, case_sensitive=true   (globetter)
Ok(1 roots, 1 files)   <- Game/*/Saves, case_sensitive=false  (walker)
Ok(1 roots, 1 files)   <- wrong casing,  case_sensitive=false (walker)
```

Before the change the bottom two were `Err(cloud_save_filesystem_error: No such file or directory)` while the real save sat in `Game/AppData/Saves` the whole time.

Nothing else in the suite moves. Your call either way, but I wanted the rationale to be about the actual code path rather than about partial-scan semantics, which this does not touch.

native/hydra-native/src/cloud_save/save_scanner/scan_path.rs
Moyasee: Splitting my own comment, because your `WalkDir` rationale does apply to half of it and I should say so.

The symlink loop under `**` is a genuine `WalkDir` error, and deciding whether to swallow it, warn, or fail really is an API-semantics call rather than a mechanical fix. Fair, withdrawing that half.

The trailing `a/**` set difference is not an error-handling question, just two matchers disagreeing on the same tree:

```
a/**   with a/top.dat and a/b/deep.dat
  globetter     roots=1  files=["deep.dat"]
  hand-rolled   roots=2  files=["deep.dat", "top.dat"]
```

Low priority and not blocking anything. Pinning it in a test whenever you next touch this file would be enough to make the choice deliberate.

Moyasee (approved):
Approving. The discovery layer holds up, and everything I could verify as a real defect in this slice has either been fixed or answered.

Fixed and confirmed on my side: the manifest cache no longer deep-clones per lookup, normalized lookup keys are built once, and case-insensitive Wine and Proton matching genuinely works now. I re-ran that last one on a case-sensitive volume rather than trusting default macOS behavior, and the wildcard-plus-literal-tail case that used to find nothing now resolves. The runtime filesystem probe in the tests is the right way to gate those.

I withdrew three of my own findings along the way. Your reading of the Ludusavi `when` semantics was correct and mine was not, and the cross-root dedup and the symlink-loop-under-`**` calls were both fair.

The deferrals to #2612 and to product policy all read as reasonable scoping to me. Scan caps and a root denylist genuinely are policy decisions rather than mechanical fixes, and I would rather they be designed with the upload layer than bolted on here.

One thing left open, and I am explicitly not blocking on it: the `can_descend` broken-symlink abort. My last reply on that thread has the detail, but the short version is that it is a plain `metadata` call rather than a `WalkDir` error, so the partial-scan semantics concern does not apply, and the one-line version keeps the whole suite green on both filesystems. Wine creates `My Documents` and friends as symlinks into host XDG directories that dangle when those directories are absent, so I think it will show up in practice. Your call whether it lands here or in a follow-up.

Nice work on the turnaround through these rounds.

PR #2607, opened 2026-07-30
feat(cloud-save): add native snapshot building

Hachi-R:
## Summary

- hash local save files in parallel with a reusable metadata cache
- build normalized local snapshots and aggregate hashes
- preserve variant and local-resolution metadata in snapshot files

## Context

This PR is stacked on #2606. It replaces the useful native snapshot work from #2532 and #2533, while intentionally omitting the obsolete snapshot comparator from #2534.

## Validation

- `cargo check`
- `git diff --check`

### DIFF
--- a/native/hydra-native/src/cloud_save/hashing/batch.rs
+++ b/native/hydra-native/src/cloud_save/hashing/batch.rs
@@ -1,0 +1,241 @@
+use std::collections::{BTreeSet, HashMap};
+use std::fs;
+use std::sync::OnceLock;
+
+use rayon::prelude::*;
+use time::OffsetDateTime;
+
+use super::file::hash_file;
+use super::types::{HashFilesResult, HashedLocalFile, LocalFileHashCacheEntry};
+
+pub struct HashFilesBestEffortResult {
+    pub result: HashFilesResult,
+    pub failures: Vec<(String, String)>,
+}
+
+pub const MAX_CONCURRENT_HASHES: usize = 8;
+const SHA256_HEX_HASH_LENGTH: usize = 64;
+const HASH_ALGORITHM: &str = "sha256";
+static HASHING_POOL: OnceLock<rayon::ThreadPool> = OnceLock::new();
+
+fn hashing_pool() -> &'static rayon::ThreadPool {
+    HASHING_POOL.get_or_init(|| {
+        rayon::ThreadPoolBuilder::new()
+            .num_threads(MAX_CONCURRENT_HASHES)
+            .thread_name(|index| format!("cloud-save-hash-{index}"))
+            .build()
+            .expect("failed to create cloud save hashing pool")
+    })
+}
+
+pub(crate) fn format_modified_at(modified: std::time::SystemTime) -> String {
+    let datetime = OffsetDateTime::from(modified);
+    format!(
+        "{:04}-{:02}-{:02}T{:02}:{:02}:{:02}.{:09}Z",
+        datetime.year(),
+        u8::from(datetime.month()),
+        datetime.day(),
+        datetime.hour(),
+        datetime.minute(),
+        datetime.second(),
+        datetime.nanosecond()
+    )
+}
+
+fn is_valid_hash(hash: &str) -> bool {
+    hash.len() == SHA256_HEX_HASH_LENGTH
+        && hash
+            .bytes()
+            .all(|byte| byte.is_ascii_digit() || (b'a'..=b'f').contains(&byte))
+}
+
+fn hash_file_with_cache(
+    absolute_path: String,
+    cached: Option<&LocalFileHashCacheEntry>,
+) -> Result<HashedLocalFile, String> {
+    let metadata = fs::metadata(&absolute_path).map_err(|error| error.to_string())?;
+    if !metadata.is_file() {
+        return Err("cloud_save_hash_path_is_not_file".to_string());
+    }
+
+    let size_bytes = metadata.len() as f64;
+    let last_modified_at = metadata
+        .modified()
+        .map(format_modified_at)
+        .map_err(|error| error.to_string())?;
+    let hash = cached
+        .filter(|entry| {
+            entry.absolute_path == absolute_path
+                && entry.size_bytes == size_bytes
+                && entry.last_modified_at == last_modified_at
+                && entry.algorithm.as_deref() == Some(HASH_ALGORITHM)
+                && is_valid_hash(&entry.hash)
+        })
+        .map(|entry| entry.hash.clone())
+        .map_or_else(|| hash_file(&absolute_path), Ok)?;
+
+    Ok(HashedLocalFile {
+        absolute_path,
+        size_bytes,
+        last_modified_at,
+        hash,
+    })
+}
+
+pub fn hash_files(
+    absolute_paths: Vec<String>,
+    hash_cache: Vec<LocalFileHashCacheEntry>,
+) -> Result<HashFilesResult, String> {
+    let result = hash_files_best_effort(absolute_paths, hash_cache);
+    if let Some((_, error)) = result.failures.into_iter().next() {
+        return Err(error);
+    }
+    Ok(result.result)
+}
+
+pub fn hash_files_best_effort(
+    absolute_paths: Vec<String>,
+    hash_cache: Vec<LocalFileHashCacheEntry>,
+) -> HashFilesBestEffortResult {
+    let paths = absolute_paths
+        .into_iter()
+        .collect::<BTreeSet<_>>()
+        .into_iter()
+        .collect::<Vec<_>>();
+    let cache_by_path = hash_cache
+        .into_iter()
+        .map(|entry| (entry.absolute_path.clone(), entry))
+        .collect::<HashMap<_, _>>();
+
+    let results = hashing_pool().install(|| {
+        paths
+            .into_par_iter()
+            .map(|absolute_path| {
+                let cached = cache_by_path.get(&absolute_path);
+                let path = absolute_path.clone();
+                hash_file_with_cache(absolute_path, cached).map_err(|error| (path, error))
+            })
+            .collect::<Vec<_>>()
+    });
+    let mut files = Vec::new();
+    let mut failures = Vec::new();
+    for result in results {
+        match result {
+            Ok(file) => files.push(file),
+            Err(failure) => failures.push(failure),
+        }
+    }
+    let hash_cache = files
+        .iter()
+        .map(|file| LocalFileHashCacheEntry {
+            absolute_path: file.absolute_path.clone(),
+            size_bytes: file.size_bytes,
+            last_modified_at: file.last_modified_at.clone(),
+            hash: file.hash.clone(),
+            algorithm: Some(HASH_ALGORITHM.to_string()),
+        })
+        .collect();
+
+    HashFilesBestEffortResult {
+        result: HashFilesResult { files, hash_cache },
+        failures,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::fs;
+
+    use tempfile::tempdir;
+
+    use super::*;
+
+    #[test]
+    fn reuses_valid_cache_and_processes_duplicate_paths_once() {
+        let temp = tempdir().unwrap();
+        let path = temp.path().join("save.dat").display().to_string();
+        fs::write(&path, b"save").unwrap();
+        let initial = hash_files(vec![path.clone()], vec![]).unwrap();
+        let mut cache = initial.hash_cache;
+        let cached_hash = "a".repeat(64);
+        cache[0].hash = cached_hash.clone();
+
+        let result = hash_files(vec![path.clone(), path], cache).unwrap();
+
+        assert_eq!(result.files.len(), 1);
+        assert_eq!(result.hash_cache.len(), 1);
+        assert_eq!(result.files[0].hash, cached_hash);
+        assert_eq!(hashing_pool().current_num_threads(), MAX_CONCURRENT_HASHES);
+    }
+
+    #[test]
+    fn invalidates_cache_mismatches() {
+        let temp = tempdir().unwrap();
+        let path = temp.path().join("save.dat").display().to_string();
+        let expected = "157dca92e4250458339d4b835250d44c238f3355e1b7986195188ee434e9baff";
+        fs::write(&path, b"save").unwrap();
+        let initial = hash_files(vec![path.clone()], vec![]).unwrap();
+
+        for mutate in [
+            |entry: &mut LocalFileHashCacheEntry| entry.hash = "invalid".into(),
+            |entry: &mut LocalFileHashCacheEntry| entry.size_bytes += 1.0,
+            |entry: &mut LocalFileHashCacheEntry| entry.last_modified_at = "changed".into(),
+            |entry: &mut LocalFileHashCacheEntry| entry.algorithm = None,
+        ] {
+            let mut cache = initial.hash_cache.clone();
+            mutate(&mut cache[0]);
+            assert_eq!(
+                hash_files(vec![path.clone()], cache).unwrap().files[0].hash,
+                expected
+            );
+        }
+    }
+
+    #[test]
+    fn drops_removed_files_and_sorts_results() {
+        let temp = tempdir().unwrap();
+        let first = temp.path().join("a.dat").display().to_string();
+        let removed = temp.path().join("removed.dat").display().to_string();
+        let second = temp.path().join("z.dat").display().to_string();
+        fs::write(&first, b"a").unwrap();
+        fs::write(&removed, b"removed").unwrap();
+        fs::write(&second, b"z").unwrap();
+        let initial =
+            hash_files(vec![removed.clone(), second.clone(), first.clone()], vec![]).unwrap();
+
+        let result = hash_files(vec![second.clone(), first.clone()], initial.hash_cache).unwrap();
+
+        assert_eq!(
+            result
+                .files
+                .iter()
+                .map(|file| file.absolute_path.as_str())
+                .collect::<Vec<_>>(),
+            vec![first.as_str(), second.as_str()]
+        );
+        assert_eq!(result.hash_cache.len(), 2);
+        assert!(!result
+            .hash_cache
+            .iter()
+            .any(|entry| entry.absolute_path == removed));
+    }
+
+    #[test]
+    fn accepts_empty_files_and_rejects_non_files() {
+        let temp = tempdir().unwrap();
+        let empty = temp.path().join("empty.dat").display().to_string();
+        fs::write(&empty, b"").unwrap();
+
+        let result = hash_files(vec![empty], vec![]).unwrap();
+
+        assert_eq!(result.files[0].size_bytes, 0.0);
+        assert_eq!(
+            result.files[0].hash,
+            "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+        );
+        assert_eq!(
+            hash_files(vec![temp.path().display().to_string()], vec![]).unwrap_err(),
+            "cloud_save_hash_path_is_not_file"
+        );
+    }
+}
--- a/native/hydra-native/src/cloud_save/hashing/batch.rs
+++ b/native/hydra-native/src/cloud_save/hashing/batch.rs
@@ -1,9 +1,10 @@
 use std::collections::{BTreeSet, HashMap};
 use std::fs;
 use std::sync::OnceLock;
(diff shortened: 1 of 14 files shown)


Moyasee (commented):
Went through the hashing and snapshot layer. The structure is good: streaming 64 KB hashing, a dedicated rayon pool instead of the global one, a canonical JSON aggregate so there is no unlengthed-concat ambiguity, byte-wise total sort keys, and a real three-stat TOCTOU recheck around the hash. Every FS entry point goes through `spawn_blocking`, so a worker panic surfaces as a JS error rather than killing the main process. I confirmed that last one by hand.

The problems are all in the degraded paths.

**`local_snapshot/guardrails.rs:56` — both guardrails are `#[cfg(test)]` dead code**

`prepare_snapshot_files` is the only function that checks `MAX_SNAPSHOT_FILE_COUNT` and `MAX_SNAPSHOT_TOTAL_SIZE_BYTES`, and it is gated behind `#[cfg(test)]`. Production calls `prepare_snapshot_files_best_effort`, which has no limits, plus `validate_built_files`, which computes the total but never compares it against the constant, so `SnapshotTooLarge` is only reachable from `checked_add` overflow at 2^64 bytes. A 1200-file snapshot and a single 3 GB file both build fine. The test at :201 even asserts that `best_effort` accepts a file of `MAX_SNAPSHOT_TOTAL_SIZE_BYTES + 1`. There is also no per-file cap, so a symlink to a very large file gets streamed through SHA-256 with no ceiling.

**`local_snapshot/build.rs:56` — an all-unavailable scan returns a successful empty snapshot**

If every file fails to stat (drive unmounted, prefix recreated, permissions changed), `prepare_snapshot_files_best_effort` marks them all unavailable, `build_snapshot` drops them with a `partial` coverage warning, and still returns `Ok` with `file_count: 0`. That snapshot's aggregate hash is byte-identical to the one produced for a caller who passed zero files. The native layer cannot distinguish "this game has no saves" from "the save location disappeared", and both are success. The only signal is advisory and lives in the same success payload. A `complete: bool` on the result would make this much harder to get wrong downstream.

**`hashing/batch.rs:31` — `format_modified_at` panics on a far-future mtime**

`OffsetDateTime::from(SystemTime)` panics on overflow. `touch -d @3153600000000` on a save file makes `build_local_game_snapshot` fail with `task 26 panicked with message "overflow adding duration to date"`. `spawn_blocking` contains it so the process survives, but it is deterministic: one file with a garbage timestamp (bad RTC, unpacked archive, some Proton writes) disables cloud save for that game permanently behind an opaque error. Pre-epoch mtimes are the other end, formatting as `-249-10-15T19:33:20.000000000Z`, which is an Invalid Date on the JS side.

**On greptile's `batch.rs:75` stale-cache comment**

Agreeing and adding a confirmation: write `AAAA`, snapshot, overwrite with `BBBB` at the same size with the mtime restored, snapshot again with the returned `hash_cache`, and the aggregate hash is unchanged. The realistic trigger is not a malicious writer, it is coarse mtime granularity (FAT32 2s, exFAT 10ms, HFS+ 1s, many SMB/NFS mounts) combined with fixed-size save slots. inode+dev in the key, or git's racy-timestamp rule where an mtime within N seconds of now is untrusted, would close it. Note that `reuses_valid_cache_and_processes_duplicate_paths_once` at :154 plants a fake hash and asserts it comes back verbatim, so the current test encodes the blind trust as intended behaviour.

Smaller things:

- `local_snapshot/build.rs:70` — `hashed_by_path.get(...)?` inside the `filter_map` drops a file with no coverage warning. Unreachable today since both misses are pre-filtered, but it is the one path that can truncate a snapshot while it still looks complete.
- `identity/mod.rs:67` — `relative_path` and `raw_path` only go through `normalize_path`, not `normalize_text`. macOS readdir returns NFD while Windows and Linux return NFC, so the same save set with a non-ASCII filename hashes differently across platforms and reads as changed. `normalize_text` already exists and is applied to rule paths.
- `hashing/mod.rs:34` — `build_snapshot_aggregate_hash` is a sync `#[napi]` fn, so sort plus serialize plus SHA-256 runs on the main thread over an unbounded caller-supplied list. It also skips the `variant_id ∈ variants` check that `build_snapshot` does, so the two entry points can disagree about which hashes are constructible.
- `hashing/mod.rs:10` — `pub(crate) use file::hash_file;` is unused and warns; `batch.rs:8` imports it directly.
- `build.rs:181` — the returned `hash_cache` keeps entries for files rejected as `file-changed-during-snapshot`. Self-healing in practice, but it should not retain them.
- CI provisions a Rust toolchain but never runs `cargo test` or `cargo clippy`, so the roughly 20 tests this PR adds do not run anywhere. Worth wiring up given how much logic now lives in this crate.

Hachi-R:
Follow-up after tracing every finding through the full stacked implementation rather than evaluating this slice in isolation.

### Changes that belong in this PR

The next revision of #2607 will address these points here:

1. **Stale metadata-cache hashes.** Cache persistence is introduced later in #2610, but the cache validation algorithm and its native contract are introduced here. A save rewritten with the same path, size, and timestamp must not be allowed to reuse stale content identity.
2. **Fallible timestamp conversion.** `format_modified_at` is introduced here, so invalid or out-of-range filesystem timestamps should become a per-file failure/partial-coverage result here rather than a panic or a timestamp that later JavaScript contracts reject.
3. **Unicode normalization of portable snapshot identity.** #2606 should preserve the filesystem path representation needed to access the real file. #2607 is where `rawPath` and `relativePath` become portable snapshot identity and aggregate-hash inputs, so NFC normalization belongs at this boundary while `absolutePath` remains untouched.
4. **Rejected cache entries.** Entries for files rejected by the final consistency check will be removed from the returned cache as part of the stale-cache correction. This is mostly self-healing today, so it does not need a separate design.

These issues are also present in the original final feature branch; they are inherited bugs, not behavior lost while splitting the stack.

### Findings intentionally handled elsewhere or left unchanged

- **500-file / 2 GB limits:** snapshot construction is intentionally best-effort so oversized sets can still be analyzed. Upload enforcement belongs to #2612 and is already applied immediately before upload.
- **All-unavailable input returning an empty snapshot:** the result is not indistinguishable from a real empty save set because it carries partial `coverage`. #2610 consumes that coverage to mark the merge/sync partial and prevent destructive conclusions. Adding a second `complete` flag here would duplicate that contract.
- **`hash_file` re-export appearing unused:** it is consumed by the restore implementation in #2608.
- **Aggregate file-to-variant membership:** local snapshot construction validates it here, and remote inputs are validated by the #2610 TypeScript contract before real aggregate-hash callers.
- **Synchronous aggregate helper:** real callers pass bounded snapshot metadata, not file contents; sorting, serializing, and hashing a few hundred small records is not a meaningful main-thread cost.
- **`hashed_by_path.get(...)?`:** both missing cases are currently removed or recorded before this lookup, so there is no reachable silent-truncation path in the current pipeline.
- **CI, clippy, live fixtures, and test infrastructure:** intentionally outside this functional slice.

Before requesting the next review, this PR will also be rebased onto the latest #2606 so the stack includes its final case-insensitive discovery fix.

native/hydra-native/src/cloud_save/local_snapshot/build.rs
```diff
@@ -0,0 +1,345 @@
...
+    for coverage in &mut input.coverage {
+        if let Some(raw_path) = &mut coverage.raw_path {
+            *raw_path = normalize_rule_path(raw_path);
+        }
+        if let Some(relative_path) = &mut coverage.relative_path {
+            *relative_path = normalize_rule_path(relative_path);
+        }
+    }
+
+    let prepared =
```
Moyasee, line 33: Normalizing before `validate_unique` means a normalization collision now fails the whole snapshot, where every other failure mode in this function degrades to partial coverage.

Two directory entries whose names differ only by Unicode form produce two records with distinct `absolutePath` values and `relativePath` values that are equal after NFC. `validate_unique` sees one logical identity twice and returns `DuplicateFile`, which `build_snapshot` turns into a hard `Err`:

```
two real distinct files, relative paths differ only by Unicode form
both forms   -> Err(cloud_save_duplicate_file)   <- whole snapshot lost, no coverage
single form  -> Ok: 1 files
```

The user gets no snapshot for that game at all, and no coverage record explaining why, which is a worse outcome than the `file-metadata-unavailable` and `file-hash-failed` paths right below this.

To be precise about reachability, because it cuts against me: the on-disk precondition needs a filesystem that stores names byte-verbatim, so ext4, btrfs and NTFS can hold both forms while macOS APFS is normalization-insensitive and collapses them. I could not create that state locally, so the repro above feeds the two records directly. The input handling is filesystem-independent, so that half is solid, but I am not claiming to have seen a real Linux tree do it.

I do not think normalization is the wrong call, it clearly belongs here. It is the response to a collision that looks out of step: pushing a `partial` coverage entry and dropping one of the two, the way the rest of this function handles bad files, would keep the failure proportional. Dropping either one silently is also not great, which is why coverage seems like the better answer than a dedup.

native/hydra-native/src/cloud_save/hashing/batch.rs
```diff
@@ -0,0 +1,410 @@
...
+        basic_info.ChangeTime
+    ))
+}
+
+#[cfg(not(any(unix, windows)))]
+fn metadata_fingerprint(_path: &str, _metadata: &fs::Metadata) -> Option<String> {
+    None
+}
+
+fn cache_timestamp_is_stable(modified: SystemTime, hashed_at: &str) -> bool {
```
Moyasee, line 114: This plus the fingerprint is a good fix, and I want to flag one Windows consequence rather than report a defect, since I cannot test it from here.

The Windows `metadata_fingerprint` opens a handle with `fs::File::open` to read `ChangeTime`. That means every file pays a `CreateFile` on every snapshot including cache hits, which is cheap next to hashing and probably fine. The part I am less sure about is that a failed open returns `None`, and `None` disqualifies the cache entry entirely, so any file the open cannot get at is re-hashed in full on every snapshot.

Rust's `File::open` requests a permissive share mode, so this only bites when something else opened the file with an exclusive share mode. Games holding their own save files open while running is the obvious case, and running-game snapshots are exactly when the cache is supposed to be earning its keep.

If that is understood and accepted, ignore this. If not, falling back to a fingerprint built only from `creation_time` and `file_attributes` when the open fails would keep some cache validity instead of dropping to none, though it would no longer see a pure content rewrite.

native/hydra-native/src/cloud_save/hashing/mod.rs
```diff
@@ -0,0 +1,39 @@
+mod aggregate;
+pub(crate) mod batch;
+mod file;
+pub(crate) mod types;
+
+use napi::bindgen_prelude::Error;
+use napi_derive::napi;
+
+pub(crate) use aggregate::build_hash as build_aggregate_hash;
+pub(crate) use file::hash_file;
```
Moyasee, line 10: `cargo check --lib` on this branch emits four dead-code warnings where the #2606 base emits none:

```
warning: unused import: `file::hash_file`
warning: constant `MAX_SNAPSHOT_FILE_COUNT` is never used
warning: constant `MAX_SNAPSHOT_TOTAL_SIZE_BYTES` is never used
warning: variant `TooManyFiles` is never constructed
```

I know from your follow-up that `hash_file` is for #2608 and that the limits move to #2612, so none of this is a mistake. It is just that the stack will build noisy from here until those land, and new warnings are much easier to notice when the baseline is zero.

An `#[allow(dead_code)]` with a one-line reason pointing at the PR that consumes each would keep the signal, and the reason comment is the sort `.cursorrules` already asks for on lint suppressions.

native/hydra-native/src/cloud_save/local_snapshot/guardrails.rs
```diff
@@ -0,0 +1,244 @@
+use std::collections::{HashMap, HashSet};
+use std::fmt;
+use std::fs;
+
+use super::types::{BuiltLocalSaveFile, DiscoveredLocalSaveFile};
+use crate::cloud_save::hashing::batch::format_modified_at;
+
+pub const MAX_SNAPSHOT_FILE_COUNT: usize = 500;
```
Moyasee, line 8: Worth knowing that these two constants and `prepare_snapshot_files` are now test-only, since the production path goes through `prepare_snapshot_files_best_effort`, which enforces neither limit. That matches what you described, with enforcement moving to #2612 right before upload.

The risk is only that a reader finds a `MAX_SNAPSHOT_FILE_COUNT` of 500 and a 2 GB cap sitting in `guardrails.rs` and concludes snapshot construction is bounded, when nothing here bounds it. A short note on the constants, or naming them for the upload limits they actually describe, would prevent that read.

native/hydra-native/src/cloud_save/local_snapshot/build.rs
```diff
@@ -0,0 +1,345 @@
...
+                hash: hashed.hash.clone(),
+                size_bytes: hashed.size_bytes,
+                last_modified_at: hashed.last_modified_at.clone(),
+                local_bindings: file.local_bindings,
+                confidence: file.confidence,
+                provenance: file.provenance,
+            })
+        })
+        .collect::<Vec<_>>();
+    let accepted_paths = built_files
```
Moyasee, line 119: This covers point 4 from your follow-up, and it lines up: only paths that survived the consistency check reach the returned cache, so a file rejected as `file-changed-during-snapshot` cannot seed a stale entry for the next run.

The three-way check above it is the part I like. Comparing metadata from before hashing, from hash time, and again afterwards catches a write that lands mid-hash, and on a cache hit `hashed.size_bytes` and `hashed.last_modified_at` come from the fresh `fs::metadata` rather than from the cache entry, so the check still means something on that path.

Moyasee (commented):
Reviewed at 65c5bdb against the #2606 branch, not main. The branch is already rebased onto 2fda30e, so the case-insensitive discovery fix is in the stack.

All four items you committed to in the follow-up have landed, and I checked each rather than taking the commit message for it.

The metadata fingerprint is the right mechanism for the stale-cache hole. Adding `dev`/`ino`/`ctime` on Unix closes the rewrite-with-restored-mtime case that size and mtime alone cannot see, the two-second racy window handles coarse-mtime filesystems, and requiring `metadata_fingerprint.is_some()` before trusting a cache entry fails in the safe direction. `invalidates_cache_when_contents_change_but_size_and_mtime_do_not` pins exactly the case that was broken.

On fallible timestamps, I round-tripped `format_modified_at` output back through the `Rfc3339` parser that `cache_timestamp_is_stable` uses, since it hand-builds the string rather than using the formatter:

```
epoch      -> 1970-01-01T00:00:00.000000000Z     rfc3339-parses=true
pre-1970   -> 1957-04-29T08:53:20.000000000Z     rfc3339-parses=true
year 0001  -> 0001-01-07T21:46:40.000000000Z     rfc3339-parses=true
year 9999  -> 9999-12-28T12:26:40.000000000Z     rfc3339-parses=true
now        -> 2026-07-31T18:14:36.818703000Z     rfc3339-parses=true
```

No gap there. NFC normalization lands at the right boundary with `absolutePath` untouched, and rejected entries are filtered out of the returned cache.

72 tests pass.

Two things below. One is a real behavior question about how a normalization collision is handled, the other is build hygiene.

I did not re-raise anything from your "handled elsewhere" list. The best-effort framing, the coverage-instead-of-a-`complete`-flag argument, and the #2608 and #2610 deferrals all read as reasonable to me.

Hachi-R:
Thanks for the follow-up and for validating the fixes against the stacked base.

I reviewed the remaining threads against the complete stack. No additional logic changes are planned for this PR:

- **Unicode normalization collisions:** the hard failure is intentional. After NFC normalization, the two files have the same portable snapshot identity and cannot both be represented or restored safely. Keeping one arbitrarily and reporting partial coverage could produce an incomplete backup while silently choosing which real save survives. Failing the snapshot is safer than uploading an ambiguous representation.
- **Windows metadata fingerprint:** failing open deliberately disables cache reuse. Reusing an old hash when the current file cannot be inspected would weaken the cache-correctness guarantee; the subsequent hash attempt will either read the current contents or report the file as unavailable/failed. The extra handle open on cache hits is an accepted cost.
- **Dead-code warnings:** these are temporary consequences of the stacked PR boundaries. `hash_file` is consumed by #2608, while the upload-limit symbols are consumed by #2612. Adding suppressions here would only hide that relationship and add noise that disappears later in the stack.
- **Snapshot limit naming/documentation:** enforcement intentionally remains at the upload boundary in #2612 so native snapshot construction can retain best-effort analysis and coverage. This is already documented in the stack review context, so I am leaving this slice unchanged.
- **Rejected cache entries:** confirmed fixed; only files that survive the consistency checks are retained in the returned cache.

So the remaining comments are either intentional safety behavior, temporary stack hygiene, or confirmations of the implemented fixes. The functional changes committed for stale-cache handling, portable timestamps, NFC identity, and rejected cache entries remain as-is.

native/hydra-native/src/cloud_save/local_snapshot/build.rs
Moyasee: Taking your safety argument seriously made me go look harder at this rather than close it out, and I found the case is wider than what I originally reported. Two corrections, one of them to you and one to me.

First, mine. I framed this as a Unicode problem and noted I could not reproduce the on-disk state locally. That was too narrow. `normalize_rule_path` is `normalize_text(&value.replace('\\', "/"))`, so it also folds backslashes into slashes, and backslash is a legal filename character on ext4 and btrfs. That version reproduces on any filesystem:

```
both exist on disk: true and true
they are different files: true
both -> Err(cloud_save_duplicate_file)   <- entire snapshot lost
```

That is a file literally named `save\slot1.dat` sitting next to a real `save/slot1.dat`. Wine and Proton prefixes are exactly where you find literal-backslash filenames, because a game passing a Windows path to a POSIX call creates one. So the precondition is not exotic, and the result is that a single oddly-named file costs the user every cloud save for that game, with an error that does not say which files collided.

I think the root cause is narrower than the guard, though. #2606 already normalizes separators: `scan_path.rs` runs `relative_path` through `normalize_path`, which does the same backslash replacement, on the platform where a backslash actually is a separator. By the time a `relativePath` reaches `build_snapshot` on Linux, any backslash left in it is data, not a separator, and replacing it corrupts the identity. `rawPath` is a manifest rule path where the replacement is correct, so the two fields want different treatment: `normalize_rule_path` for `rawPath`, plain `normalize_text` for `relativePath`. That removes the backslash collisions entirely and leaves only the genuine Unicode case for the guard to catch.

Second, the correction to your reply. "The two files cannot both be represented" is not true of this slice. `build_aggregate_hash` accepts two files with identical `variantId`, `rawPath` and `relativePath` and different hashes, and produces a well-defined result distinct from the single-file one:

```
aggregate with two same-identity files: 8178704ddd57990d...
aggregate with one:                     8fe4297b5bcee086...
```

The structures represent it fine. What cannot handle it is restore, which has no way to pick an absolute path, and that contract lives in #2608 and #2610 rather than here.

That does not make failing closed wrong. If restore genuinely cannot disambiguate, refusing the snapshot is a defensible call and I am not going to argue you out of it. But it is worth knowing that the reasoning rests on the downstream contract rather than on this slice being unable to hold the data, and that `validate_unique` predates the normalization added in 65c5bdb, so this collision class arrived as a side effect rather than as something the guard was written for.

On your other three: the Windows fingerprint reasoning is right and I withdraw that one, and the dead-code and limit-naming calls are yours to make on hygiene grounds. Only this one seems worth another look.

Moyasee: Fixed and verified at 3b25db4c.

```
backslash vs slash -> Ok: 2 files ["save/slot1.dat", "save\\slot1.dat"]
NFC vs NFD         -> Err(cloud_save_duplicate_file)
normal nested path -> Ok: ["Saves/Profile1/slot.dat"]
```

Splitting `normalize_text` from `normalize_rule_path` kills the backslash collisions and leaves the guard catching the genuine identity case you argued for. `portable_relative_path` in #2606 fixes the upstream half.

Moyasee (approved):
Approving.

Verified at 3b25db4c: backslash and slash relative paths stay distinct, the duplicate guard still catches the genuine Unicode identity collision, and normal nested paths are unaffected. The `normalize_text` / `normalize_rule_path` split is the right seam, and #2606 covers the upstream half.

Also checked the four fixes from the earlier round rather than taking the commit message for them: the metadata fingerprint closes the stale-cache hole, `format_modified_at` round-trips through the `Rfc3339` parser it feeds across the full range, NFC lands on portable identity with `absolutePath` untouched, and rejected entries stay out of the returned cache.

Remaining open items are yours by design: the dead-code warnings clear when #2608 and #2612 land, and the limit constants stay put with enforcement at the upload boundary.

PR #2608, opened 2026-07-30
feat(cloud-save): add native transfer pipeline

Hachi-R:
## Summary

- compose manifest lookup, path discovery, scanning, identity, and snapshot building in one native pipeline
- upload local save blobs through the native addon
- resolve, download, verify, and replace restore targets with path validation and metadata preservation

## Context

This PR is stacked on #2607 and replaces #2537 and #2538 using the final Cloud Saves V2 implementation. Local cloud-save deletion is intentionally excluded and will remain a separate later cut.

## Validation

- `cargo check`
- `git diff --check`

### DIFF
--- a/native/hydra-native/src/cloud_save/restore/resolve_targets.rs
+++ b/native/hydra-native/src/cloud_save/restore/resolve_targets.rs
@@ -18,7 +18,9 @@ use super::types::{
     BlockedRestoreFile, ResolveRestoreTargetsInput, ResolveRestoreTargetsResult,
     ResolvedRestoreTarget, RestoreManifestFile,
 };
-use super::validation::{validate_hash, validate_relative_path, validate_size};
+use super::validation::{
+    validate_hash, validate_relative_path, validate_size, validate_windows_relative_path,
+};
 
 const STEAM_INDIVIDUAL_ACCOUNT_BASE: u64 = 76_561_197_960_265_728;
 
@@ -31,7 +33,7 @@ fn join_path(root: &str, relative_path: &str) -> String {
     .replace('\\', "/")
 }
 
-fn validate_file(file: &RestoreManifestFile) -> Result<(), String> {
+fn validate_file(file: &RestoreManifestFile, windows_semantics: bool) -> Result<(), String> {
     if file.variant_id.len() != 64
         || !file.variant_id.bytes().all(|byte| byte.is_ascii_hexdigit())
         || file.raw_path.is_empty()
@@ -40,6 +42,9 @@ fn validate_file(file: &RestoreManifestFile) -> Result<(), String> {
         return Err("cloud_save_invalid_restore_identity".to_string());
     }
     validate_relative_path(&file.relative_path)?;
+    if windows_semantics {
+        validate_windows_relative_path(&file.relative_path)?;
+    }
     validate_hash(&file.hash)?;
     validate_size(file.size_bytes)?;
     parse_last_modified_at(&file.last_modified_at).map(|_| ())
@@ -184,8 +189,7 @@ fn has_glob(raw_rule: &str) -> bool {
         .any(|character| matches!(character, '*' | '?' | '[' | '{'))
 }
 
-#[napi]
-pub fn resolve_restore_targets(
+fn resolve_restore_targets_inner(
     input: ResolveRestoreTargetsInput,
 ) -> napi::Result<ResolveRestoreTargetsResult> {
     let context = build_context(&ResolveSaveRulesInput {
@@ -202,6 +206,7 @@ pub fn resolve_restore_targets(
     })
     .map_err(Error::from_reason)?;
     let case_sensitive = context.platform == "linux" && !context.windows_compatibility;
+    let windows_semantics = context.platform == "windows" || context.windows_compatibility;
 
     let mut variants = HashMap::new();
     for variant in input.variants {
@@ -225,7 +230,7 @@ pub fn resolve_restore_targets(
     let mut used_variants = HashSet::new();
 
     for file in input.files {
-        validate_file(&file).map_err(Error::from_reason)?;
+        validate_file(&file, windows_semantics).map_err(Error::from_reason)?;
         if !identities.insert(identity_key(&file)) {
             return Err(Error::from_reason("cloud_save_duplicate_restore_identity"));
         }
@@ -430,6 +435,15 @@ pub fn resolve_restore_targets(
     })
 }
 
+#[napi]
+pub async fn resolve_restore_targets(
+    input: ResolveRestoreTargetsInput,
+) -> napi::Result<ResolveRestoreTargetsResult> {
+    tokio::task::spawn_blocking(move || resolve_restore_targets_inner(input))
+        .await
+        .map_err(|_| Error::from_reason("cloud_save_restore_plan_task_failed"))?
+}
+
 #[cfg(test)]
 mod tests {
     use std::fs;
@@ -515,7 +529,7 @@ mod tests {
             preferred_path: None,
         }];
 
-        let result = resolve_restore_targets(restore_input).unwrap();
+        let result = resolve_restore_targets_inner(restore_input).unwrap();
 
         assert!(result.blocked.is_empty());
         assert_eq!(result.actions.len(), 1);
@@ -556,7 +570,7 @@ mod tests {
             preferred_path: Some(approved_root.display().to_string()),
         }];
 
-        let result = resolve_restore_targets(restore_input).unwrap();
+        let result = resolve_restore_targets_inner(restore_input).unwrap();
 
         assert!(result.blocked.is_empty());
         assert_eq!(result.actions.len(), 1);
@@ -597,7 +611,7 @@ mod tests {
             preferred_path: Some(selected.display().to_string()),
         }];
 
-        let result = resolve_restore_targets(restore_input).unwrap();
+        let result = resolve_restore_targets_inner(restore_input).unwrap();
 
         assert!(result.blocked.is_empty());
         assert_eq!(
@@ -620,7 +634,7 @@ mod tests {
             .iter()
             .map(|variant| file(variant, "slot.dat"))
             .collect();
-        let result = resolve_restore_targets(input(
+        let result = resolve_restore_targets_inner(input(
             temp.path(),
             StoreUserContext::default(),
             variants,
@@ -637,7 +651,7 @@ mod tests {
         let temp = tempdir().unwrap();
         let variants = vec![variant("opaque-folder", "Unknown")];
         let files = vec![file(&variants[0], "slot.dat")];
-        let result = resolve_restore_targets(input(
+        let result = resolve_restore_targets_inner(input(
             temp.path(),
             StoreUserContext::default(),
             variants,
@@ -664,7 +678,8 @@ mod tests {
         };
         let variants = vec![variant("steam-account", "76561197960278073")];
         let files = vec![file(&variants[0], "slot.dat")];
-        let result = resolve_restore_targets(input(temp.path(), context, variants, files)).unwrap();
+        let result =
+            resolve_restore_targets_inner(input(temp.path(), context, variants, files)).unwrap();
 
         assert!(result.blocked.is_empty());
         assert_eq!(result.actions.len(), 1);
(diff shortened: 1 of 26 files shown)


Moyasee (commented):
Spent most of the time trying to break the restore path from remote-controlled input and could not. `validate_relative_path` covers `..`, absolute paths, both separators, drive-relative and UNC/verbatim prefixes, and containment is checked against the canonicalized path via `canonical_path_with_missing`, so a symlinked parent inside the save folder is rejected and the dangling-symlink variant dies at `create_dir_all`. Verification is genuinely strong too: the blob is hashed against the expected hash three times (temp, staging, post-install), a mismatch is fatal, and the install is a same-directory backup-rename/install-rename with a working rollback. Good design.

What I would not merge as-is:

**`restore/replace_targets.rs:530` — one bad target aborts the whole batch and reports zero failures**

`validate_target` failure returns `Ok(metadata_result(failure))`, and `metadata_result` builds `restored_files: []`, `skipped_files: []`, `failed_files: []` with a single metadata failure. Compare `failed_result` right above it, which correctly populates `failed_files` for every input. So if one entry in a 200-file snapshot is a symlink (`target_is_symlink` at :140) or has an unparseable `last_modified_at`, the other 199 are never attempted and never appear anywhere in the result. A caller checking `failed_files.length === 0` sees a successful restore that restored nothing.

This one matters more than it looks because of how #2610 consumes it: `sync-game/transfer.ts:98` throws on `metadataFailedPaths > 0`, `runAutomaticCloudSaveSync` swallows that, and `shouldBlockGameLaunchForCloudSave(null)` is false. So a single symlinked save file means the game launches with nothing restored and the post-exit upload silently blocked.

**`restore/resolve_targets.rs:188` — `resolve_restore_targets` is a sync `#[napi] pub fn` doing unbounded blocking FS work on the main thread**

Per file, per matching rule, per user value it calls `resolve_restore_root` (:295), which globs and can do a full recursive `WalkDir` (`restore_root.rs:307` into `scan_path.rs:79`), plus `canonicalize` (:84) and a full SHA-256 of every existing target (:367). No memoization, no cap on `input.files`. A 500-file snapshot over a large save folder freezes the UI for the whole planning phase. Everything else in the module correctly uses `spawn_blocking` (`pipeline/mod.rs:252`), so this one looks like an oversight rather than a decision.

**`restore/download_blob.rs:37` and `upload/upload_blob.rs:43` — no HTTP timeout, and no cap on downloaded size**

Both use `reqwest::Client::new()` with no connect/read/total timeout, while `manifest/cache.rs:27` correctly uses `Client::builder().timeout(...)`. The chunk loop writes the body until the server stops, and `size_bytes` is validated in `validation.rs:15` but never passed to `download_restore_blob_to_temp`, so the known expected size is not enforced. A hung presigned endpoint stalls the restore forever with no cancellation. `Client::new()` also panics if the TLS backend fails to initialize, which is reachable inside an async N-API future; `builder().build()?` avoids that and lets you reuse one client instead of re-doing the handshake per file.

**`restore/replace_targets.rs:318` — no fsync before the backup is deleted**

`tokio::fs::copy` writes the staging file at :318, `rename` installs it at :377, and `remove_backups` deletes the user's original at :626, with no `sync_all` on the staging file and no parent-directory fsync anywhere. A power loss after the batch completes can leave a torn target with the original already unlinked. `download_blob.rs:72` does call `sync_all` for the temp blob, so the omission here reads as unintentional.

**`restore/replace_targets.rs:87` — crash residue lands in the save directory and the scanner will upload it**

`sibling_path` puts `.hydra-restore-<uuid>-stage` and `-backup` next to the target, i.e. inside the save folder. If the process dies between :371 and :626 the user's original save is stranded under a random hidden name. There is no startup sweep for stale `.hydra-restore-*`, and `save_scanner/scan_path.rs:80` filters nothing while `require_literal_leading_dot` is false at :33, so the next snapshot picks that residue up as legitimate save data.

Also worth noting: `local_snapshot/guardrails.rs` caps do not actually apply (they are `#[cfg(test)]`, see my comment on #2607), so neither side of the transfer has a file-count or total-size limit right now.

Smaller things:

- `restore/replace_targets.rs:123` and `resolve_targets.rs:109` — `format!("{}/", root.trim_end_matches('/'))` yields `"/"` for a root of `/`, so any absolute target passes containment. Not remote-controlled today, but an empty-after-trim root should be rejected outright.
- `restore/resolve_targets.rs:276` — `preferred_path` is used verbatim as the restore root with no validation, bypassing the wine-prefix confinement in `restore_root.rs:178`. It is never checked for being absolute, existing, or a directory, so a relative value resolves against the process CWD. Local config rather than remote input, so hardening rather than a live hole.
- `restore/validation.rs:27` — the `:` check is position-1 only, so `slot.dat:evil` writes an NTFS alternate data stream on Windows. NUL bytes and `COM1`/`CON`/`NUL` are also unfiltered. Everything still lands inside the root, so this is hygiene.
- `restore/replace_targets.rs:519` vs `:525` — `validate_relative_path` is called on `relative_path`, but every subsequent operation uses the JS-supplied `target_path`/`restore_root_path`; `relative_path` is only echoed back. Asserting `path_key(target) == path_key(root.join(relative_path))` would stop the native layer from depending entirely on the JS glue.
- `restore/replace_targets.rs:546` and `:628` — a `remove_backups` or `cleanup_staging` failure turns a completed restore into a thrown JS error, and at :546 it discards the `failed_result` accounting entirely.
- `restore/replace_targets.rs:591` — a `Skip` target whose local hash changed mid-restore rolls back the entire batch. Fail-safe, but all-or-nothing over a long restore is fragile.
- `resolve_targets.rs:204` vs `replace_targets.rs:113` — case-sensitivity is derived from `platform == "linux" && !windows_compatibility` in one place and hardcoded `cfg!(windows)` in the other. On macOS and Wine-on-Linux the two layers disagree about target identity, so the `seen_targets` dedup at :525 can miss a collision resolve would have blocked.
- `identity/mod.rs` — this commit deletes `capture_validation_rejects_path_segments` and `local_ids_are_length_delimited`. The first asserted that `is_safe_capture` rejects `""`, `.`, `..`, `a/b`, `a\b`, `a\0b`, and `is_safe_capture` is now load-bearing since it gates `concrete_folder_id` in `validate_variant` (`resolve_targets.rs:73`), which feeds folder names on the restore path. Worth keeping.
- `restore/replace_targets.rs:269` — `.expect("expected hash is validated before target metadata")` is unreachable today but is a panic across the N-API boundary guarded only by call-order convention.

Hachi-R:
Follow-up after tracing the review findings through the actual callers in the stacked implementation.

We will address the following five items in this PR/stack revision:

1. **Move restore-target planning off the main thread.** `resolveRestoreTargets` performs globbing, filesystem traversal, canonicalization, and hashing synchronously. It will become asynchronous and execute the blocking planner through `spawn_blocking`; the #2610 caller is already inside an async flow.
2. **Bound HTTP transfers.** Upload and download clients will be built fallibly with explicit timeouts. Restore downloads will receive the expected manifest size and stop if the response exceeds it, instead of writing until the endpoint closes.
3. **Make restore installation durable before deleting backups.** Staging/installed content will be synced before the original backup is removed, with the applicable directory durability step where supported.
4. **Handle `.hydra-restore-*` crash residue.** Restore artifacts must not be discovered as legitimate save files, and stale staging/backup residue will receive explicit cleanup/recovery handling.
5. **Harden Windows restore filenames.** Relative-path validation will reject alternate-data-stream syntax, NULs, and reserved Windows device names before any target is created.

The remaining findings were also checked against the full stack and will not receive logic changes here:

- A same-size source rewrite can make an upload fail, but the required `x-amz-checksum-sha256` is validated by storage, so stale bytes cannot become a valid cloud object. This is a fail-safe upload rejection, not silent corruption.
- A target metadata validation failure is not interpreted as success: the #2610 caller checks both `failedFiles` and `metadataFailures`. Aborting the batch before mutation is the intended all-or-nothing behavior, and launch-on-non-conflict cloud failure is intentional.
- The containment TOCTOU requires a same-user local process replacing a writable parent during the commit window. It is not reachable from remote-controlled input and a complete fix requires platform-specific handle-relative filesystem operations, so it is not treated as a P1 for this feature slice.
- `/` legitimately contains every absolute Unix path; selecting it as a local root is a local-user choice.
- `preferredPath` comes from locally validated custom-path bindings in #2612, not from the remote manifest.
- `targetPath` and `restoreRootPath` are produced by the native resolver and merely carried through the main-process glue before replacement; renderer or remote input cannot supply them in the production flow.
- Rolling back when a `Skip` target changes is intentional atomic/fail-safe behavior.
- Case-folded collisions are already rejected by the resolver before replacement in the production flow.
- The guarded `expect` is unreachable after the immediately preceding required-hash validation.
- Upload file-count and aggregate-size enforcement is intentionally added at the upload boundary in #2612.
- The removed assertions are test-maintenance feedback rather than a production defect and are outside this correction scope.

I will keep each fix minimal and preserve the all-or-nothing restore semantics.

Moyasee (commented):
Reviewed at c6a1013e against the #2607 branch. Rebased on 3b25db4c, so the stack is current.

All five items from your follow-up landed and are wired: `resolve_restore_targets` runs through `spawn_blocking`, the blob client has connect/read/total timeouts, downloads are bounded by both the declared `Content-Length` and the streamed byte count, `remove_backups` runs after the install is synced, `is_restore_artifact_path` is hooked into all three scanner paths, and the Windows validator rejects ADS syntax, NULs and reserved device names. Signed URLs are scrubbed from error text via `without_url`, which is a good catch to have made yourself.

Suite is 116/118. One failure is a broken test, inline. The other is `preserves_literal_backslashes_in_filesystem_paths` from #2606, which is macOS-only (`/var` versus `/private/var` against `canonicalize`) and not a Linux or Windows defect, so I am not counting it against this PR.

Not re-raising the containment TOCTOU or the upload checksum item. Your threat-model reasoning on both holds up.

native/hydra-native/src/cloud_save/restore/resolve_targets.rs
```diff
@@ -0,0 +1,987 @@
...
+            .ends_with("/Game/save.dat"));
+        assert!(result.actions[0]
+            .restore_root_path
+            .replace('\\', "/")
+            .ends_with("/Game"));
+        assert_eq!(result.actions[0].last_modified_at, LAST_MODIFIED_AT);
+    }
+
+    #[test]
+    fn resolves_an_approved_custom_directory_and_can_create_it() {
```
Moyasee, line 862: This test only passes on a Windows host. It fails on Linux and macOS:

```
assertion failed: result.blocked.is_empty()
```

It declares `platform: "windows"` but builds `raw_path` from `tempdir()`, and `validate_absolute_path` requires a drive letter when the platform is windows:

```
BLOCKED   <- /var/folders/ts/abc/T/.tmpXYZ/Custom Saves
BLOCKED   <- /tmp/.tmpXYZ/Custom Saves
accepted  <- C:/Users/hydra/AppData/Local/Temp/.tmpXYZ/Custom Saves
```

Linux is a shipping target and the usual dev and CI platform here, so this is not the macOS-only category. Using a synthetic `C:/...` root instead of the host temp directory keeps the test honest about the platform it declares, since nothing in this path touches the real filesystem for the `create` case anyway.

Hachi-R: Fixed in 6c46010ee. The test now uses a synthetic Windows root (`C:/Users/hydra/AppData/Local/Temp/Custom Saves`) and a matching Windows home path instead of `tempdir()`, so it no longer depends on the host OS. The targeted test passes, and the full native suite passes (113/113) on Windows.

Moyasee: Confirmed at 6c46010e. The test passes here now, and the suite is 117/118 with the only failure being the macOS-only `/var` symlink one from #2606.

Your 113 checks out too: there are exactly 5 `#[cfg(unix)]` tests, and `preserves_literal_backslashes_in_filesystem_paths` is one of them, so 118 minus 5 is the Windows count.

Moyasee (approved):
Approving.

The broken test is fixed and verified at 6c46010e, and all five hardening items from your follow-up are in and wired: `spawn_blocking` for restore planning, blob client timeouts, downloads bounded by both the declared and streamed size, backups removed only after the install syncs, artifact exclusion in all three scanner paths, and Windows filename validation covering ADS, NULs and reserved devices.

Scoping note so the approval is not read as broader than it is: I went deep on transfer, restore, and validation, and read `pipeline/mod.rs` and `restore_root.rs` less closely. Nothing there looked wrong, but I would not claim the same level of checking on those two.

The containment TOCTOU and the upload checksum decision both rest on reasoning that holds, so I am not blocking on either.

PR #2609, opened 2026-07-30
feat: check hoster file availability on game page load

Moyasee:

**Fill in the PR content:**

Wires up the new `POST /hosters/availability` endpoint so we can tell the user upfront when a download option is dead instead of letting them find out after picking it.

The endpoint only supports VikingFile right now, so that's the only hoster we send. Adding more later is a one-line change to `VIKINGFILE_URI_PREFIXES` / `supportsHosterAvailabilityCheck` in `src/shared/hoster-availability.ts`.

### How it works

The call goes through `HydraApi` like the rest of the API, and lives in the main process (`HosterAvailabilityApi`), reached from the renderer through a new `checkHostersAvailability` IPC event. Keeping it in main means the batching and the cache are shared between the main window and Big Picture instead of being duplicated per renderer.

- URLs are deduped and batched in groups of 50, matching the documented `1...50` limit
- Results are cached for 5 minutes, so bouncing between game pages doesn't spam the endpoint
- On a 429 we stop sending the remaining batches instead of pushing through the rate limit
- Sent with `needsAuth: false`, since the endpoint doesn't require authentication

### Where it runs

Right after the download options are fetched, on both the regular game details page and the Big Picture download modal. Results get merged into `unavailableUris`, which means the availability orbs in the repacks modal and the per-downloader availability in the download settings modal pick it up without any changes to those components. A live check overrides whatever the download sources API said for that specific URL, since it's the more recent answer.

If the request fails for any reason, the options are left exactly as the download sources API returned them.

### Notes

Response parsing follows the documented `{ results: [{ url, available }] }` contract and ignores any entry that doesn't match that shape, so a contract drift degrades to "no availability info" rather than marking everything offline.

### DIFF
--- a/src/big-picture/src/hooks/queries/use-game-download-options.hook.ts
+++ b/src/big-picture/src/hooks/queries/use-game-download-options.hook.ts
@@ -2,6 +2,7 @@ import { useEffect, useState } from "react";
 import { IS_DESKTOP } from "../../constants";
 import type { DownloadSource, Game, GameRepack } from "@types";
 import { orderBy } from "lodash-es";
+import { applyHosterAvailability, getCheckableHosterUris } from "@shared";
 
 export type DownloadOptionsEmptyStateReason =
   | "no-configured-sources"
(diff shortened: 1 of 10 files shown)


Moyasee:
Closing this, going a different direction.

PR #2610, opened 2026-07-30
feat(cloud-save): add sync core

Hachi-R:
## Summary

- add the main-process Cloud Saves V2 sync orchestration, policies, and conflict handling
- persist local hash caches, environment generations, automatic-sync settings, and sync anchors
- connect the native snapshot/transfer APIs to the Node services and IPC events
- replace the obsolete native snapshot comparator with variant-aware TypeScript merge logic

## Context

This PR is stacked on #2608 and supersedes #2539. Renderer and Big Picture UI remain in later cuts.

## Validation

- `npm.cmd run typecheck:node`
- `cargo check`
- `git diff --check`


Moyasee (commented):
Split this across three passes: the sync decision layer, the transfer/environment glue, and the persistence/IPC/native-binding surface.

Things I checked and found correct, so nobody has to re-check them: the native binding table in `native-addon.ts` matches the Rust exactly, including sync-vs-async for all 11 functions, the snake-to-camel field mapping, the `string_enum` values, and the absence of any `u64`/`i64` (so no BigInt surprises). `map-with-concurrency.ts` preserves ordering, catches every mapper rejection inside its own worker, and rethrows the first error. The local hash cache is self-pruning (`batch.rs:128` rebuilds from only the files hashed that run, and `build-local-game-snapshot.ts:28` dels on empty). Entitlement gating is applied at every entry point and the shared and main copies agree. Level keys are `JSON.stringify` arrays, so no delimiter collisions. `services/index.ts` adding `export * from "./cloud-save"` introduces no name collisions and no cycle. Anchor writes are ordered after the transfer on all four paths, and `hasDeferredLocalChanges` correctly refuses to anchor when upload-only leaves work pending.

The merge algorithm itself is well built on the axes it was designed for: an empty local snapshot short-circuits to full restore, incomplete coverage never proves a deletion, and genuine both-sides divergence reaches `conflicts` on every branch I traced.

## The one I would hold on

**`merge-user-variant-snapshots.ts:158` — remote files that were never restored locally get deleted from the cloud on a later sync**

`unresolvedRemoteEntryIds` is written by every anchor writer and read only by the UI (`get-cloud-save-overview.ts:24`, `get-cloud-save-v2-file-details.ts:33`). The merge never consults it: `baseById` is built from `base.entries` alone at :74. Meanwhile both anchor writers put those entries into `entries` anyway — `saveCurrentHeadAnchor` writes the entire remote manifest (`sync-game-cloud-save.ts:92`), and `saveSnapshotAnchor` writes `merge.files`, which includes remote-only entries that were pushed but deliberately not restored (:151).

So:

1. Sync 1 hits a transient FS error on a save root. Coverage is `partial`, `remote.sav` goes into `unresolvedRemoteEntryIds` and is not restored. `proposalChanged` is false, so `sync-game-cloud-save.ts:338` anchors the full remote head with `remote.sav` in `entries`.
2. Sync 2 with a healthy scan. `remote.sav` is remote-only, `baseEntry` now exists, `coverageStateFor` returns `provesDeletion: true`, and `sameBytes(remoteFile, baseEntry)` is true, so :163 hits `deleteRemoteEntryIds.add`. The file drops out of `merge.files`, `proposalChanged` flips, and a snapshot without it is committed.

The cloud copy is gone and it never existed on disk. `merge-user-variant-snapshots.test.ts:190` covers the first half and `:131` covers the second half; nothing covers the sequence. Same trigger via the pre-launch path at `sync-game-cloud-save.ts:255` for entries returned as blocked. Either exclude unresolved entries from `entries`, or have the merge treat anything in `base.unresolvedRemoteEntryIds` as unable to prove a deletion.

## Rollout

**`automatic-sync-settings.ts:36` — V2 automatic sync defaults to on, and the existing toggle cannot turn it off**

`v2Enabled: storedV2Enabled ?? true` means a game with nothing stored resolves to mode `"v2"`, so `launch-game.ts:454` and `process-watcher.ts:66` run pre-launch and post-exit sync for every game of every subscriber. The switch the user actually sees is bound to `game.automaticCloudSync`, which is `false`, so the UI reads "off". Toggling it calls `setLegacyCloudSaveAutomaticSyncEnabled(..., false)`, which lands on `getNextCloudSaveAutomaticSyncMode("v2", "legacy", false)`, and since `currentMode !== targetMode` that returns `"v2"` unchanged. `automatic-sync-mode.test.ts:69` asserts exactly this. The new `setCloudSaveAutomaticSyncEnabled` IPC handler that would work has no renderer caller yet.

I realise the UI lands in a later cut, so flagging this as a sequencing constraint rather than a code bug: this must not ship ahead of the UI, or default to off until it does.

**`automatic-sync-settings.ts:85` — a getter rewrites the user's persisted setting**

`getCloudSaveAutomaticSyncMode` is called from `launch-game.ts:452`, `process-watcher.ts:39,52`, and the `getCloudSaveOverview` handler. For any existing user with legacy automatic cloud sync on, the first call resolves to `"v2"`, then writes `automaticCloudSync: false` and broadcasts `on-library-batch-complete`. Their toggle visibly flips itself off and the legacy Ludusavi backup at `process-watcher.ts:41` stops. If that migration is intended it should be an explicit one-time step, not a side effect of a read.

## Worth fixing before merge

**`process-watcher.ts:348` — one false "game closed" tick permanently disables that session's upload**

`NativeAddon.getSystemProcessMap()` degrades to empty maps on any transient failure (worker catch at `native-addon.ts:165`, and `drainResolvers` at :332 on worker error or exit). Every running game then takes the `else if (gamesPlaytime.has(gameKey)) onCloseGame(game)` branch, which calls `runAutomaticCloudSavePostExit`, whose first statement is `consumeCloudSaveLaunchGuard` — it deletes the guard. The next tick re-opens the session via `onOpenGame`, but only `launchGame` ever sets a guard, so at the real exit the post-exit sync logs `missing_launch_guard` and uploads nothing. The spurious close has also already published a mid-play snapshot as the remote head and moved the anchor to it.

**A failed pre-launch restore lets the game start anyway**

`restoreRemoteState` (`sync-game/transfer.ts:98`) throws after `replaceRestoreTargets` has already run, `runAutomaticCloudSaveSync` swallows it and returns `null`, and `shouldBlockGameLaunchForCloudSave(null)` is false because of the `?.` — so the launch proceeds. `canCreateCloudSaveUploadGuard` needs a non-null pre-launch result, so `uploadAllowed` is false too. Net result: the user plays on stale or partially restored saves, nothing uploads afterwards, and there is no visible error. A restore failure should block launch the way a conflict does.

This composes badly with `replace_targets.rs:530` in #2608 (a single symlinked save file returns success-shaped with empty `failed_files` after restoring nothing) and with the fact that `metadataFailedPaths > 0` throws here even when every file landed correctly — on a filesystem where directory mtimes cannot be set, that repeats forever.

**Fire-and-forget syncs can take the main process down**

`scan-installed-games.ts:109`, `select-game-wine-prefix.ts:29,47`, `update-executable-path.ts:40`, and `game-files-manager.ts:373` all `void runAutomaticCloudSaveSync(...)` with no `.catch`. That function only wraps `syncGameCloudSave`; the `await getCloudSaveAutomaticSyncEnabled(...)` at `automatic-sync.ts:52` is unguarded and both reads and `batch.write()`s to LevelDB. Quitting the app while an install finishes rejects there, and there is no `unhandledRejection` handler anywhere in `src/main`. `process-watcher.ts:422,548` does this correctly with `.catch`; these five do not. `scan-installed-games.ts` additionally fires one unawaited full sync per newly found game with no cap.

**`snapshot-retry-policy.ts:3` — every HTTP error is classified as a transport failure**

A real `AxiosError` for a non-2xx is an `Error` whose message is `Request failed with status code NNN`, so the second clause matches for 401, 409, 429 and 500, and `HydraApi` rethrows the original error unchanged. `commitPendingSnapshot` therefore re-POSTs the commit on any HTTP failure. `snapshot-retry-policy.test.ts:24` asserts `false` for 429 only because its fake is a plain object literal with no `message` and is not an `Error`. The first clause already covers genuine transport failures; the regex fallback wants a `!isAxiosError(error)` guard.

**`sync-game-cloud-save.ts:455` — the "remote changed while you played" guard is threaded through four layers and never applied**

`guard.baseRemoteHash` reaches `syncGameCloudSave` as `expectedRemoteHash` and is then only concatenated into `operationKey`. `hasRemoteChangedSinceBase` (`sync-game/policy.ts:9`) has no production caller, and `getSyncAction`'s `remoteChangedSinceAnchor` is always defaulted false since its only call site passes two args. `cloud-save-policy.test.ts:42` tests it as if it were live. The anchor-based merge still catches genuine byte divergence, so this is a missing redundant net rather than an open hole — but the guard advertises a protection that does not exist. Related: `sync-game-cloud-save.ts:208` returns a stale `remoteHash` on the untracked path (always null, even when `runFirstSync` just committed), which becomes a real bug the moment the guard is wired up.

**`events/cloud-save/resolve-cloud-save-conflict.ts:11` — destructive resolution has no concurrency token**

The IPC call carries only `(objectId, shop, resolution)`. `resolveCloudSaveConflict` re-runs `analyzeCloudSaveState` from scratch and applies the single resolution to every conflict in the fresh analysis (`:48`). If a second machine uploads between the UI rendering the conflict list and the user clicking "keep local", their choice is applied to conflicts they were never shown, discarding the other machine's data. `expectedRemoteHash` exists for exactly this and is not plumbed here.

**Automatic syncs have no `isGameRunning` guard**

`events/cloud-save/sync-game-cloud-save.ts:17` and `resolve-cloud-save-conflict.ts:26` both refuse to sync a running game; `runAutomaticCloudSaveSync` does not. `update-executable-path.ts:40` and `select-game-wine-prefix.ts:29,47` fire a bidirectional `environment-changed` sync from UI actions that are perfectly possible mid-session. Local deletion is not reachable there (the environment id changed so the anchor is null), which caps the blast radius, but a restore onto a running game's files is not great.

**`umu.ts:207` — `preparePrefix` has no timeout and sits on the launch path**

`launch-game.ts:486` awaits `prepareCompatibilityPrefixForCloudSave`, which awaits an unbounded `new Promise` that only settles on the child's `close`/`error`. `umu-run createprefix` downloading a runtime on a slow or offline network, or a wedged wineserver, leaves the launcher window open forever. This path is only entered when v2 cloud save is on, so enabling cloud saves can make a game unlaunchable.

**`events/auth/sign-out.ts:36` — new sublevels are not cleared, and the automatic-sync setting is not user-scoped**

Sign-out clears games/downloads/layout but none of the four new cloud-save sublevels. `cloudSaveAutomaticSyncSettingsSublevel` is keyed by `shop:objectId` only, so if user A disables V2 sync for a game and user B signs in on the same machine, B inherits A's opt-out. Anchor keys do embed the user id so those stay correct, but they are never pruned either.

## Smaller

- `merge-user-variant-snapshots.ts:50` and `cloud-save-v2-file-details.ts:64` — variant equality via `JSON.stringify` is key-order sensitive across the napi/API boundary. Local variants come from the napi struct (with `steamId64`/`concreteFolderId` present as own null properties for `None`), remote ones are raw server JSON that `validateSnapshotVariant` forces to exactly `["variantId","kind"]` for `kind: "default"`. A mismatch makes every sync for that game throw "Divergent Cloud Save metadata for the same variant". The Rust side compares field by field (`pipeline/mod.rs:87`); the TS side should too. The test cannot catch it because it passes the same object literal as both sides.
- `cloud-save-error-details.ts:8` — `rawErrorCode` from `error.code` shadows the `cloud_save_*` code the regex exists to extract. Axios always sets `code`, and napi surfaces `Error::from_reason` with `code: "GenericFailure"`, so a native failure like `cloud_save_invalid_restore_temp_root` reaches the renderer as `errorCode: "GenericFailure"`.
- `download-remote-snapshot-to-temp.ts:97` — blobs go to `app.getPath("temp")`, which is tmpfs on Arch/CachyOS/SteamOS, at the guessable path `<temp>/hydra-cloud-saves/<snapshotId>-<version>/<hash>.blob`. Whole snapshot is downloaded before anything is applied, so a large restore can fill RAM-backed `/tmp`, and another local user can read the blobs. Content-swap TOCTOU is fine since `prepare_target` re-hashes the staging copy.
- `restore-remote-snapshot.ts:175` — the version-change retry re-passes `requestedEntryIds` that were validated against the old manifest, so `:102` throws "Requested restore file is missing from manifest" if the other device renamed any of them. A partial restore racing a remote push becomes a hard error instead of a re-analysis.
- `upload-local-game-snapshot.ts:146` — `groupUploadsByHash` groups by `[hash, sizeBytes]` and uploads only the first member's `uploadUrl`. Correct only if the server's object key is content-addressed; nothing asserts that. Fails loudly server-side rather than corrupting, so low risk.
- `sync-anchor.ts:71` — anchor keys embed `environmentId`, and `getCloudSaveSyncAnchorForEnvironment` only ever deletes at the current key. Every prefix change, generation rotation, or executable move orphans an anchor carrying the game's full file list, permanently.
- `automatic-sync-settings.ts:42,79` and `executable-path-guard.ts:42` — read-modify-write of the whole `Game` record with awaits in between, invoked fire-and-forget from `onOpenGame`/`onCloseGame`, which concurrently write playtime to the same key. Lost update.
- `native-addon.ts:369` — the cloud-save wrappers are `return this.load().foo(...)` with no guard, so they throw synchronously despite `Promise`-typed signatures. `load()` throws at :203 when the addon is missing. All current callers are inside `async` functions so it becomes a rejection, but the declared types invite `NativeAddon.x(...).catch(...)`, which would escape.
- `sync-anchor.ts:115,131`, `cloud-save-v2-file-details.ts:144,223`, `restore-remote-snapshot.ts:214` — `localeCompare` as a sort key for JSON identity keys, hashes, and warning codes. It is ICU-dependent and can return 0 for distinct strings, so persisted anchor entry order is not reproducible across machines. Plain `<`/`>` is what you want here.
- `restore/delete_targets.rs:139` — pre-validation is properly all-or-nothing and `rollback` covers a failed rename, which is good, but a failed `remove_file` on a backup returns early and leaves `.hydra-delete-*-backup` files in the save directory permanently. Combined with the scanner not filtering `.hydra-*` residue (see my #2608 comment), that residue then gets uploaded as save data. This is the same shape as greptile's `:151` comment.
- `resolve-remote-snapshot-targets.ts:102` — the `.map` strips to `kind`/`rawPath`/`source`, so Rust's `extra_rules` (`pipeline/types.rs:25`) and `preferred_path` (`restore/types.rs:22`) have no TS counterpart. Harmless today since `manifest/rules.rs:36` always emits `preferred_path: None`, but a custom-save-path feature would silently restore to the guessed root instead of the user's chosen one.
- `cloud-save-environment.ts:83` — the prefix fingerprint includes `stat.birthtimeMs`, which on Linux without statx falls back to ctime, so a ctime change rotates the generation and orphans the anchor. Only reachable in `local-fallback` mode.
- `executeGameCloudSaveSync` calls `assertEnvironmentCurrent` four to six times per sync, and each one is a full `getCloudSaveGameContext`: Steam `loginusers.vdf` plus `userdata` reads, synchronous `Wine.validatePrefix`/`getPrefixUserProfiles` on the main thread, and `ensurePrefixGeneration`, which can write the marker file into the prefix as a side effect of what reads like a read-only assertion.

## Dead code

Several exported, unit-tested policy helpers have no production call site, which makes the decision layer's test coverage look thicker than it is:

- `run-first-sync.ts:30,53,94` — `runFirstSync` only runs when `initialState === "untracked"`, and `analyze-cloud-save-state.ts:63` sets that only when `!activeRemoteSnapshot`, so `analysis.state.activeRemoteSnapshot` is always null inside. `getFirstSyncState` can only return `"local-ahead"` or `"untracked"`; the "both sides non-empty means conflict" protection it looks like it provides never executes. The real case is handled correctly by the merge, so this reads as a safety net that is not one.
- `sync-anchor-policy.ts` — `canMigrateLegacyCloudSaveAnchor` unconditionally returns false and has no callers.
- `sync-anchor.ts:89` — `_localSnapshotHash`/`_localSnapshotFileCount` are unused but still threaded through from `analyze-cloud-save-state.ts:44`.
- `should-skip-restore-file.ts` and `NativeAddon.shouldSkipRestoreFile` are unused, and `cloud-save/index.ts:17` re-exports them alongside raw primitives (`replaceRestoreTargets`, `downloadRemoteSnapshotToTemp`) that bypass the verify-then-apply ordering if called directly.
- `operation-coordinator.ts:6` and `sync-game-cloud-save.ts:404` are two copies of the same serializer; the automatic path acquires both. No cycle so no deadlock, and the `.finally` delete does run before the chained `.then`, so no lost wakeup. One of them should still go.
- `sync-game/policy.ts:52` — `getSuggestedCloudSaveAction("partial", n)` returns `"none"` (asserted at `cloud-save-policy.test.ts:53`). A partial state by definition has unresolved entries, so the UI would offer no way to finish the sync.
- `automatic-sync.ts:180` logs a `warn` with `reason: "missing_launch_guard"` on every close of a game launched outside Hydra, which is a normal path.
- `restore-remote-snapshot.ts:51` — `current?.id === expected.id && current.version === expected.version ? current : (current ?? null)` reduces to `current ?? null`.

Hachi-R:
Follow-up after reproducing the review paths against `e72eee3`. I am narrowing the correction scope for this PR to the issues that affect correctness or can lose/suppress save data.

### Will address here

1. **Merge correctness**
   - Entries in `base.unresolvedRemoteEntryIds` will not be allowed to prove a local deletion on a later sync.
   - Snapshot variants will be compared semantically by their fields instead of with `JSON.stringify`.
   - Regression coverage will include the two-sync unresolved-entry sequence and N-API/API variant shapes.

2. **Safe local deletion**
   - Moved targets will be revalidated after the rename, at the mutation point.
   - Cleanup/recovery will not return a failure after the logical deletion has already committed, and it will not attempt an impossible rollback after backups have begun to be removed.
   - `.hydra-delete-*` artifacts will be excluded from every scanner path.
   - Failure-path tests will cover rename, revalidation and cleanup.

3. **Lifecycle safety**
   - A process-enumeration failure will skip that watcher tick instead of being interpreted as “all games closed”.
   - Pre-launch sync will distinguish skipped/offline operation from a required restore that actually failed. Only a conflict or failed required restore will block launch; an offline check before analysis will launch without enabling the post-exit upload guard.

4. **Predictable automatic-sync state**
   - An absent V2 setting will mean disabled.
   - The mode getter will be read-only.
   - Existing legacy automatic sync will remain legacy until V2 is explicitly enabled.

5. **Correct retry classification**
   - Axios errors with an HTTP response will not be treated as transport failures.
   - Tests will use real `AxiosError` instances rather than plain object mocks.

### Will not address in this PR

- I will not wire the global `expectedRemoteHash` into post-exit sync. The three-way merge already combines independent changes and conflicts same-entry divergence; a global hash check would reject safe merges. A concurrency token for explicit user conflict resolution belongs with the UI integration.
- UMU timeout handling, restore-temp relocation/permissions, sign-out cache cleanup, prefix fingerprint hardening, native wrapper cleanup, dead-code removal and custom-path `preferredPath`/`extraRules` propagation will stay with their owning follow-up cuts.
- I will not change `localeCompare`, content-hash upload grouping, environment rechecks or environment-scoped anchor retention; those are not correctness defects in this flow.

The two existing inline deletion threads remain valid and are included in item 2; I am not resolving them until the implementation and regression tests land.

native/hydra-native/src/cloud_save/restore/delete_targets.rs
```diff
@@ -0,0 +1,334 @@
...
+
+fn target_is_symlink(path: &Path) -> Result<bool, String> {
+    match std::fs::symlink_metadata(path) {
+        Ok(metadata) => Ok(metadata.file_type().is_symlink()),
+        Err(error) if error.kind() == ErrorKind::NotFound => Ok(false),
+        Err(_) => Err("cloud_save_delete_target_inspection_failed".to_string()),
+    }
+}
+
+fn backup_path(target: &Path) -> Result<PathBuf, String> {
```
Moyasee: A random UUID here means an interrupted deletion strands the user's save under a name nothing can find again.

`move_to_backups` renames every target before anything is validated or removed. If the process dies in that window, or if `rollback`'s rename fails (it discards the error with `let _ =`), the original path is gone and the bytes sit under `.hydra-delete-<uuid>-backup`:

```
original gone:      true
stranded backup at: .hydra-delete-0aa31d5e-da32-4da1-9149-80899d91656f-backup
scanner sees:       []
re-run recovers:    Some("GenericFailure, cloud_save_delete_target_missing")
bytes still on disk under uuid name: true
```

The scanner cannot see it because `is_cloud_save_artifact_path` filters it out, a re-run just reports the target missing, and no recovery routine exists for delete artifacts.

The restore path already solves exactly this: `restore_artifact_paths` derives the name from `sha256(target_key)`, and `recover_target_artifacts` runs at the top of `replace_restore_targets` and renames a stranded backup back when the target is gone. Deletion could use the same deterministic name plus an equivalent recovery pass, which would also make the `let _ =` in `rollback` survivable rather than terminal.

Moyasee (commented):
Reviewed at c7a7454 against the #2608 branch.

All five items from your follow-up landed and check out: unresolved base entries are filtered out of `baseById` so they cannot prove a deletion, `validate_moved` now runs after the rename at the mutation point, `cleanup_backups` returns a count instead of failing a committed deletion, `isValidProcessWatcherScan` skips the tick on enumeration failure, `storedV2Enabled === true` makes an absent setting mean disabled, and `snapshot-retry-policy` only treats `isAxiosError(error) && !error.response` as transport.

```
typecheck:node   clean
lint             0 errors, no new warnings in files this PR touches
yarn test        166 pass, 0 fail   (CI gates this via lint.yml)
cargo test       122 pass, 2 fail
```

Both Rust failures are inherited, not from this slice. `resolves_an_approved_custom_directory_and_can_create_it` is the one you already fixed in 6c46010e, and this branch is cut from c6a1013e so it does not have it yet. The other is the macOS-only `/var` symlink one. A rebase clears both.

One finding inline on the deletion backup naming. Not re-raising the two greptile deletion threads, since item 2 covers them and the fixes are in.

native/hydra-native/src/cloud_save/restore/delete_targets.rs
Hachi-R: Fixed in 274df46af. Delete backups now use a deterministic `.hydra-delete-<sha256(path_key(target))>-backup` name and are recovered before target validation: a missing target is restored, an identical target removes the redundant backup, and invalid or divergent artifacts preserve both sides and fail explicitly. Rollback now attempts every target, reports incomplete recovery, and leaves deterministic backups available for the next run. Parent directories are synced after rename, rollback, and removal; post-commit cleanup remains best-effort. The scanner ignores both the new SHA-256 format and legacy UUID artifacts. Added coverage for stable naming, interrupted deletion recovery, failed rollback recovery, divergent/invalid artifacts, cleanup behavior, and both scanner formats. Full Rust suite passes (125/125), along with typechecks, formatting, and lint with no errors.

Moyasee: Verified at 274df46a.

```
backup name deterministic:    true
scanner still hides artifact: true
re-run outcome:               Ok(1)     recovers, then completes the deletion
no artifact left behind:      true
divergent -> Err("cloud_save_delete_recovery_conflict"), target and backup both preserved
```

Recovering before validation rather than after is the right placement, and treating a divergent artifact as an explicit conflict instead of picking a side is better than what I suggested.

Your 125 lines up with what I see: 128 pass and 2 fail here out of 130, and 130 minus the 5 `#[cfg(unix)]` tests is 125 on Windows. Both failures are still the inherited pair, so a rebase onto 6c46010e clears them.

Moyasee (approved):
Approving.

The deletion recovery is fixed and verified at 274df46a, and all five items from your follow-up landed.

```
typecheck:node   clean
lint             0 errors, no new warnings in files this PR touches
yarn test        166 pass, 0 fail
cargo test       128 pass, 2 fail   (both inherited, cleared by a rebase onto 6c46010e)
```

Conventions look right too: no `console` in the new code, `logger.warn` on the process-watcher skip, named exports throughout, no `Array<T>`.

Scoping note so the approval is not read wider than the work: at 9155 lines I went deep on the native deletion path, the merge logic, the lifecycle changes and the conventions sweep. `sync-game-cloud-save.ts` and the LevelDB sublevels got a lighter pass, and nothing there looked wrong, but I would not claim the same depth on them.

Not blocking on the rebase, though it is the one thing left before the two Rust failures go away.

PR #2611, opened 2026-07-30
feat(cloud-save): add desktop UI

Hachi-R:
## Summary

- add the Cloud Save V2 desktop widget, manager and file comparison UI
- wire renderer and preload APIs to the existing sync core
- expose the V2 panel in game options while keeping the legacy cloud panel separate
- keep custom paths, Big Picture and cloud-save deletion out of this cut

## Validation

- npm.cmd run typecheck:node
- npm.cmd run typecheck:web


src/renderer/src/pages/game-details/cloud-save-v2/cloud-save-v2-context.tsx
```diff
@@ -0,0 +1,496 @@
...
+  };
+
+  const handleConfirmResolution = () => {
+    const resolution = pendingResolution;
+    setPendingResolution(null);
+    if (resolution) void runCloudSaveOperation(resolution);
+  };
+
+  const hasError = hasRefreshError || hasSyncError;
+  const value = useMemo<CloudSaveV2ContextValue>(
```
Moyasee, line 410: This `useMemo` can never hit, so every consumer re-renders whenever the provider does.

`openManager` (290), `runCloudSaveOperation` (311) and `setAutomaticSyncEnabled` (367) are plain arrow functions in the component body and all three are in the dependency array. There is no `useCallback` anywhere in this file, so they get a fresh identity every render, the deps always differ, and the memo hands back a new context value each time.

ESLint says it three times:

```
290:9  The 'openManager' function makes the dependencies of useMemo Hook (at line 416) change on every render
311:9  The 'runCloudSaveOperation' function makes the dependencies of useMemo Hook (at line 416) change on every render
367:9  The 'setAutomaticSyncEnabled' function makes the dependencies of useMemo Hook (at line 416) change on every render
```

These are the only new lint warnings in the PR; the `no-explicit-any` ones in `preload/index.ts` and `game-options-modal.tsx` sit on pre-existing lines.

Wrapping the three in `useCallback` makes the memo do what it looks like it does. Three components consume this context, and `cloud-save-widget.tsx` sits on the game details page, so it is a path that re-renders often.

Moyasee (commented):
Reviewed at 1bc9b30 against the #2610 branch. No other review on this one yet.

```
lint             0 errors
yarn test        184 pass, 0 fail
cargo test       119 pass, 2 fail (both inherited from the stale base, below)
```

Conventions are clean: no `console`, no `Array<T>`, no default exports, no bare `eslint-disable`. i18n is done properly, 96 new keys in `en` with every other locale file touched, and no untranslated JSX text in the new components.

### Stale base, and it has a sharp edge at #2613

This branch forks at e72eee33, two commits behind #2610, so it is missing `c7a74544` and `274df46a`. Nothing in this PR touches the native addon, so it reverts nothing by itself. The problem is what happens on rebase.

`delete_targets.rs` across the stack:

```
2610 (274df46a)  538 lines   delete_backup_path + recover_backup   Uuid::new_v4: 0
2611 / 2612      239 lines   backup_path, rollback -> ()           Uuid::new_v4: 1
2613 / 2614      439 lines   backup_path, no recovery              Uuid::new_v4: 1
```

#2612, #2613 and #2614 each sit correctly on their immediate parent, so this fork point is the only stale link, but everything above it carries the pre-fix deletion code. The version that reaches #2613 has random UUID backup names, `rollback` discarding errors and returning `()`, and no `recover_backup` pass, which is the interrupted-deletion case that strands a save under a name nothing can find.

So when this gets rebased, the `delete_targets.rs` conflict wants 2610's side, not the branch's. Worth a deliberate check rather than a routine resolution. The two `cargo test` failures I see here clear at the same time, since they are also artifacts of the stale base.

One React finding inline.

Moyasee (approved):
Approving. Both findings from my last pass are resolved at dd2aff0f.

The merge in e909ca82 puts 2610's head in this branch, so the deletion hardening is present rather than pending a careful rebase: 9 references to `delete_backup_path` and `recover_backup`, zero `Uuid::new_v4`. That also removes the sharp edge I was worried about for #2613.

`dd2aff0f` wraps the three handlers in `useCallback`, so the context `useMemo` does something now and the three exhaustive-deps warnings are gone.

```
lint             0 errors, no warnings in cloud-save-v2-context.tsx
yarn test        191 pass, 0 fail
cargo test       128 pass, 2 fail
```

The two Rust failures are both inherited and neither belongs to this slice: `resolves_an_approved_custom_directory_and_can_create_it` is fixed in #2608 at 6c46010e and arrives once #2610 picks that up, and `preserves_literal_backslashes_in_filesystem_paths` is the macOS-only `/var` symlink one.

PR #2612, opened 2026-07-30
feat(cloud-save): add variants, validation, and custom paths

Hachi-R:
## Summary

- add custom path binding, persistence, approval, rebind, removal and manual sync flows
- preserve variant-aware snapshot merging, unresolved destinations and Steam account context
- validate custom path overlaps and upload limits across native and Node layers
- expose custom path management and approval flows in the desktop Cloud Saves V2 UI
- prepare Linux compatibility prefixes before restore and sync operations

## Context

This PR is stacked on #2611 and supersedes #2582. Big Picture support and cloud-save deletion remain in later cuts.

## Validation

- `npm.cmd run typecheck:node`
- `npm.cmd run typecheck:web`
- `cargo check`
- focused Node tests: 117 tests, 115 passed and 2 platform-specific skips
- `cargo test custom_path_overlap` (6 passed)
- `git diff --check`


Moyasee (commented):
Reviewed at 7edf7ec against the #2611 branch. No other review on this one yet.

```
lint             0 errors, no new warnings of its own
yarn test        260 pass, 2 fail  (both macOS-only, see below)
cargo test       125 pass, 2 fail  (both inherited from the stale base in #2611)
```

Conventions clean across 78 changed TS files: no `console`, no `Array<T>`, no default exports, no `eslint-disable`.

The `preferredPath` gap I raised back on #2606 is closed here the way you said it would be. `customPathToCloudSaveRule` builds it from a stored `CloudSaveCustomPath`, and that only exists after going through the encode/validate path, so the native side never sees an unvalidated value in the production flow.

The two `yarn test` failures are macOS-only and I do not think they are worth changing. `custom-path.test.ts` derives `platform` from `process.platform`, so a Mac gets `"mac"`, and `/private` is in the mac-only protected roots while `os.tmpdir()` resolves through `/var` to `/private/var`:

```
canonicalizes an existing selected directory        -> cloud_save_custom_path_protected
keeps an approved missing destination valid ...     -> cloud_save_custom_path_symlink_ancestor
```

On Linux the platform is `"linux"`, `/tmp` is neither protected nor a symlink, and both pass. Since Hydra does not ship macOS builds this is not a target defect, so I am only noting it because it means the suite is red for anyone reviewing on a Mac, same as the two Rust ones.

Nothing else to flag. The protected-roots list looks right for Windows (per-drive `Windows`, `Program Files`, `Program Files (x86)`, `ProgramData`, plus the `SystemRoot`/`ProgramFiles` environment values), and the lint warnings that show up are `cloud-save-v2-context.tsx` from #2611 and pre-existing `any` in `preload/index.ts`.

Moyasee (commented):
Reviewed against `codex/cloud-saves-v2-desktop-ui`, not main.

The security envelope on custom paths holds up. `decodeCloudSaveCustomPath` re-encodes and rejects non-canonical raw paths, `assertSafeRelativeSegments` blocks traversal and glob metacharacters, `assertNotProtected` covers system roots, and `assertNoSymlinkAncestor` realpaths the deepest existing ancestor before any write. Concurrency is handled properly too: `isGameRunning` and `isCloudSaveSyncActive` are re-checked after the directory dialog returns and again inside `assertCurrentBindings`, under the store mutation coordinator.

Two things are broken. Removing a custom path that holds every tracked file reports success while never touching the remote, and the removal flow unregisters before syncing with no rollback and a failure toast that states the opposite of what happened. Details inline.

What is left after those: the launch path now pays for two full analyses, `analyzeCloudSaveState` gave up its local/remote parallelism, and the Rust overlap check recompiles globs per candidate. Also `remote-target-mapped` collapses into the wrong user-facing string, `cloud_save_v2_path_approval_manual_sync_description` is a byte-identical copy of its sibling with wording that does not fit the flow, and only `en` and `pt-BR` ship translations (`ru` and `es` are at 6 `cloud_save_v2` keys against 156, which matches the rest of the stack rather than starting here).

Test coverage is real, 14 files and roughly 1900 lines, and the encode/decode round-trip and binding migration cases are well chosen. Two gaps line up with the blocking findings: no test for a removal whose merge ends up empty against a live snapshot, and none for a removal whose sync throws after `unregister()` has already run.

src/main/services/cloud-save/custom-path-removal.ts
```diff
@@ -0,0 +1,113 @@
...
+}: {
+  isRegistered: boolean;
+  unregister: () => Promise<void>;
+  sync: () => Promise<SyncGameCloudSaveResult>;
+}) => {
+  if (isRegistered) {
+    await unregister();
+  }
+
+  const result = await sync();
```
Moyasee, line 48: Blocking. `unregister()` runs before `sync()`, and nothing puts the binding back if the sync throws.

The renderer catches everything and shows `cloud_save_v2_custom_path_remove_error_description`: "The location is still being tracked and its local files were not changed." The first half is false, the binding is already gone.

Either re-register in a catch, or sync first and unregister after, or change the string to match what actually happened.

src/main/services/cloud-save/sync-game/transfer.ts
```diff
@@ -53,7 +53,6 @@ export const uploadLocalState = async (
       assertEnvironmentCurrent,
     }
   );
-  if (!snapshot) throw new Error("Local cloud save snapshot is empty");
   return snapshot;
```
Moyasee, line 56: Blocking, together with the unconditional `mergedAggregateHash` in `sync-game-cloud-save.ts`. Dropping the `if (!snapshot) throw` lets a null upload flow on silently, and there is a path where that reports success on a sync that did nothing.

Remove the only custom path on a game whose mapped rules resolve to nothing locally:

- `excludeCloudSaveRawPathsFromMerge` empties `merge.files`
- `mergedAggregateHash` is now computed unconditionally, so `proposalChanged` is true
- `activeSnapshot` exists, so the new `!activeSnapshot && merge.files.length === 0` early return does not fire
- `executeAppliedSync` calls `uploadLocalState`, which reaches `create-remote-snapshot-from-local-state.ts:112` `if (files.length === 0) return null`
- with this throw gone the null propagates, `committedSnapshot` is null, `proposalChanged` is true, so `saveAppliedSyncAnchor` writes no anchor
- result is `finish("upload", getPostSyncState(true, ...))` = `local-ahead`
- `executeCloudSaveCustomPathRemoval` only rejects `finalState === "conflict"`, so the renderer shows `cloud_save_v2_custom_path_removed`: "The custom location was removed and synchronized."

Nothing was synchronized. The anchor is stale, the remote snapshot still holds the removed files, and every later sync repeats the same no-op, so the game sits at `local-ahead` with nothing that can move it.

The `fileCount < 1` to `fileCount < 0` relaxations in `validateRemoteSnapshotSummary` and `validateCommitResponse` suggest empty snapshots were meant to be supported, but the client bails before it can ever request one. Since the PR body says deletion is a later cut, the honest fix is to detect an empty proposal against an existing snapshot explicitly and return `partial`, or refuse the removal, rather than report success.

src/main/services/cloud-save/sync-game-cloud-save.ts
```diff
@@ -375,24 +453,148 @@ const executeGameCloudSaveSync = async (
...
+    registeredExcludedRawPaths,
+    analysis
+  );
+  const initialState = analysis.state.state;
+  const merge = excludeCloudSaveRawPathsFromMerge(
+    analysis,
+    resolvedMerge(analysis, resolution),
+    excludedRawPaths
+  );
+  const mergedAggregateHash = NativeAddon.buildSnapshotAggregateHash({
```
Moyasee, line 507: This lost the `merge.files.length > 0 ? ... : null` guard, so `proposalChanged` is now true for an empty merge instead of comparing against null. See the note on `sync-game/transfer.ts` for where that lands.

src/main/helpers/launch-game.ts
```diff
@@ -507,6 +625,38 @@ export const launchGame = async (
         return null;
       })
     : null;
+  const customPathApproval =
+    prefixReadyForRestore && cloudSaveContext
+      ? await createPendingCloudSaveCustomPathApproval(
```
Moyasee, line 630: Every launch on the V2 path now runs the full cloud-save analysis twice. This call goes to `analyzeCloudSaveState`, then `runAutomaticCloudSaveSyncDetailed` a few lines down goes through `syncGameCloudSave` to `analyzeCloudSaveState` again. Each pass is `listRemoteGameSnapshots` plus `getRemoteSnapshotRestoreManifest` plus `buildLocalGameSnapshotContext`, so two round trips and two full native tree walks with hashing before the game starts.

The approval pass returns null in the common case and its analysis is thrown away. Either thread that analysis into the sync, or gate the probe on whether the remote manifest contains any `<custom>` rawPath at all.

src/main/services/cloud-save/analyze-cloud-save-state.ts
```diff
@@ -30,6 +31,16 @@ export const analyzeCloudSaveState = async (
   ) {
     throw new Error("Active Cloud Save snapshot belongs to another game");
   }
+  const scanStoreUserContext = storeUserContextWithSnapshotAccounts(
+    context.pathContext.storeUserContext,
+    remoteManifest?.variants ?? []
+  );
+  const localSnapshotContext = await buildLocalGameSnapshotContext(
```
Moyasee, line 38: This used to sit in the same `Promise.all` as `listRemoteGameSnapshots`. Now the local scan waits on both the snapshot list and the manifest fetch, because `scanStoreUserContext` needs `remoteManifest.variants`.

Only the extra-accounts case actually needs the variants. The local build could start against the local `storeUserContext` in parallel as before, and re-scan only when the manifest introduces accounts the device does not already know.

native/hydra-native/src/cloud_save/custom_path_overlap.rs
```diff
@@ -0,0 +1,536 @@
...
+        paths.push(current.clone());
+        let Some(parent) = parent_path(&current) else {
+            break;
+        };
+        current = parent;
+    }
+    paths
+}
+
+fn pattern_matches(pattern: &str, value: &str, case_sensitive: bool) -> bool {
```
Moyasee, line 112: `GlobBuilder` is constructed and compiled on every call. `pattern_covers_path` calls this once per ancestor of the selected path, and `pattern_can_match_in_subtree` calls it per pattern segment inside the recursion, so a deep path against a rule set with several resolved paths each is hundreds of glob compiles per check. `partitionCloudSaveCustomPathBindingsByOverlap` then runs one check per binding, and that whole thing now sits inside `buildLocalGameSnapshotContext` on every analysis.

Compile each candidate pattern once and reuse the matcher across ancestors and segments.

native/hydra-native/src/cloud_save/custom_path_overlap.rs
```diff
@@ -0,0 +1,536 @@
...
+            .any(|candidate| mapped_rule_overlaps(&selected_path, candidate))
+        {
+            return Ok(overlap(MAPPED_LOCATION_OVERLAP, rule.raw_path.clone()));
+        }
+
+        if remote_targets
+            .iter()
+            .any(|target| remote_target_is_mapped(target, rule))
+        {
+            return Ok(overlap(REMOTE_TARGET_MAPPED, rule.raw_path.clone()));
```
Moyasee, line 294: `remote-target-mapped` never reaches the user as itself. `overlapErrorCode` in `custom-path-overlap.ts` maps anything that is not `custom-location-overlap` to `cloud_save_custom_path_mapped_location_overlap`, which reads "Hydra already recognizes the saves in this destination automatically."

That is a different cause. Here the selected folder is fine, but a remote file would land on a mapped location once restored into it. Either add a string for it or drop the constant and return `MAPPED_LOCATION_OVERLAP`.

src/locales/en/translation.json
```diff
@@ -512,6 +512,27 @@
     "cloud_save_v2_confirm_remote_title": "Use cloud saves?",
     "cloud_save_v2_confirm_remote_description": "The local saves on this device will be replaced by the active cloud version.",
     "cloud_save_v2_cancel": "Cancel",
+    "cloud_save_v2_path_approval_title": "Confirm where your saves will be restored",
+    "cloud_save_v2_path_approval_description": "We found cloud saves created on another device.<br />Since their location has not been confirmed on this computer yet, Hydra needs your approval before opening the game.",
+    "cloud_save_v2_path_approval_manual_sync_description": "We found cloud saves created on another device.<br />Since their location has not been confirmed on this computer yet, Hydra needs your approval before opening the game.",
```
Moyasee: Byte-identical to `cloud_save_v2_path_approval_description` above it, and it still says "before opening the game" even though the manual-sync flow is not launching anything. Same duplication at `pt-BR/translation.json:369-370`.

src/main/services/cloud-save/resolve-remote-snapshot-targets.ts
```diff
@@ -99,10 +121,24 @@ export const resolveRestoreManifestTargets = async (
...
+    winePrefixPath: effectiveWinePrefixPath,
+    approvedRules: [
+      ...approved.rules.map(({ kind, rawPath, source }) => ({
+        kind,
+        rawPath,
+        source,
+        preferredPath: undefined,
+      })),
+      ...customPaths.map(customPathToCloudSaveRule),
+    ].map(({ kind, rawPath, source, preferredPath }) => ({
```
Moyasee, line 132: The array above is built with `preferredPath: undefined` on the approved rules only so this second map can pick the same four fields back off. Collapse into one map.

src/main/services/cloud-save/restore-remote-snapshot.ts
```diff
@@ -34,12 +34,100 @@ import {
...
+      .map(({ rawPath }) => rawPath)
+      .filter((rawPath) => rawPath.startsWith(CLOUD_SAVE_CUSTOM_PATH_PREFIX))
+  );
+  if (restoredCustomRawPaths.size === 0) return;
+
+  const customPathContext =
+    cloudSaveCustomPathContextFromPathContext(pathContext);
+  const boundCustomPaths = [...restoredCustomRawPaths]
+    .map((rawPath) => {
+      const target = actions.find((action) => action.rawPath === rawPath);
```
Moyasee: `actions.find(...)` inside a map over a Set that was derived from `actions`. Build `Map<rawPath, action>` once above the loop.

src/main/services/wine-prefix.ts
```diff
@@ -9,6 +11,77 @@ const systemProfileNames = new Set([
...
+  let end = value.length;
+  while (end > 0 && value[end - 1] === "/") end -= 1;
+  return value.slice(0, end);
+};
+
+export const getWinePrefixWindowsUserProfilePath = (
+  winePrefixPath: string
+): string | null => {
+  try {
+    const userReg = fs.readFileSync(
```
Moyasee, line 30: `readFileSync` here plus two `realpathSync` calls in `getWinePrefixUserProfile` below, all on the main process. `cloudSaveCustomPathContextFromPathContext` reaches this on the launch path, so it blocks the event loop while the prefix is read. `.cursorrules` asks for async/await here.

Hachi-R:
## Review follow-up

### Addressed

- Added a distinct surfaced error for `remote-target-mapped` overlaps, with dedicated English and Portuguese copy. Desktop and the stacked Big Picture surface now preserve this reason instead of presenting it as a regular mapped-location overlap.
- Updated the manual-sync approval copy to say that approval is required before synchronizing, while keeping the pre-launch copy unchanged.
- Collapsed the redundant restore-rule projection into a single map over the combined approved and custom rules.
- Replaced the repeated `actions.find(...)` lookup with a single `Map` keyed by `rawPath`, explicitly preserving the first action for duplicate raw paths.

The main fix is in `24258de15`. The Big Picture follow-up is in `3ebcdd3a6`, and both were propagated through the stack up to #2647 with normal merge commits.

### Already addressed downstream

The removal-ordering and final-custom-path/empty-snapshot threads are already covered in the downstream reconciliation work in #2646:

- `6c54c4682` publishes the remote removal before deleting the local binding.
- `d7d99c209` adds explicit remote snapshot update/deletion handling for custom-path removal.
- `4d88e699a` handles the last declared path, remaining empty tracked paths, and orphaned variants correctly.

Those threads remain open here because GitHub does not automatically resolve review threads across stacked PRs.

### Intentionally unchanged

- **Duplicate launch analysis:** sharing or caching the analysis requires changing the pre-launch approval/sync data flow and its invalidation rules, so it is not a safe mechanical follow-up.
- **Local/remote scan parallelism:** restoring the old parallelism requires a conditional re-scan policy when the remote manifest introduces additional store accounts; that is a behavioral performance change.
- **Glob compilation reuse:** this requires restructuring the native matcher lifecycle and should be handled as a focused native performance change.
- **Async Wine-prefix reads:** converting these calls requires an async API and caller-chain migration on the launch path, which changes sequencing and is broader than this review cleanup.

### Validation

- Targeted custom-path overlap tests: **4 passed**.
- `typecheck:node`: passed.
- `typecheck:web`: passed.
- Prettier full-repository check: passed.
- ESLint on the changed TypeScript/TSX files: passed.
- Final Node suite on the top of the stack: **387 tests; 382 passed, 4 platform skips, 1 known Windows `EBUSY` cleanup failure in `notification-icon.test.ts`**. The same cleanup failure reproduced in isolation and is unrelated to the changed files.
- Stack ancestry and GitHub mergeability: #2612 through #2647 are synchronized and mergeable.
- PR checks at the time of this comment: SonarCloud and Windows build passed; lint and text-integrity were still queued after a 15-minute wait; Ubuntu build was cancelled by workflow concurrency rather than failing.

src/main/services/cloud-save/custom-path-removal.ts
Moyasee: Still open at 24258de15. `executeCloudSaveCustomPathRemoval` continues to `await unregister()` before `await sync()` with nothing restoring the binding when the sync throws.

Flagging that it is fixed downstack rather than here: #2645 deletes this file outright and replaces it with `untrackCloudSaveCustomPath`, and #2646 then splits remote publish from local removal so a failure on either side reports distinctly. Since this PR targets `codex/cloud-saves-v2-desktop-ui` and not main, the broken window never ships.

I am not asking you to re-fix it here. Worth saying explicitly in the PR body that this one is resolved upstack, so it does not read as an unanswered finding.

src/main/services/cloud-save/sync-game/transfer.ts
Moyasee: Same status as the removal thread: still open at 24258de15, resolved downstack.

`uploadLocalState` still returns a possibly-null snapshot with no throw, and `mergedAggregateHash` at `sync-game-cloud-save.ts:507` is still computed unconditionally. #2645 adds `requireCommittedCloudSaveSnapshot` and wires it into both upload sites, which closes it before the stack reaches main.

Moyasee (commented):
Re-reviewed at 24258de15.

The security envelope on custom paths still holds, same conclusion as my earlier pass: `decodeCloudSaveCustomPath` re-encodes and rejects non-canonical raw paths, `assertSafeRelativeSegments` blocks traversal and glob metacharacters, and `assertNotProtected` covers the system roots.

Two of my earlier findings are still live in this PR, and both are fixed downstack in #2645 rather than here. Details on those two threads. Since this PR merges into `codex/cloud-saves-v2-desktop-ui` and not main, neither defect ever reaches users, so I am treating fixed-downstack as acceptable rather than asking for the fixes to move. Worth stating that in the PR body so they do not read as unanswered.

Separately: none of the twelve inline comments here have replies, though the head commit says it addresses remaining feedback. Hard to tell from the outside which ones were intentionally deferred and which were missed. A one-line reply per thread would make the next pass much cheaper, and on #2613 exactly this is what left another reviewer's block sitting there stale.

CI needs a re-run before anyone calls it green: `build (ubuntu-latest)` is CANCELLED and neither `lint` nor `text-integrity` reported.

Validation at 24258de15:

```
Rust:      136 passed; 0 failed
Node:      275 passed; 0 failed; 0 skipped
typecheck: node + web pass
prettier:  pass
Sonar:     gate SUCCESS, 2 open CODE_SMELL in launch-game.ts
```

Moyasee (approved):
Approving at 24258de15.

Two CI signals never reported on this head, so I ran them locally rather than approve around them:

```
eslint            0 errors, 81 warnings (all the pre-existing no-explicit-any class)
text-integrity    no CR characters in any tracked file
                  LICENSE sha256 32619612c2e0223e86c4908747ec14bef64c3c423fee80910c1aa944769b66f9, matches
Rust              136 passed; 0 failed
Node              275 passed; 0 failed; 0 skipped
typecheck         node + web pass
prettier          pass
Sonar             gate SUCCESS
```

`build (ubuntu-latest)` is still CANCELLED on GitHub's side. Worth a re-run before merge, though `build (windows-2022)` passed and both targets build from the same source.

Since the head commit says it addresses remaining feedback without replies on the threads, I went back through my earlier comments to see which actually landed. Five are fixed: the double cloud-save analysis on the launch path, `actions.find` inside the map in `restore-remote-snapshot.ts`, the redundant second map in `resolve-remote-snapshot-targets.ts`, `remote-target-mapped` now reaching the user through `custom-path-overlap-error.ts`, and the duplicated approval string, which now ends "before synchronizing" rather than repeating "before opening the game".

Three are still open, all efficiency rather than correctness, and none worth holding this on:

- `analyze-cloud-save-state.ts` still serializes the local scan behind the manifest fetch, because `scanStoreUserContext` needs the remote manifest.
- `custom_path_overlap.rs` still builds and compiles a `GlobBuilder` per call, once per ancestor and once per pattern segment.
- `wine-prefix.ts` still does `readFileSync` plus two `realpathSync` calls on the main process, reached from the launch path.

The two blocking ones are covered on their own threads: still live here, closed in #2645 before anything reaches main.

The security envelope is unchanged from my earlier pass and still holds. `decodeCloudSaveCustomPath` re-encodes and rejects non-canonical raw paths, `assertSafeRelativeSegments` blocks traversal and glob metacharacters, and `assertNotProtected` covers the system roots.

PR #2613, opened 2026-07-30
feat(cloud-save): add cloud save deletion

Hachi-R:
## Summary

- delete remote snapshots and validated local save files/directories
- clear per-game local cloud-save state and restore the automatic-sync setting
- serialize deletion against sync and custom-path mutations
- add desktop deletion confirmation and empty-state UI

## Validation

- `npm run typecheck:node`
- `npm run typecheck:web`
- 29 targeted TypeScript tests
- 7 native deletion tests

## PR chain

Depends on #2612.


native/hydra-native/src/cloud_save/restore/delete_targets.rs
```diff
@@ -68,6 +68,110 @@ fn backup_path(target: &Path) -> Result<PathBuf, String> {
     Ok(parent.join(format!(".hydra-delete-{}-backup", Uuid::new_v4())))
```
Moyasee: Random UUID backup names with no recovery pass. This is the pre-fix version from before 274df46a on #2610.

If the process dies between the rename loop and the removal loop, the save sits at `.hydra-delete-<uuid>-backup`, the scanner filters it out, a re-run reports `cloud_save_delete_target_missing`, and nothing ever looks for it again. I confirmed that on #2610 before you fixed it there:

```
original gone:      true
scanner sees:       []
re-run recovers:    Some("cloud_save_delete_target_missing")
bytes still on disk under uuid name: true
```

#2610 now derives the name from `sha256(path_key(target))` and runs `recover_backup` before validation: it restores a missing target, drops a redundant identical backup, and fails explicitly with `cloud_save_delete_recovery_conflict` when the two diverge. None of that is here, purely because the stack forks two commits earlier.

native/hydra-native/src/cloud_save/restore/delete_targets.rs
```diff
@@ -151,6 +256,23 @@ pub async fn delete_local_save_targets(
             .map_err(|_| Error::from_reason("cloud_save_delete_cleanup_failed"))?;
```
Moyasee: Cleanup failure returns an error after every target has already been renamed away, with no rollback. This is greptile's original P1 on #2610 and it is back.

By this point the deletion is logically committed. A failure removing any backup leaves the caller with an error, the originals gone from their paths, and backups orphaned under UUID names. The #2610 version returns a `cleanup_failure_count` and treats post-commit cleanup as best-effort.

Also missing: there is no revalidation between the rename and the removal. #2610 added `validate_moved`, which rechecks size and hash on the moved file at the mutation point so a save that changed after `prepare` does not get deleted anyway. That was greptile's second P1.

Moyasee (commented):
Reviewed at 3399a4c against the #2612 branch. No other review on this one yet.

```
lint             0 errors, no new warnings of its own
yarn test        276 pass, 2 fail  (the macOS-only custom-path pair inherited from #2612)
cargo test       129 pass, 2 fail  (inherited from the stale base at #2611)
```

Conventions clean across 28 changed TS files.

This slice reintroduces the deletion implementation, and because the stack forks at e72eee33 it is the pre-fix version. All three deletion fixes that landed in #2610 are absent. Detail inline.

The new directory pruning looks careful: `remove_dir` rather than `remove_dir_all` so only empty directories go, `DirectoryNotEmpty` tolerated, symlinks skipped, deepest-first ordering, ancestors bounded by `target_is_within_root`. One thing I would look at twice is that `collect_target_ancestors` inserts the restore root itself, so an emptied custom save directory gets removed too. Low harm since it is empty by then, but it may not be what a user expects of a directory they picked themselves.

src/main/services/cloud-save/delete-game-cloud-save-data.ts
```diff
@@ -0,0 +1,109 @@
...
+          withCloudSaveCustomPathStoreMutation(
+            shop,
+            objectId,
+            async (storageKey) => {
+              customPathStorageKey = storageKey;
+              await operation();
+            }
+          ),
+        assertGameNotRunning,
+        prepareLocalDeletion: async () => {
```
chubbygrannychaser: `prepareLocalDeletion()` still reads the analysis and bindings before `runWithLocalStateLock` acquires the custom-path mutation lock. A select/rebind that already passed its deletion-active check can finish between this snapshot and the lock, adding a binding/files that are absent from `cleanupRootPaths`. The later `clearCloudSaveLocalState` removes that binding, leaving those local save files untracked and undeleted. Build the analysis/bindings and deletion targets under the same mutation lock (or revalidate after acquiring it) so target discovery, deletion, and state clearing use one consistent custom-path snapshot.

chubbygrannychaser (dismissed):
The deletion target snapshot is still captured before the shared custom-path mutation lock.

src/main/services/cloud-save/delete-game-cloud-save-data.ts
Moyasee: This is addressed at 5767e5358, flagging it since the request is still blocking the stack.

`prepareLocalDeletion()` and `runWithLocalStateLock` no longer exist. `runWithLocalDeletionSnapshot` now wraps the whole critical section in `withCloudSaveCustomPathStoreMutation`, and the analysis, `localEntryIds` and `cleanupRootPaths` are all built inside that callback from the `bindings` the lock hands out.

The mutual exclusion is real, not incidental: `withCloudSaveCustomPathStoreMutation` and `mutateStoredEntriesByKey` (which `registerCloudSaveCustomPaths` goes through) share the same `storeMutationCoordinator` and derive the same `cloudSaveCustomPathStorageKey(userId, shop, objectId)`. So target discovery, the remote delete, the local delete and `clearCloudSaveLocalState` all run under one hold, and a concurrent select or rebind queues behind it instead of interleaving.

Could you re-review when you get a chance?

native/hydra-native/src/cloud_save/restore/delete_targets.rs
Moyasee: Closed. `delete_backup_path` now derives the name as `SHA256(path_key(target))` rather than a random id, so a later run recomputes the same path, and `recover_backup` is the recovery pass: it renames an orphaned backup back when the target is missing, drops the backup when the target already matches on size and hash, and refuses with `cloud_save_delete_recovery_conflict` otherwise.

native/hydra-native/src/cloud_save/restore/delete_targets.rs
Moyasee: Closed. `cleanup_backups` returns a `u32` failure count instead of erroring, the count is folded into `cleanup_failure_count`, and the commit point is stated in code. An applied deletion can no longer be turned back into a failure.

src/main/services/cloud-save/delete-game-cloud-save-data-policy.ts
```diff
@@ -0,0 +1,60 @@
...
+  deleteRemoteSnapshots,
+}: DeleteGameCloudSaveDataDependencies) => {
+  let pendingPhase = await beginPendingDeletion();
+  try {
+    await runWithLocalDeletionSnapshot(
+      async ({ deleteLocalFiles, clearLocalState }) => {
+        assertGameNotRunning();
+        pendingPhase = "remote-started";
+        await markRemoteDeletionStarted();
+        await deleteRemoteSnapshots();
```
Moyasee: `pendingPhase` is set to `remote-started` before `markRemoteDeletionStarted()` resolves, so if that write itself fails, the catch skips the `prepared` rollback and leaves a pending marker behind even though nothing was deleted remotely. The marker then blocks syncs via `assertCloudSaveDeletionNotPending` until the user retries the delete.

That is the safe direction, over-assuming the remote delete began rather than under-assuming it, and `beginCloudSavePendingDeletion` returning the existing phase means a retry resumes cleanly. So I read it as deliberate. It just is not visible as deliberate from the code, and the obvious "fix" of moving the assignment after the await would silently make it unsafe. Worth a line in the commit message so nobody tidies it later.

Moyasee (commented):
Re-reviewed at 5767e5358. Nothing blocking from my side.

Both findings I raised on `delete_targets.rs` are closed, details on those threads. The deletion sequencing reads correctly now: backups are validated before the commit point, cleanup cannot demote an applied deletion to a failure, and `recover_backup` handles the interrupted case deterministically.

`prune_empty_target_ancestors` is the part I looked hardest at, since it is the one place this PR can remove a directory the user chose. It holds up. `collect_target_ancestors` breaks on the root key so the root is never collected, only `remove_dir` is used so a non-empty directory survives, symlinks and non-directories are skipped, and the deepest-first sort makes nested empties collapse bottom-up. `removes_empty_target_ancestors_without_deleting_root` pins the important half of that.

One ordering note inline, not blocking.

The `CHANGES_REQUESTED` on this PR is stale. I have replied on that thread with what changed, but it needs the original reviewer to clear it, and until then it blocks everything stacked above.

Validation at 5767e5358:

```
Rust:      141 passed; 0 failed
Node:      304 passed; 0 failed; 0 skipped
typecheck: node + web pass
prettier:  pass
Sonar:     gate SUCCESS, 2 open CODE_SMELL
```

Hachi-R:
## Review follow-up

### Addressed

- Added an inline comment in `3ed70723a` documenting why `pendingPhase` is advanced to `remote-started` before awaiting `markRemoteDeletionStarted()`. This is intentionally conservative: an ambiguous persistence failure keeps deletion quarantined instead of clearing a phase that may already have been saved. No behavior changed.

### Intentionally unchanged

- The previous custom-path snapshot-before-lock finding is already fixed at the current head. `runWithLocalDeletionSnapshot` builds the analysis, bindings, deletion targets, remote/local deletion, and local-state cleanup under the shared custom-path mutation lock. The old `CHANGES_REQUESTED` review is stale and still needs the original reviewer to re-review or dismiss it.
- Deterministic backup recovery, mutation-point revalidation, and best-effort post-commit cleanup are already present. Their corresponding threads are resolved and outdated, so no additional code changes were made.
- Remote-boundary failures intentionally keep the pending-deletion quarantine so a later retry can resume safely. Moving the in-memory assignment after the persistence `await` would weaken that guarantee, so the ordering was documented rather than changed.

### Validation

- Targeted deletion tests: 15 passed, 0 failed.
- ESLint on the changed file: passed.
- Prettier on the changed file: passed.
- Pre-push checks: Node/Web typechecks passed; ESLint completed with 0 errors (81 existing warnings).
- The update was propagated through the stack, and #2613, #2614, #2645, #2646, and #2647 are all mergeable.

src/main/services/cloud-save/delete-game-cloud-save-data-policy.ts
```diff
@@ -0,0 +1,60 @@
...
+  await setAutomaticSyncEnabled(false);
+
+  let operationError: unknown;
+  try {
+    const deleteLocalFiles = await prepareLocalDeletion();
+    await runWithLocalStateLock(async () => {
+      assertGameNotRunning();
+      await deleteRemoteSnapshots();
+      assertGameNotRunning();
+      await deleteLocalFiles();
```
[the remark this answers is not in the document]
Moyasee: I don't think this one holds, at least not the second half.

The pending-deletion marker introduced in this PR keeps the deletion quarantined. When the second running check throws, `clearPendingDeletion()` is never reached, so the marker stays at `remote-started` and both sync paths refuse to run: `syncGameCloudSave` gates on `assertCloudSaveDeletionNotPending`, and `runAutomaticCloudSaveSyncDetailed` bails out through `isPendingDeletionBlockingAutomaticSync`. A post-exit sync cannot upload the surviving saves back.

Traced the call order to be sure:

```
beginPendingDeletion
assertGameNotRunning#1
markRemoteDeletionStarted
deleteRemoteSnapshots
assertGameNotRunning#2   <- throws
                         clearPendingDeletion  not called
                         deleteLocalFiles      not called
```

The first half is accurate: remote snapshots are gone while local files remain, and the user sees an error. That state is intentional though. It is quarantined rather than partial, and retrying the delete resumes from `remote-started` and finishes the local half.

src/main/services/cloud-save/delete-game-cloud-save-data-policy.ts
```diff
@@ -0,0 +1,62 @@
...
+  assertGameNotRunning,
+  deleteRemoteSnapshots,
+}: DeleteGameCloudSaveDataDependencies) => {
+  let pendingPhase = await beginPendingDeletion();
+  try {
+    await runWithLocalDeletionSnapshot(
+      async ({ deleteLocalFiles, clearLocalState }) => {
+        assertGameNotRunning();
+        // Advance in memory first so an ambiguous persistence failure keeps
+        // deletion quarantined instead of rolling back a potentially saved phase.
```
Moyasee: Could this come out? The reasoning is good, it just belongs in the commit body rather than in the file. Not blocking on its own.

Naming it works better than explaining it, since the whole point is that the assignment happens before the await:

```ts
const advanceToRemoteStarted = async () => {
  pendingPhase = "remote-started";
  await markRemoteDeletionStarted();
};
```

Then the call site reads `await advanceToRemoteStarted();` and "advance" carries the ordering. The other option is naming the guard in the catch, something like `const rollbackIsSafe = () => pendingPhase === "prepared"`, which puts the meaning where it actually gets used.

Moyasee (commented):
Took another pass at 3ed70723a. Nothing new that blocks, one small thing inline.

Also left a note on the greptile P1 thread. I traced the call order and the surviving-saves half of it does not reproduce, since the pending marker blocks both sync paths.

The `CHANGES_REQUESTED` on this PR is still the thing holding it, and everything stacked above. It is stale, the code it describes was restructured, and I replied on that thread earlier with the details. It just needs the original reviewer to clear it.

Once the comment below is fixed we can approve and merge this.

Moyasee (approved):
LGTM

PR #2614, opened 2026-07-30
feat(cloud-save): add Big Picture cloud save support

Hachi-R:
## Summary

- add a Big Picture cloud save context and provider for game pages
- expose cloud save status and actions through the game hero
- support conflict resolution, custom path approval, snapshot browsing, upload and deletion flows
- route blocked launches to the active game window, including Big Picture

## Why

This completes the cloud saves v2 PR chain by adding the Big Picture consumer after the native, sync, desktop, variants and deletion layers are available.

## Validation

- `npm run typecheck:node`
- `npm run typecheck:web`
- 2 focused Big Picture presentation tests

## PR chain

Depends on #2613.


Moyasee (approved):
Reviewed at 60363db against the #2613 branch. No other review on this one yet.

```
lint             0 errors, no new warnings
yarn test        278 pass, 2 fail  (the macOS-only custom-path pair from #2612)
cargo test       129 pass, 2 fail  (inherited from the stale base at #2611)
```

Nothing to flag in this slice.

Conventions clean across 20 changed TS files: no `console`, no `Array<T>`, no default exports, no `eslint-disable`. It adds no new locale keys and none are needed, because the Big Picture components reuse the keys #2611 introduced and every one of them goes through `t()`. No untranslated JSX text in the nine new `cloud-save-v2` components.

Worth noting since it is the same shape of code: the Big Picture `cloud-save-v2-context.tsx` wraps its handlers in `useCallback` and does not have the dead `useMemo` problem I flagged on the desktop context in #2611. Two parallel implementations, and this one got it right.

The lint warnings that surface in `cloud-tab.tsx` and `game-settings-modal/index.tsx` are on pre-existing lines. This PR adds no `any` of its own to either file.

Approving on the content of this diff. That is scoped to this slice and is not a statement about the stale fork point at #2611 or the deletion regression at #2613, both of which still want resolving before the stack lands.

PR #2617, opened 2026-07-31
fix: extract archives in paths containing wildcard characters

Moyasee:
Closes [LBX-818](https://linear.app/los-broxas/issue/LBX-818/auto-extraction-failing-in-folders-with-asterisk)

**Fill in the PR content:**

Auto-extract fails when the download folder has a `*` in its name. The download finishes fine, then extraction ends with "Hydra could not extract this archive". There are two separate causes, both from 7-Zip giving `*` a meaning in paths we hand it as plain arguments.

**The output directory.** 7-Zip substitutes the archive base name for a `*` appearing anywhere in the `-o` switch. Extracting `/NVME/Teste*/Celeste-SteamRIP.com.rar` into `/NVME/Teste*/Celeste-SteamRIP.com` actually writes to `/NVME/TesteCeleste-SteamRIP.com`:

```
$ 7zz x "/NVME/Teste*/Celeste-SteamRIP.com.rar" -y "-o/NVME/Teste*/Celeste-SteamRIP.com"
$ find /NVME
/NVME/Teste*/Celeste-SteamRIP.com.rar
/NVME/TesteCeleste-SteamRIP.com/...
```

7-Zip exits 0, so the progress bar runs all the way to completion, and only afterwards does `extractFilesInDirectory` fail to find the directory it expected and mark the extraction as failed. That matches the report exactly: progress completes, error appears, download folder untouched. There is no way to escape the `*`, so the fix is to stop putting user paths in the switch. The 7-Zip process now gets its own working directory and `-o.`.

Doing that meant spawning it ourselves. The `$spawnOptions` key the wrapper was already passing does nothing: node-7z reads `stream._spawnOptions`, which it never assigns from the options object, so the `cwd` in `extractFilesInDirectory` was silently ignored and only worked because the same path was also being passed as `-o`. Switched to node-7z's documented `$defer` plus `listen` so the child can be spawned with a real `cwd`.

**The archive argument.** 7-Zip also expands wildcards in the archive path itself, on every platform, not only where a shell would glob:

```
$ 7zz x "/downloads/Some Folder/*.7z" -o/out -y    # extracts, so 7z is globbing internally
```

So `Some Game *Repack` stops being a path and becomes a pattern. A sibling folder that happens to match gets pulled into the same run, and if its archive is half-downloaded or unrelated the whole extraction fails. Passing `-spd` makes 7-Zip treat the argument literally. This one applies to `listFiles` too. It does not help with the `-o` problem above, `-spd` only covers the archive and file name arguments.

Both changes live in `src/main/services/7zip.ts`, and every 7z call in the app goes through `SevenZip`, so the emulator installer, the Decky plugin and the nested-archive pass are covered as well.

Checked that multi-volume archives still chain under `-spd`, extracting a `.001` still picks up `.002` and `.003`, so the `part1.rar` handling in `extractFilesInDirectory` is unaffected. Verified the tester's exact layout end to end: a `.rar` inside a folder named `Teste*` now extracts into `Teste*/Celeste-SteamRIP.com`, and a nested archive inside that folder extracts too.

Worth noting the issue is labelled Frontend but the change is main process only.


Hachi-R (approved):
Reviewed at `10dfdbf`. No blocking findings. The shared 7-Zip wrapper now treats archive paths literally for both extraction and listing, and the local node-7z declaration matches the supported `noWildcards` option. Current CI checks are green.

PR #2618, opened 2026-07-31
Halt downloads when the disk runs out of space

Moyasee:
Closes LBX-993 (https://linear.app/los-broxas/issue/LBX-993/halt-downloads-when-users-storage-is-full)

A download that fills up its volume never fails. libtorrent keeps the handle alive, the transfer drops to a few KB/s and just sits there, and nothing in the UI says why. The report behind this was someone 62 GB into a 67 GB download watching the speed fall to 100 KB/s with hours of ETA left, who only worked out what happened by checking their free space themselves.

New `disk-space.ts` reads the free space at the download path with `check-disk-space`, which is already a dependency and is what the download settings modal uses to show space left on disk. A download needs `max(fileSize - bytesDownloaded, 512 MiB)` free: enough for what is left to fetch, and a floor so a nearly dry volume stops things even when the remaining bytes are small.

`watchDownloads` runs the check every 10 seconds against the active download. When it fails, the download is paused through the normal pause path (so both the torrent and the JS HTTP downloader stop cleanly), persisted as `error`, and moved into the paused order. Partial data stays on disk and resuming works as usual once space is freed, since `resumeDownload` already accepts downloads in the error state.

The queue is deliberately not advanced. Starting the next item would write to the same full volume, so `processNextQueuedDownload` runs the same check and holds the queue instead of activating anything.

For feedback the renderer gets a new `on-download-halted` event and shows an error toast, and there is an OS notification gated on `downloadNotificationsEnabled` like the download complete one. The notification matters here because this happens hours into a download that people are not watching.

One behavior worth calling out: since the check compares against the remaining bytes and not just a floor, starting a 67 GB download with 40 GB free now stops early rather than filling the disk first and stalling at 90 percent. That seemed like the point of the issue, but it is easy to relax to floor only if you would rather it not refuse up front.

Tested against a 600 MB volume filled to 37 MB free, which reports 37 MB free against 5 GB needed and refuses to continue. The halt itself has not been exercised against a live download on a full disk.

---


Moyasee:
@greptileai review

Moyasee:
@greptile review

Hachi-R (approved):
Reviewed at `b372973`. No blocking findings. The selected-payload sizing and held-queue retry edge cases are resolved, and the active-download halt, persistence, notification, and resume flow is coherent. Current CI checks are green.

Moyasee:
@greptileai review

Moyasee:
@greptile review

PR #2620, opened 2026-07-31
fix: rank duplicate game executables by structure when auto-detecting

Moyasee:
Closes LBX-857 (https://linear.app/los-broxas/issue/LBX-857).

Automatic executable detection matched on the executable's file name alone and returned the first hit from a recursive `readdir`, so enumeration order decided which copy got bound. Some download options ship extra folders that duplicate part of the install tree, executable included, and whichever the OS happened to enumerate first won. On the layout from the ticket that meant Play pointed at a duplicate rather than the real install.

The ticket suggested fixing this with a list of folder names to skip. I built that first, then went looking for something that did not depend on knowing the names in advance, because a list only ever covers folders somebody already ran into, and everything else stays a coin flip on enumeration order. Three structural signals turned out to be enough.

**Known relative suffix.** The executables data already carries a short relative path per entry, not just the file name, and we were discarding everything but the file name. Candidates matching that suffix now outrank ones matching the name alone. It is a preference rather than a filter, since about 15% of entries record a bare file name and a few name a directory the download root already is, so filtering on it can eliminate every candidate.

**Nested copies lose.** A duplicate is the real path with an extra folder wrapped around it. So when one candidate's relative path is a segment aligned suffix of another's, the longer one is a copy. That single rule settles the reported layout for any wrapper folder, whether or not anybody has seen it before.

**Unshared files break the rest.** Where neither of the above applies, the folder holding more files that appear nowhere else in the tree wins. A folder meant to be copied over an install is mostly duplicates of files that already exist elsewhere; the real one carries its own.

If two candidates still tie on every signal, nothing gets bound. That is deliberate. Nothing in the download flow clears `executablePath` once it is set, so a wrong automatic bind is sticky, keeps pointing Play at the wrong file, and re-downloading will not correct it. Leaving it unset lets the user pick once and have it stick.

Ranking is pure and lives apart from the filesystem walk, so it can be exercised directly without an Electron context.

Also folds in the duplicate of this search that lived in `scanInstalledGames`, so the library scan gets the same behavior, and restores the error log that the first version of this branch dropped.

While reading the surrounding code I found a second, unrelated reason detection could silently skip a game, and fixed it here since it is two lines in the same area. The process map is keyed by lowercased process names, but the executable entries were compared at whatever casing the remote data uses, so any entry with an uppercase letter never matched. Either the lookup missed the key outright, or, where the file name happened to be lowercase, the path comparison failed on an uppercase directory segment instead. Lowercasing the entry once at load time makes both comparisons case insensitive, which is what the code already intended given it lowercases the path side. That is 8 entries reachable on Windows and Linux builds, with no change for the other 11078.

Verified against the layout on the ticket plus nested trees, flat duplicates, wrapper names not known ahead of time, a real folder that sits one level deeper than expected, two known executables in one folder, and the tie case. Every layout was run under every ordering of the candidate files, since enumeration order is the root cause and differs between NTFS and ext4. Also run against a case sensitive volume, because ordering and casing on Linux do not match what macOS reports.

Two things I noticed and did not touch: 12 entries in the executables data are not lowercase, so process-watcher's `path.endsWith` comparison silently misses them, and `scanInstalledGames` does a full recursive walk per game rather than one walk for all of them.


src/main/helpers/find-game-executable.ts
```diff
@@ -0,0 +1,67 @@
...
+  let entries: fs.Dirent[];
+
+  try {
+    entries = await fs.promises.readdir(folderPath, {
+      withFileTypes: true,
+      recursive: true,
+    });
+  } catch {
+    return null;
+  }
```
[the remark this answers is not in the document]
Moyasee: Right, that log got lost when I pulled the function out. Restored in 353ecba4e, now inside the shared helper so both call sites get it.

Moyasee:
@greptile review

Hachi-R (approved):
Reviewed at `353ecba`. No blocking findings. The shared executable finder consistently prefers matches outside normalized helper folders while retaining a fallback, supports both Dirent path variants, and preserves scan-error diagnostics across both call sites. Current CI checks are green.

Hachi-R (changes requested):
Requesting changes for one correctness issue:

`src/main/events/library/scan-installed-games.ts:40` passes an entire library root (for example `C:\Games` or Steam's `common` directory) into a ranking heuristic that assumes candidates are copies of one installation. For catalog entries with only a bare executable name, the candidate set can contain unrelated games with the same basename; because there is no suffix signal, `countUnsharedSiblings` can select whichever unrelated folder has more unique files and persist that path. The current executable catalog contains repeated bare basenames, so this is reachable.

Please scope ranking to a single installation directory, or treat candidates spanning different top-level folders as ambiguous unless a known relative suffix identifies the target.

The structural ranking otherwise looks appropriate when the search root is a single downloaded game. Typecheck and the PR's GitHub workflows pass.

Moyasee:
Valid, fixed in fe3622962.

You were right that the premise does not hold at that scope. The ranking treats candidates as copies of one installation, which is true for a download folder and false for a games root.

Measured the exposure against the current catalog before changing anything:

```
win32 basenames shared by 2+ different games: 737
  at least one side has no relative suffix:   129
  no suffix on any side:                      104
games affected:                               217
worst: game.exe, shared by 221 games
```

Worth adding that the suffix rescues this less often than it looks. On six real download folders I checked, the suffix tier matched 0 of 6, because repackers rename the containing folder and the catalog records the Steam name. So `game.exe` collisions largely fall through to the sibling heuristic, which is exactly the path you identified.

Callers now declare the scope. The download path is unchanged. The scan groups candidates by installation, lets a known relative suffix select one where it can, and treats candidates spanning more than one installation as ambiguous, binding nothing. Once a single installation is selected, the existing rules run inside it, so duplicate folders within one install are still resolved.

Re-verified the six real layouts still bind under a shared scan root, and that the reported layout is untouched on the download path.

src/main/helpers/game-executable-ranking.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  const suffixes = executables
+    .map((executable) => executable.name)
+    .filter((name) => toSegments(name).length > 1);
+
+  const preferred = candidates.filter((candidate) =>
+    suffixes.some((suffix) => endsWithSegments(candidate, suffix))
+  );
+
+  const pool = preferred.length > 0 ? preferred : candidates;
```
Hachi-R: `preferred` is applied before installation ambiguity is checked, but a matching suffix is not unique enough to identify a game. The live catalog currently assigns identical multi-segment paths to unrelated game IDs; for example, `p5s/game.exe` appears for Steam IDs `1382330`, `2490910`, `3938830`, and `3987860`. If the target game's repack is installed as `Renamed/game.exe` while another installed game retains `p5s/game.exe`, this reduces the pool to the unrelated installation and binds it instead of returning `null`.

Please determine the installation set from all basename candidates before applying suffix ranking, and return `null` whenever those candidates span installations (or otherwise prove the suffix is globally unique). A regression test with `Renamed/game.exe` plus `p5s/game.exe` would cover the unsafe case.

Hachi-R (changes requested):
The update correctly introduces installation boundaries, but suffix preference is still allowed to select one installation before cross-installation ambiguity is checked. Because catalog suffixes are not globally unique, the wrong executable can still be persisted in a reachable renamed-install layout. Please address the inline finding before merging.

src/main/helpers/game-executable-ranking.ts
Moyasee: Confirmed and fixed in ec8b880c8. The suffix is not a game identifier, which I had implicitly assumed.

Measured it against the live catalog:

```
multi-segment suffixes shared by 2+ different games: 385
games involved:                                      1065
p5s/game.exe -> 1382330, 2490910, 3938830, 3987860
worst: nioh3demo/nioh3.exe across 134 games
```

The installation set is now derived from every basename candidate, before anything else, and the scan binds nothing whenever those span more than one installation. The suffix is only consulted inside the selected installation, where it is a tiebreak between copies of the same install rather than a claim about which game a folder holds.

Added the regression you described, `Renamed/game.exe` alongside `p5s/game.exe` with the shared catalog entry, asserting null. Also added the weaker case where the suffix matches exactly one installation, which now abstains too, since matching one is not evidence it is the right one.

This makes the scan noticeably more conservative. With 737 colliding basenames, anyone holding two of those games under one scan root now gets no automatic bind for either, and has to set the path by hand. Wrong and sticky is worse than absent, since nothing in the download flow clears executablePath once set, so I think that trade is right.

Hachi-R (approved):
Re-reviewed at `ec8b880c8`. No blocking findings.

The library-scan path now derives installation ambiguity from every basename candidate before suffix ranking, so both the original bare-name collision and the shared-suffix/renamed-install case correctly abstain instead of persisting a potentially unrelated executable. The download-folder behavior remains scoped to one installation.

Build, lint, and text-integrity checks are green. I could not find the regression test mentioned in the thread among the PR's changed files; adding unit coverage for the pure ranking helper is still recommended, but I do not consider that blocking for this fix.

PR #2628, opened 2026-08-02
Let the library scan add the games it finds

Moyasee:
Closes LBX-916 (https://linear.app/los-broxas/issue/LBX-916/improve-the-scan-feature-on-the-library)

The scan only ever linked executables to games that were already in the library. If your library is empty it finds nothing, which is exactly the situation for anyone coming over from another launcher, so the feature reads as broken to the people most likely to reach for it.

It now also adds the games it finds. There is a checkbox in the scan modal, checked by default, and the completion notification says how many games were added and how many already in the library got an executable linked. The modal splits the two in its results list as well.

The bigger change is how matching works. `game-executables.json` stores each executable as a relative path, and `getExecutablesForGame` was throwing that away and keeping only the file name. Basename matching is fine when you already know which game you are looking for, but it falls apart the moment you go the other direction: `game.exe` appears under 221 different games, so a single find would have pulled all 221 into the library. Matching now uses the full relative path, and when several games still tie on the same path, the most specific one wins and an outright tie is skipped rather than guessed. Against the current catalog that resolves 9745 of 10240 executable paths uniquely; the rest, mostly data artifacts like `nioh3demo/nioh3.exe` sitting under 134 app ids, get skipped.

Ambiguity is resolved across the whole catalog before library games are filtered out, and that ordering matters. Doing it the other way round means a game you already own makes a shared executable look unique: with Counter-Strike Source installed and in the library, the scan would add Garry's Mod, because dropping 240 from the candidate set leaves gmod as the only remaining owner of a bare `hl2.exe`. Caught that while testing the first version of the resolver.

The path matching also fixes a smaller existing bug. `counter-strike source/hl2.exe` and `day of defeat source/hl2.exe` used to be indistinguishable, so whichever folder the walk hit first won for both games.

Scanning itself was rewritten while I was in there. It used to run a full recursive `readdir` per library game per directory, so fifty games missing an executable across five directories meant 250 walks of the same trees. It now walks each directory once, keeps only the files whose name appears in the catalog, and matches everything against that index. Directories are resolved through `realpath` so the same tree is not walked twice, symlinked directories are skipped so there are no cycles, and a permission error inside one subdirectory no longer aborts the rest of that tree.

Titles and artwork for added games come from `getGameAssets`, four at a time, and each game goes through `createGame` the same way `addGameToLibrary` does.

Tested the resolver against the real catalog with a synthetic install tree covering the ambiguous cases above. The scan cannot be exercised end to end on macOS, since the catalog filters to nothing there and the default directories are Windows paths.

---


src/main/events/library/scan-installed-games.ts
```diff
@@ -19,45 +25,265 @@ const SCAN_DIRECTORIES = [
...
+  executablePath: string
+): Promise<FoundGame | null> => {
+  const assets = await getGameAssets(objectId, DISCOVERED_GAMES_SHOP).catch(
+    () => null
+  );
+
+  if (!assets?.title) return null;
+
+  const gameKey = levelKeys.game(DISCOVERED_GAMES_SHOP, objectId);
+  const existingGame = await gamesSublevel.get(gameKey);
```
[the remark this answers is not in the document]
Moyasee: `get()` does not reject on a missing key here. `abstract-level` 2.x resolves `undefined`, `NotFoundError` was the old `levelup` behavior. Against this repo's own `node_modules`:

```
existing key: { title: 'Half-Life' }
missing key resolved to: undefined
Promise.all over missing keys: [{"title":"Half-Life"},null,null]
```

The rest of the codebase already depends on this, `add-game-to-library.ts:43` and `import-launchbox-roms.ts:209` both `get()` then branch on truthiness with no catch. If it threw, adding any game to the library would already be broken.

The failure mode you describe is real for other reasons though, a `put` failure or a LevelDB I/O error would take down the whole `Promise.all` and lose the rest of the batch. Fixed in 4f0514bd5 by catching per game instead of at the key read. Same commit attaches a catch to `firstSyncWithRemoteIfNeeded`, which is unawaited and rethrows when `mergeAchievements` fails.

Moyasee:
@greptile review

Moyasee:
@greptile review

src/main/events/library/scan-installed-games.ts
```diff
@@ -1,68 +1,746 @@
...
+        ).filter((choice) => choice !== null);
+
+        if (choices.length === 0) {
+          logger.info(
+            `[ScanInstalledGames] Skipping ${executablePath}, none of its ${objectIds.length} candidates are in the catalogue`
+          );
+          return;
+        }
+
+        if (choices.length === 1) {
```
Hachi-R, line 677: `resolveChoice` converts both a real "no catalogue entry" response and any `getGameAssets` rejection into `null`. After filtering those values out, this branch treats one surviving response as a unique candidate. For a path claimed by several object IDs, a transient failure or rate limit affecting all but one request therefore auto-adds the surviving?potentially wrong?game. Please preserve lookup failure separately and only auto-resolve when every candidate lookup completed successfully; otherwise keep the path ambiguous or skip it.

src/main/events/library/scan-installed-games.ts
```diff
@@ -1,68 +1,746 @@
...
+  );
+
+  const resolveChoice = createChoiceResolver();
+  const ambiguousMatches: AmbiguousMatch[] = [];
+
+  for (const entries of chunk(undecided, DISCOVERED_GAMES_CHUNK_SIZE)) {
+    await Promise.all(
+      entries.map(async ({ executablePath, objectIds }) => {
+        const choices = (
+          await Promise.all(objectIds.map(resolveChoice))
```
Hachi-R, line 649: The chunk size limits executable records, not asset requests: each record still launches every `objectIds.map(resolveChoice)` concurrently. The catalog has paths with more than 100 candidates, so one batch can start hundreds of `getGameAssets` calls despite the intended four-at-a-time limit. Please apply a shared concurrency limiter to the individual metadata lookups; `nameByFolder` has the same pattern.

src/main/events/library/scan-installed-games.ts
```diff
@@ -1,68 +1,746 @@
...
+  if (addedCount > 0) return "scan_games_complete_added_description";
+
+  return "scan_games_complete_linked_description";
+};
+
+async function publishScanNotification(
+  addedCount: number,
+  linkedCount: number
+): Promise<void> {
+  const hasResults = addedCount + linkedCount > 0;
```
Hachi-R: `hasResults` ignores `ambiguousMatches`. When a scan finds only executables that need user selection, the notification says no installed games were found even though reopening the modal presents actionable matches. Pass the ambiguous count into this function and use a result/pending-choice message instead of the no-results message.

Hachi-R (changes requested):
Requesting changes for three issues in the discovery flow. Most importantly, partial metadata failures can turn an ambiguous executable into an apparently unique match and automatically add the wrong game. The current batching also does not bound individual metadata requests, and ambiguous-only scans produce a false "nothing found" notification. Details and suggested fixes are inline.

Hachi-R (changes requested):
Re-reviewed the current head d0cb6131 and the existing discussion. The concurrency limiter and pending-only notification issues from my previous review are fixed, but I am still requesting changes for two matcher correctness gaps.

1. [P1] Resolve ownership before removing library candidates (src/main/events/library/scan-installed-games.ts:390). partitionMatches filters existing-library object IDs out of an unresolved candidate set. If one external candidate remains, the later metadata stage treats it as unique and automatically adds it, even though the executable may belong to an existing game with a stale or different configured path. This is the false-uniqueness ordering problem described in the PR itself. Keep all catalog candidates through ownership resolution, then exclude or suppress library entries after the owner is known.

2. [P2] Do not present partial choices after a candidate lookup fails (src/main/events/library/scan-installed-games.ts:648). The failed guard only skips when fewer than two lookups succeeded. If two candidates succeed and another fails, the code creates an AmbiguousMatch containing only the successful candidates; the UI can then bind the executable incorrectly because the actual owner may be the missing candidate. nameByFolder has the same gap. If any candidate lookup fails, retry or skip that path instead of presenting an incomplete choice set.

All current CI checks are green.

Hachi-R (approved):
Re-reviewed the latest changes. The full candidate set is now preserved until ambiguity is established, and failed asset lookups stop matching instead of producing incomplete choices. This addresses my previous blocking comments, and the current CI checks are green. Approved.

PR #2634, opened 2026-08-02
fix: show custom/non-Steam cover art in game launcher

cybx:
## What

The game launcher window never showed cover art for custom (and other non-Steam) games, even when a cover was set. It always fell back to the placeholder icon.

## Why

`getGameAssets` returns `null` for `shop === "custom"`, and a custom game's artwork lives on the `Game` object (`customCoverImageUrl`), not on `ShopAssets`. But the launcher only read `gameAssets?.coverImageUrl` plus a Steam-only URL fallback, so it had nothing to show for custom games.

## Fix

Resolve the launcher cover from the assets the launcher already loads, in this order:

`customCoverImageUrl → shop cover → library image → custom icon → icon → guessed Steam capsule (last)`

It picks the first present candidate — not an on-error retry ladder like the library card. No new IPC or data fetching. The `local:` protocol is registered app-wide, so local custom covers render in the launcher window too.

Also from review:
- Treat a cover that fails to decode as ready (`imageError`), so a broken local cover can't leave the launcher window hidden and never-closing.
- Quote and `encodeURI` the backdrop `url()` so cover paths containing a space aren't dropped by Chromium.

## Testing

Set a custom cover on a manually-added game (Personalization → Cover), then launch it. Before: placeholder icon. After: the cover shows.


src/renderer/src/pages/game-launcher/game-launcher.tsx
```diff
@@ -129,13 +129,23 @@ export default function GameLauncher() {
...
   const fallbackSteamCoverImage =
-    !normalizedCoverImage && shop === "steam" && objectId
+    shop === "steam" && objectId
       ? `https://shared.steamstatic.com/store_item_assets/steam/apps/${objectId}/library_600x900_2x.jpg`
       : "";
-  const coverImageSource = normalizedCoverImage || fallbackSteamCoverImage;
+
+  // Mirror the library card cover precedence so custom/non-Steam games
+  // (whose assets live on the Game object, not ShopAssets) still show artwork.
+  const coverImageSource =
```
Moyasee, line 144: This routes every custom game with a cover into the non-HTTP branch at line 211, which sets `coverImage` but never clears it on failure. `isReady` at line 267 requires `imageLoaded` whenever `coverImage` is truthy, and `onError` only sets `imageError`. So a cover that fails to decode leaves `isReady` false forever, `showGameLauncherWindow()` at line 274 never fires, and because the auto-close timer is gated on `windowShown` the hidden window never closes either. The game still launches, you just get no splash and a stray invisible window until the next launch.

The HTTP branch already handles this: `getImageDataUrl` returning null clears `coverImage` and sets `colorExtracted`. The local branch needs the same escape hatch. It is reachable because update-custom-game.ts:105 unlinks the previous asset on every save.

Standalone Electron 40.9.3 window with the same `protocol.handle("local")` and CSP:

```
present file : onload,  naturalWidth 4
corrupt file : onerror, no onload
missing file : onerror, no onload
```

`const isReady = imageResolved && (coverImage && !imageError ? imageLoaded : true) && colorExtracted;` covers it.

src/renderer/src/pages/game-launcher/game-launcher.tsx
```diff
@@ -129,13 +129,23 @@ export default function GameLauncher() {
...
+    normalizeImageUrl(game?.customCoverImageUrl) ||
+    normalizeImageUrl(gameAssets?.coverImageUrl) ||
+    fallbackSteamCoverImage ||
+    normalizeImageUrl(gameAssets?.libraryImageUrl) ||
+    normalizeImageUrl(game?.customIconUrl) ||
+    normalizeImageUrl(game?.iconUrl ?? gameAssets?.iconUrl);
```
Moyasee: Line 296 interpolates into `url(${coverImage})` unquoted. `ASSETS_PATH` is `userData/Assets`, so on Windows a custom cover lands in `C:\Users\<name>\AppData\Roaming\hydra\Assets\custom-games\`. Any account with a space in the username produces a path with a space, Chromium rejects the whole declaration, and the blurred backdrop silently disappears. The `` is unaffected since the browser percent-encodes for you.

Same harness, path `.../John Doe/Assets/custom-games/grid-space.png`:

```
<img src>                                        -> loads
style.backgroundImage = url(<path>)              -> computed 'none'
style.backgroundImage = url("<encodeURI(path)>") -> accepted
```

Pre-existing line, but this PR is what starts feeding local paths through it. `url("${encodeURI(coverImage)}")` works, and the main handler already decodeURIs.

src/renderer/src/pages/game-launcher/game-launcher.tsx
```diff
@@ -129,13 +129,23 @@ export default function GameLauncher() {
...
   };
 
-  const normalizedCoverImage =
-    gameAssets?.coverImageUrl?.replaceAll("\\", "/").trim() || "";
+  const normalizeImageUrl = (url?: string | null) =>
+    url?.replaceAll("\\", "/").trim() || "";
+
   const fallbackSteamCoverImage =
-    !normalizedCoverImage && shop === "steam" && objectId
+    shop === "steam" && objectId
```
Moyasee, line 136: Dropping the `!normalizedCoverImage` half of the guard makes this unconditional for steam, so it is always a non-empty string and the three levels below it can never run. The capsule URL is a guess from the appid and 404s when there is no capsule:

```
appid 220       -> 200
appid 730       -> 200
appid 2050650   -> 200
appid 999999999 -> 404
appid 1234      -> 404
```

On a 404 `getImageDataUrl` returns null and you land on the placeholder, so `libraryImageUrl` and `iconUrl` are dead weight for the one shop that would benefit from them. Either restore the guard so the fallback sits last, or make it a real ladder like the library card's `fallbackIndex`.

Separately, the description says this mirrors the library card precedence. The library card retries the next candidate on `onError`; this picks the first present one and stops. Worth rewording.

Moyasee (changes requested):
Premise checks out. `getGameAssets` short-circuits on `shop === "custom"` (get-game-assets.ts:29) while `getGameByObjectId` returns the full record, so the cover was already sitting in the component and just wasn't being read.

Confirmed the rendering side works: `local` is privileged (index.ts:57), handled on the default session (index.ts:82), and renderer/index.html allows it in the CSP. Typecheck passes.

One blocker below, two smaller things.

cybx:
Thanks, all three fixed in `7180004c`.

`isReady` now bails when the cover errors, so a broken local cover can't hang the window open anymore. Backdrop `url()` is quoted + `encodeURI`'d for paths with spaces. And I moved the Steam capsule guess to the end so it stops shadowing the library/icon fallbacks — reworded the description too, you're right it's first-present-wins, not a retry ladder like the card.

Moyasee (approved):
All three addressed, verified against 7180004c.

Checked what the `isReady` fix rests on: color.js rejects on image error (color.esm.js:46), so `colorExtracted` still flips and a broken local cover now falls through to the placeholder and auto-closes. `encodeURI` leaves data URLs untouched so the HTTP path is unaffected, and the space case roundtrips through the handler's `decodeURI`.

The reorder turns out to do more than I asked for. Hydra's API returns `coverImageUrl: null` for Steam games and puts the 600x900 capsule in `libraryImageUrl`, sometimes behind a content hash the guessed URL cannot reproduce:

```
appid 220: coverImageUrl null, libraryImageUrl .../apps/220/ac2f074d79.../library_600x900_2x.jpg
appid 730: coverImageUrl null, libraryImageUrl .../apps/730/library_600x900_2x.jpg
```

So Steam games were taking the guess instead of the real asset URL. They now take `libraryImageUrl`. Same portrait dimensions, no shape change in the cover slot.

Typecheck passes.

PR #2636, opened 2026-08-03
fix: guard HowLongToBeat section against non-array data

cybx:
## What

Opening a game's details page could crash the whole page with:

```
TypeError: howLongToBeatData.map is not a function
    at HowLongToBeatSection
```

## Why

`HowLongToBeatSection` receives `howLongToBeatData` typed as `HowLongToBeatCategory[] | null` and renders it with `howLongToBeatData ? howLongToBeatData.map(...) : <skeleton>`. For some games/shops the API response resolves to a non-array object instead of an array or `null`. A non-array object is truthy, so it passes the check and then `.map` throws, taking down the entire game details view.

## Fix

Coerce the prop to an array (`Array.isArray`) before rendering. Unexpected shapes now hide the section (same as the empty case) instead of crashing.

No behavior change for valid data.


Moyasee:
Guard works for the desktop sidebar, but the same response still crashes Big Picture, which this PR doesn't touch.

`use-game-details.hook.ts:112` stores the body raw:

```ts
globalThis.window.electron.hydraApi
  .get<HowLongToBeatCategory[] | null>(
    `/games/${shop}/${objectId}/how-long-to-beat`,
    { needsAuth: false }
  )
  .then(setHowLongToBeat)
```

`game.tsx:1335` gates on `(howLongToBeat?.length ?? 0) > 0`, which an object body fails harmlessly, but a string body passes. Anything returning HTML with a 200 (Cloudflare interstitial, ISP block page) lands in `how-long-to-beat/index.tsx:33` and throws the same `map is not a function`.

Cheapest fix is validating where the value enters state instead of at each render site, in both places:

```ts
.then((data) => setHowLongToBeat(Array.isArray(data) ? data : null))
```

Same in `sidebar.tsx:155`. That also makes the `HowLongToBeatCategory[] | null` state type true rather than an unchecked cast, so the next consumer doesn't inherit the trap.

Moyasee:
Also, it would help us understand why this error is actually happening is by knowing how to reproduce it, could you share some details on how you encountered such error?

cybx:
Repro: it was a custom game (Minecraft, added manually). The desktop sidebar hits `/games/${shop}/${objectId}/how-long-to-beat` for every shop including `custom`, and that response isn't an array, so it hit `.map`. Big Picture skips custom so it didn't blow up there, but like you said any non-array 200 would.

Moved the check to state entry in both `sidebar.tsx` and `use-game-details.hook.ts`, dropped the render-site guard. `ab467311`.

Moyasee:
> Repro: it was a custom game (Minecraft, added manually). The desktop sidebar hits `/games/${shop}/${objectId}/how-long-to-beat` for every shop including `custom`, and that response isn't an array, so it hit `.map`. Big Picture skips custom so it didn't blow up there, but like you said any non-array 200 would.
> 
> Moved the check to state entry in both `sidebar.tsx` and `use-game-details.hook.ts`, dropped the render-site guard. `ab467311`.

This doesn't seem right, rendering of hltb box is disabled on custom games, so the request is not being made and i'm seeing that from the devtools that no request is being made to the hltb endpoint

cybx:
You're right, my mistake. The sidebar is gated behind `shop !== "custom"` in `game-details-content.tsx:450`, and `HowLongToBeatSection` only renders from inside it, so a custom game never mounts it at all. Retracting that.

The trace in the description was a real crash I hit, not something I reconstructed from the code, but I don't have the report anymore. Where I went wrong was the attribution. The fetch in `sidebar.tsx` builds the URL from `${shop}` with no filter, so I read it as running for everything and never checked whether the sidebar mounts for custom in the first place. It doesn't. So custom is out, and I can't tell you which shop it actually was.

On the unchecked cast you mentioned, it goes all the way down. Preload hands the call to `hydra-api-call.ts`, which hands it to `HydraApi.get`, which is just `axios.get().then(r => r.data)`. Nothing validates it anywhere along the way, and `handleUnauthorizedError` re-throws, so errors can't come back as values either. The type is a claim about the response, not a guarantee, and without the guard a 200 with a non-array body reaches `.map`.

Hachi-R (approved):
The fix makes sense. Validating the response at the state boundary prevents both renderers from storing an impossible value, while `null` preserves the existing empty-state behavior. No issues found.

PR #2637, opened 2026-08-03
fix: drop unfinished download when transferring a game

cybx:
## What

After transferring a game to another folder, an already-playable game could show a resumable "paused" download, making a fully-installed game look downloadable again.

## Why

`transferGameFiles` re-points the game's download record to the new install folder unconditionally. If the game still had an **unfinished** download left over (e.g. a paused partial from an earlier attempt), that partial record got re-pointed at the freshly-installed folder and resurfaced in the UI as a resumable download for a game the user can already play.

## Fix

In `updateDatabaseAfterTransfer`, only re-point downloads that are actually finished (`complete` / `seeding` / `progress >= 1`) — those should keep seeding from the new location. For unfinished downloads, cancel via `DownloadOrchestrator.cancelDownloadById`, which marks the record `removed` and stops the handle **without deleting any files** (the game is fully present on disk after the transfer).

## Notes

- No file deletion; only the stale download record is dropped.
- Completed/seeding transfers behave exactly as before.


cybx:
false positive.

PR #2638, opened 2026-08-03
Update Polish translation

juliazero:


src/locales/pl/translation.json
```diff
@@ -499,125 +705,63 @@
...
+    "eta": "Zakończenie: {{eta}}",
+    "paused": "Wstrzymano",
+    "verifying": "Weryfikowanie…",
+    "completed": "Ukończono",
     "removed": "Nie pobrano",
-    "cancel_download": "Anulować pobieranie?",
-    "cancel_download_description": "Czy na pewno chcesz anulować to pobieranie? Wszystkie pobrane pliki zostaną usunięte.",
+    "cancel": "Anuluj",
+    "cancel_download": "Anuluj pobieranie",
+    "cancel_download_description": "Pobieranie {{title}} zostanie anulowane. Pobrane pliki mogą pozostać na dysku.",
```
Moyasee: This adds a `{{title}}` placeholder that the call sites don't pass. `download-group.tsx:1028` and `:1061` both call it as a bare `t("cancel_download_description")`, so with i18next 23.16.8 this renders literally:

```
Pobieranie {{title}} zostanie anulowane. Pobrane pliki mogą pozostać na dysku.
```

The meaning also flipped: EN says the downloaded files will be deleted, this says they may stay on disk. Suggest keeping the old wording without a placeholder:

```json
"cancel_download_description": "Czy na pewno chcesz anulować to pobieranie? Wszystkie pobrane pliki zostaną usunięte.",
```

src/locales/pl/translation.json
```diff
@@ -826,110 +1060,361 @@
...
+    "platform_classics": "Klasyki",
+    "sort_by": "Sortuj według",
     "achievements_earned": "Zdobyte osiągnięcia",
     "played_recently": "Ostatnio grane",
     "playtime": "Czas gry",
-    "total_play_time": "Łączny czas gry",
-    "manual_playtime_tooltip": "Ten czas gry został ręcznie zaktualizowany",
-    "no_recent_activity_title": "Hmm… nic tutaj",
-    "no_recent_activity_description": "Ostatnio nie grano w żadną grę. Czas to zmienić!",
+    "total_play_time": "Łączny czas gry: {{amount}}",
```
Moyasee: Same problem, `{{amount}}` is never passed. `user-stats-box.tsx:103` renders this as the `<h3>` label and prints the formatted play time in a separate `<p>` below it, so the output is:

```
Łączny czas gry: {{amount}}
```

Drop the suffix and keep it as a label:

```json
"total_play_time": "Łączny czas gry",
```

Moyasee (commented):
Two keys gained placeholders that no call site interpolates, so Polish users see the raw braces. Details inline.

Rest of the file checks out: key parity with EN on main is exact, the new `_few`/`_many` forms resolve correctly at runtime (0 -> "Znaleziono 0 gier", 1 -> "1 grę", 2 -> "2 gry", 5 -> "5 gier", 22 -> "22 gry"), every other placeholder and `<0>` tag matches EN, no duplicate keys, prettier-clean.

Also needs the `translation` label before merge, release notes are generated from it.

Moyasee (approved):
LGTM! Thanks for a quick response.

PR #2639, opened 2026-08-03
fix: render custom games in big picture instead of hanging on load

cybx:
## What

In Big Picture, opening a **custom game** (one added manually, `shop === "custom"`) hangs on the `Loading...` screen forever and can never be played. Normal catalogue games (steam/epic/etc.) load fine.

This PR unblocks that (commit 1) and cleans up the now-visible empty sidebar shells (commit 2).

## Repro

1. Add a custom game to the library (a game not in the catalogue, e.g. Minecraft, with a local executable).
2. Open Big Picture and navigate to that game.
3. The page stays on `Loading...` indefinitely — no hero, no Launch button.

## Why it hangs

The whole page is gated behind a single early return in `src/big-picture/src/pages/game/game.tsx`:

```tsx
if (isLoading || !shopDetails) {
  return (/* Loading... */);
}
```

For a custom game, `shopDetails` is **deliberately never fetched**: the main handler returns `null` for custom, and `useGameDetails` sets `shopDetailsPromise = Promise.resolve(null)` for `shop === "custom"`, then `setShopDetails(null)`. `isLoading` **does** resolve to `false` — so this isn't a stuck promise; the sole reason for the hang is the `!shopDetails` half of the gate, which is `true` forever for custom. `false || !null` → `true`, permanently.

The Play button already handles custom games — the Hero renders `Launch Game` off `game?.executablePath` (populated for custom via `getGameByObjectId`), independent of `shopDetails`. It just never gets the chance because the page never leaves the loading branch. This is the same situation the **desktop** game-details already handles gracefully (it keeps `shopDetails` null for custom and each section guards on `shop !== "custom"` / `!shopDetails`); Big Picture was missing that treatment.

## Fix (commit 1)

- Make the gate custom-aware: `if (isLoading || (shop !== "custom" && !shopDetails))`. Custom games leave the loading branch once `isLoading` is false; normal games are unchanged.
- Null-safe the consumers that read `shopDetails` after the gate: optional-chain in `Hero` (`assets`, `short_description`) and the screenshot carousel (`movies`/`screenshots`), and widen the props of `RequirementsToPlay` / `SupportedLanguages` / `ControllerSupportBox` to `| null`. Sections with no data collapse via their own existing empty-guards.

No new IPC, no synthesized fake shop details — a custom game renders with its hero + Launch button.

## Cleanup (commit 2)

Once rendering, a custom game still showed two empty sidebar shells: a **Game Stats** box (rating/downloads/players all `0`) and an empty **Metadata** section. Commit 2 hides the Stats box for custom games and gates the Metadata section on actual content (developer/publisher/release date, or a launchbox game) — so it also stops rendering empty for non-custom games missing that data.

The Stats box is the sidebar's focus-navigation entry point, so its entry target is made `undefined` for custom too. The Hero and screenshot carousel already treat an undefined target as a no-op block, so gamepad navigation stays correct with an empty sidebar (verified against the focus resolver: an override targeting an unmounted item null-checks and falls back to tree navigation, never crashing or trapping focus).

## Testing

- Big Picture, custom game: renders, shows **Launch Game**, launches the executable; no empty Stats/Metadata shells; gamepad right-nav from the hero is a no-op (nothing to enter); no console errors.
- Big Picture, normal game (steam): unchanged — hero, description, screenshots, stats, metadata, requirements, languages all render.
- `tsc` green.

### DIFF
--- a/src/big-picture/src/pages/game/game.tsx
+++ b/src/big-picture/src/pages/game/game.tsx
@@ -1246,7 +1246,7 @@ export default function Game() {
     };
   }, [descriptionBlocks]);
 
-  if (isLoading || !shopDetails) {
+  if (isLoading || (shop !== "custom" && !shopDetails)) {
     return (
       <VerticalFocusGroup regionId={GAME_PAGE_REGION_ID} asChild>
         <div className="game-page">
@@ -1305,8 +1305,8 @@ export default function Game() {
           <div className="game-page__main-layout">
             <div className="game-page__main-column">
               <ScreenshotCarousel
-                videos={shopDetails.movies ?? []}
-                screenshots={shopDetails.screenshots ?? []}
+                videos={shopDetails?.movies ?? []}
+                screenshots={shopDetails?.screenshots ?? []}
                 onActiveItemChange={setActiveMediaItemId}
                 nextContentEntryTarget={
                   descriptionEntryTarget ?? commentsEntryTarget
@@ -1487,7 +1487,7 @@ export default function Game() {
                     className="game-page__sidebar-section game-page__metadata"
                     aria-label="Game info"
                   >
-                    {isLaunchboxGame && shopDetails.platform ? (
+                    {isLaunchboxGame && shopDetails?.platform ? (
                       <div className="game-page__metadata-row">
                         <Typography className="game-page__metadata-label">
                           Platform
--- a/src/big-picture/src/pages/game/game.tsx
+++ b/src/big-picture/src/pages/game/game.tsx
@@ -614,10 +614,11 @@ export default function Game() {
     };
   }, [activeMediaItemId, hasMedia]);
   const sidebarStatsEntryTarget = useMemo(
-    () => getItemFocusTarget(GAME_SIDEBAR_STATS_ID),
-    []
+    () =>
+      shop === "custom" ? undefined : getItemFocusTarget(GAME_SIDEBAR_STATS_ID),
+    [shop]
   );
-  const bodyRightNavigationTarget = useMemo<FocusOverrideTarget>(
+  const bodyRightNavigationTarget = useMemo<FocusOverrideTarget | undefined>(
     () => sidebarStatsEntryTarget,
     [sidebarStatsEntryTarget]
   );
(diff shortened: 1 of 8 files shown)


src/shared/supported-languages.ts
```diff
@@ -0,0 +1,22 @@
+export interface SupportedLanguage {
+  language: string;
+  hasAudio: boolean;
+}
+
+/**
+ * Parses a Steam `supported_languages` HTML string into a list of languages,
+ * flagging the ones that ship full audio (marked with an asterisk).
+ */
```
Moyasee: This should come out. We do not keep comments here, docblocks on new exported functions included.

The function name plus the `SupportedLanguage` return type already say what it does, and `hasAudio` is what the asterisk means, so the second sentence is not carrying anything the signature does not. Put the Steam-format detail in the body of the `refactor: extract supported-languages parsing into a shared helper` commit instead, which is where a reviewer would actually go looking for it.

src/big-picture/src/pages/game/game.tsx
```diff
@@ -1373,91 +1376,93 @@ export default function Game() {
...
-                <FocusItem
-                  id={GAME_SIDEBAR_METADATA_ID}
-                  navigationOrder={5}
-                  navigationOverrides={sidebarCarouselNavigationOverrides}
-                  asChild
-                >
-                  <section
-                    className="game-page__sidebar-section game-page__metadata"
-                    aria-label="Game info"
+                {(developer || publisher || releaseDate || isLaunchboxGame) && (
```
Moyasee: `isLaunchboxGame` goes in unconditionally, but every launchbox row inside is separately gated on `shopDetails?.platform`, `launchboxGenres.length` and `launchboxRegions.length`. A launchbox game with none of those and no developer, publisher or release date still renders an empty `<section>` plus a focusable `FocusItem`, which is the shell commit 2 is meant to remove.

Gating on the row data instead covers it:

```tsx
{(developer || publisher || releaseDate ||
  (isLaunchboxGame && (shopDetails?.platform || launchboxGenres.length > 0 || launchboxRegions.length > 0))) && (
```

src/big-picture/src/pages/game/game.tsx
```diff
@@ -1150,7 +1151,7 @@ export default function Game() {
     };
   }, [descriptionBlocks]);
 
-  if (isLoading || !shopDetails) {
+  if (isLoading || (shop !== "custom" && !shopDetails)) {
```
Moyasee: Follow-up rather than something to fix here.

This still hangs a non-custom game whose fetch fails. `getGameShopDetails(...).catch(() => null)` in `use-game-details.hook.ts` swallows the error, `isLoading` goes false, `shopDetails` stays null, and a steam game with the API unreachable sits on Loading forever, same as custom did.

Every consumer downstream is null-safe after this PR, so plain `if (isLoading)` would cover both cases. I walked each one and none crash on a null `shopDetails` with `shop === "steam"`: Hero optional-chains, the carousel defaults to `[]`, `descriptionBlocks` uses `?? ""`, metadata is gated, and requirements, languages and controller support all return null. Separate PR if you would rather keep this one scoped.

Moyasee (commented):
Traced the diagnosis and it holds. `use-game-details.hook.ts` sets `shopDetailsPromise` to `Promise.resolve(null)` for custom and `isLoading` does flip false in the `finally`, so `!shopDetails` really is the only thing pinning the page to Loading.

Checked the parts that looked risky and they are all fine:

- The sidebar is genuinely empty for custom, so making `sidebarStatsEntryTarget` undefined does not strand anything reachable. The hook's `else` branch resets `howLongToBeat`, `protonDBData` and `achievements` when shop is custom, controller support gates on steam, and requirements, languages and metadata all return null.
- Hook order is intact. `RequirementsToPlay`'s early return sits after all four of its hooks, and `ControllerSupportBox` only has `useTranslation` ahead of its return.
- The widened types match the callees. `resolvePreferredGameAssets` already takes `ShopAssets | null | undefined`, and `normalizeRequirementsHtml` gets `raw ?? ""`.
- The three `shopDetails` derefs left non-optional (game.tsx 444, 581, 1396) are each behind a guard.

Three comments inline. One is a doc comment, one is a gap in the commit 2 cleanup, and the last is a follow-up you can leave out of this PR.

src/shared/supported-languages.ts
cybx: Removed, and moved the Steam format note into the `refactor:` commit body.

src/big-picture/src/pages/game/game.tsx
cybx: Good catch, that's fixed, used your version. An empty launchbox game doesn't render the section or the FocusItem anymore.

src/big-picture/src/pages/game/game.tsx
cybx: Yeah, agreed, and agreed it belongs in its own PR. Leaving the gate as is here. I'll open the follow-up for the plain `if (isLoading)` if you want it. Resolving this one so it doesn't block the merge, but unresolve if you'd rather keep it here.

cybx:
Rebased on main since the branch was behind, and amended each fix into the commit it belongs to instead of stacking a fixup, since you were reading this commit by commit. It was a force push, so a couple of your anchors show as outdated.

src/big-picture/src/pages/game/game.tsx
cybx: Changed my mind on this one, it's in as `a618c8ab`. You'd already done the verification and it's a one-line change, so leaving it for a second PR was just more work for you later.

One extra data point for your walkthrough: with the gate down to plain `isLoading`, TypeScript loses the narrowing it used to get from `!shopDetails`, so `tsc` has to prove every deref below it on its own now. It passes, which matches what you found by hand.

cybx:
Pulled the follow-up into this PR after all, as `a618c8ab`. The gate is now plain `if (isLoading)`, so a steam or launchbox game whose metadata fetch fails stops sitting on the loading screen too. Commits 1 to 3 are untouched, so your earlier review still applies to them.

src/big-picture/src/pages/game/game.tsx
```diff
@@ -1150,7 +1151,7 @@ export default function Game() {
     };
   }, [descriptionBlocks]);
 
-  if (isLoading || !shopDetails) {
+  if (isLoading) {
```
Moyasee, line 1154: The library case works and is the point of this commit: an installed game still opens with the API down. The catalogue case is a new empty shell.

A game that is not in the library (`catalogue/card-shell.tsx` and `profile.tsx` both navigate here) with a failed fetch leaves `game` and `shopDetails` both null, and the page renders:

- the title falling back to the literal `"Download Game"` (game.tsx:390)
- Hero catalogue actions, since `shouldShowCatalogActions` is `!game && canAddToLibrary` (hero/index.tsx:113)
- Add to Library wired to `handleAddToLibrary`, which early-returns on `!shopDetails` at game.tsx:569 before it reaches `setIsAddingToLibrary`, so the click is a silent no-op with no spinner

Three ways into that state, all in `use-game-details.hook.ts`: the `.catch(() => null)` on `getGameShopDetails`, plus `getGameStats` and `getGameAssets`, which are not caught at all and reject the whole `Promise.all` while the `finally` still clears `isLoading`.

Loosening the `handleAddToLibrary` guard is not the fix, that would add a library row titled "Download Game". The missing piece is telling "custom, no details expected" apart from "fetch failed". Track the failure in the hook and render an error with a retry when `!game && !shopDetails && shop !== "custom"`, leaving library and custom on this gate.

Moyasee (commented):
All three addressed, and the docblock reasoning moved into the refactor commit body.

Re-checked the rest of the range: the pt-BR churn between the two revisions is just #2643 landing on main, nothing of yours. Prettier is clean on all six touched files against the repo config, and CI is green.

The follow-up commit does introduce one new state. Comment inline.

src/big-picture/src/pages/game/game.tsx
cybx: Confirmed, all of it. `resolvedGameTitle` falls through to the literal at `game.tsx:390`, `shouldShowCatalogActions` is `!game && canAddToLibrary` at `hero/index.tsx:113`, and the `!shopDetails` guard in `handleAddToLibrary` sits above `setIsAddingToLibrary` at `game.tsx:565`, so the click is silent. The three entry points hold too: `getGameStats` and `getGameAssets` have no `.catch` and the block is `try/finally` with no `catch`, so either one rejecting clears `isLoading` with `shopDetails` still null.

So I traded a permanent loading screen for a dead "Download Game" shell. Better for library and custom, worse for catalogue.

Pulling the follow-up in here was my call and it was the wrong one. What you're describing is hook failure state plus an error render with a focusable retry, which is its own review, and the rest of this PR is done.

I'd rather drop that commit from this PR and open the follow-up separately, taking your `!game && !shopDetails && shop !== "custom"` gate and the retry as the design. Want me to do that, or would you rather it stay here?

Moyasee: Drop it and open the follow-up separately, taking the gate and the retry as the design. The first three commits are the fix and they are done.

src/big-picture/src/pages/game/game.tsx
```diff
@@ -1281,56 +1282,58 @@ export default function Game() {
 
             <VerticalFocusGroup regionId={GAME_SIDEBAR_REGION_ID} asChild>
               <div className="game-page__sidebar">
```
Moyasee: Hiding the stats box empties the sidebar for custom games but the wrapper still renders, and `game.scss:115-120` holds it at `width: 500px` in a flex row with a 64px gap. Every child is gone for custom, so on a 1280 handheld:

```
sidebarEmptyChildren: 0   sidebarWidth: 500
mainWithEmptySidebar: 524   mainWithoutSidebar: 1088
```

Skipping the group when it has no content is safe, the only other uses of `GAME_SIDEBAR_REGION_ID` are `isRegionWithinTree` checks at 359 and 1031 and both want false when nothing there can hold focus.

`GameReviews` at 1275 has the same shape: `game-reviews/index.tsx:58` early-returns before `setReviewsLoading(true)` for custom, so the condition is false and custom games get a "Comments / 0 / No comments yet" block that can never fill.

Moyasee (changes requested):
First three commits are sound and they fix the real thing, the `!shopDetails` in the gate. Answered your question on the open thread: drop the fourth. One nit inline.

src/big-picture/src/pages/game/game.tsx
cybx: Dropped, `a618c8ab` is out. I'll open the follow-up separately with the plain `isLoading` gate plus the retry as the design.

src/big-picture/src/pages/game/game.tsx
cybx: Both fixed, amended into the `feat:` commit (now `3cb60123`), which I retitled to "hide empty sections" since it covers the reviews block too:

- The sidebar group is skipped entirely for custom. Custom is the only shop where it can end up empty, since every other shop always renders the Stats box, so the gate is `shop !== "custom"` and the now-redundant inner gate on the Stats box came out. Your two `isRegionWithinTree` call sites get false, as wanted.
- `Divider` + `GameReviews` get the same gate. With `GameReviews` unmounted, `hasNavigableComments` keeps its initial `false` and `commentsEntryTarget` stays `undefined`, which every consumer already treats as absent.

cybx:
Force push again: dropped `a618c8ab` per the thread, amended the empty-sidebar/reviews fix into the `feat:` commit (now `3cb60123`), and rebased on main since the branch was behind. Commits 1 and 3 are content-unchanged, so your review of them still applies.

Hachi-R (changes requested):
The custom-game hang this PR started from is already gone on `main`. `useGameDetails` now synthesizes `buildFallbackShopDetails` when the catalogue payload is null, so `if (isLoading || !shopDetails)` no longer sticks forever for `shop === "custom"`.

This branch is also conflicting. Please rebase onto current `main`.

Keep the empty Stats / Metadata / Reviews cleanup if it still applies after the fallback (custom games currently render those empty shells). The loading-gate commit is redundant with the fallback and can be dropped.

`game-achievements.tsx` still has the same `isLoading || !shopDetails` gate; it is similarly unblocked by the fallback, so it does not need to be in this PR.

The earlier request to drop the fourth commit is already done. Re-request review after the rebase.

PR #2642, opened 2026-08-05
feat(compatibility): add option to disable proton logs and set global environment variables

Cryptorial0:
## What

- Adds a setting to disable Proton debug logs by default (`PROTON_LOG=1` is now opt-in via `protonLogEnabled`).
- Adds per-game toggle overrides for Proton logging in both Desktop and Big Picture compatibility settings.
- Adds global compatibility environment variables setting to inject user-defined environment variables into compatibility game launches.

## Why

- Proton logging by default creates large log files on disk for every game run on Linux. Making it opt-in saves disk space and reduces logging overhead.
- Global environment variables allow users to configure global compatibility tweaks (such as `PROTON_FSR4_UPGRADE=1`, `MANGOHUD=1`, etc.) without manually editing launch scripts.

## Checkbox Guidelines

- [x] I have tested these changes locally.
- [x] I have verified that all linting and typecheck tasks pass (`npm run typecheck`, `npm run lint`).


Moyasee:
@greptile

docs/PR_TEST.md
```diff
@@ -0,0 +1,9 @@
+## Purpose
```
Moyasee: Leftover test artifact, unrelated to Proton logs. The file says so itself: "a tiny test PR created by an automated assistant to exercise the project's Pull Request CI". Drop it.

src/main/constants.ts
```diff
@@ -4,7 +4,8 @@ import { SystemPath } from "./services/system-path";
 
 export const defaultDownloadsPath = SystemPath.getPath("downloads");
 
-export const isStaging = import.meta.env.MAIN_VITE_API_URL.includes("staging");
+const mainViteApiUrl = import.meta.env.MAIN_VITE_API_URL ?? "";
```
Moyasee: Unrelated to this PR, and it swaps a hard failure for a silent one.

`isStaging` feeds the db directory, the log directory, and the app version. With this change a staging build that is missing `MAIN_VITE_API_URL` no longer crashes at startup, it quietly writes to the production db path instead.

Revert it here. If it is a real bug it deserves its own PR.

src/main/helpers/launch-game.ts
```diff
@@ -259,6 +296,15 @@ const launchWindowsBinaryOnLinux = async (
...
   await cleanupStaleCompatibilityProcesses(objectId, winePrefixPath);
 
+  const userPreferences = await db
+    .get<string, UserPreferences | null>(levelKeys.userPreferences, {
+      valueEncoding: "json",
+    })
+    .catch(() => null);
+
+  const protonLogEnabled =
+    game?.protonLogEnabled ?? userPreferences?.protonLogEnabled === true;
```
Moyasee: `??` only falls through on `null` and `undefined`. Once a user turns a game's Proton logging off, `false` is stored and the global setting is ignored for that game permanently.

The UI disagrees. `compatibility-section.tsx` renders `checked={protonLogEnabled || globalProtonLogEnabled}` and disables the box with an "enabled globally" tooltip, so the user sees a checked, locked checkbox while launch writes `PROTON_LOG=0`.

`useMangohud` and `useGamemode` twelve lines up already have the right shape:

```ts
const protonLogEnabled =
  userPreferences?.protonLogEnabled === true || game?.protonLogEnabled === true;
```

src/main/helpers/launch-game.ts
```diff
@@ -259,6 +296,15 @@ const launchWindowsBinaryOnLinux = async (
 
   await cleanupStaleCompatibilityProcesses(objectId, winePrefixPath);
 
+  const userPreferences = await db
```
Moyasee: `launchGame` already reads `userPreferences` and passes `useMangohud` and `useGamemode` down. Pass `protonLogEnabled` the same way instead of hitting the db a second time per launch.

Same area: `parseCompatibilityEnvironmentVariables` runs on every launch including win32, where the result is never used.

src/main/services/umu.ts
```diff
@@ -234,14 +236,15 @@ export class Umu {
...
-      PROTON_LOG: "1",
+      ...(options?.compatibilityEnvironmentVariables ?? {}),
       ...(options?.gameId ? { GAMEID: `umu-${options.gameId}` } : {}),
       ...(options?.winePrefixPath
         ? { WINEPREFIX: options.winePrefixPath }
         : {}),
       ...(options?.protonPath ? { PROTONPATH: options.protonPath } : {}),
       ...(options?.useMangohud ? { MANGOHUD: "1" } : {}),
       ...resolvedLaunchCommand.env,
+      PROTON_LOG: options?.protonLogEnabled ? "1" : "0",
```
Moyasee: Putting `PROTON_LOG` last means it overrides both `compatibilityEnvironmentVariables` and `resolvedLaunchCommand.env`.

Two effects. `PROTON_LOG=+timestamp,+seh` typed into the new global textarea is silently discarded, which is surprising for a box whose whole job is setting env vars. And per-game launch options of the form `PROTON_LOG=+all %command%` used to work, since the old code had `PROTON_LOG` first and `resolvedLaunchCommand.env` last. That part is a regression.

Proton takes channel strings, not just `1` and `0`. Keeping the default first restores the old precedence and makes the variable behave like every other one the textarea accepts:

```ts
const launchEnv = {
  PROTON_LOG: options?.protonLogEnabled ? "1" : "0",
  ...(options?.compatibilityEnvironmentVariables ?? {}),
  ...(options?.gameId ? { GAMEID: `umu-${options.gameId}` } : {}),
  ...
  ...resolvedLaunchCommand.env,
};
```

src/big-picture/src/pages/settings/compatibility.tsx
```diff
@@ -78,13 +87,18 @@ function getProtonSourceDescription(version: ProtonVersion | null) {
...
   const { showSuccessToast } = useBigPictureToast();
   const [form, setForm] = useState<CompatibilityForm>(DEFAULT_FORM);
   const [protonVersions, setProtonVersions] = useState<ProtonVersion[]>([]);
   const [protonVersionsLoaded, setProtonVersionsLoaded] = useState(false);
   const [gamemodeAvailable, setGamemodeAvailable] = useState(false);
   const [mangohudAvailable, setMangohudAvailable] = useState(false);
+  const [
+    compatibilityEnvironmentVariables,
+    setCompatibilityEnvironmentVariables,
+  ] = useState("");
```
Moyasee: This state is never populated from preferences, so the textarea always renders empty.

The sync effect below writes `form.compatibilityEnvironmentVariables`, which nothing ever reads, while the textarea binds to this standalone state instead.

The second half is worse: `onBlur` persists `compatibilityEnvironmentVariables || null`, so focusing and leaving the box that is empty for this reason destroys whatever the user configured in the desktop settings.

Drop this `useState`, bind `value={form.compatibilityEnvironmentVariables}`, and update through `setForm`.

src/big-picture/src/pages/settings/compatibility.tsx
```diff
@@ -387,6 +408,61 @@ export function CompatibilitySettingsSection({
...
+              <div className="compatibility-settings-section__env-vars-group">
+                <label
+                  htmlFor="compatibility-environment-variables"
+                  className="compatibility-settings-section__env-vars-label"
+                >
+                  {t("Compatibility environment variables")}
+                </label>
+                <textarea
+                  id="compatibility-environment-variables"
+                  className="compatibility-settings-section__env-vars-textarea"
```
Moyasee: Raw `textarea` with no `focusId`, nested inside the proton-logging focus item where only the `Checkbox` registers one. Big Picture is controller driven, so as written this setting is visible there but not editable.

`components/common/input` already supports `focusId`. Either wire it in or keep env vars on the desktop settings only.

src/big-picture/src/pages/settings/compatibility.tsx
```diff
@@ -387,6 +408,61 @@ export function CompatibilitySettingsSection({
...
+                  onChange={(event) => {
+                    setCompatibilityEnvironmentVariables(event.target.value);
+                  }}
+                  onBlur={() => {
+                    void updateCompatibilityPreferences({
+                      compatibilityEnvironmentVariables:
+                        compatibilityEnvironmentVariables || null,
+                    });
+                  }}
+                  placeholder="PROTON_FSR4_UPGRADE=1\nMANGOHUD=1\n# One variable per line"
```
Moyasee: JSX string-literal attributes are not escape processed, so this renders on a single line as `PROTON_FSR4_UPGRADE=1\nMANGOHUD=1\n# One variable per line`.

The desktop side avoids it by going through `t()`, where JSON parsing turns `\n` into real newlines. Use a braced template literal here, or the same translation key.

src/renderer/src/pages/settings/settings-behavior.scss
```diff
@@ -107,4 +107,23 @@
...
     gap: calc(globals.$spacing-unit * 1.25);
   }
+
+  &__env-vars {
+    display: flex;
+    flex-direction: column;
+    gap: calc(globals.$spacing-unit / 2);
+  }
+
+  &__textarea {
```
Moyasee: Neither of these two rules matches anything. The JSX uses `settings-context-compatibility__env-vars` and `settings-context-compatibility__textarea`:

```
$ git grep -n "settings-behavior__textarea\|settings-behavior__env-vars" -- '*.tsx'
(no matches)
```

So the desktop textarea renders with browser defaults. Rename one side.

While you are here, `__env-vars-label` and `__env-vars-help` are referenced in both UIs and have no rules anywhere.

src/renderer/src/pages/settings/settings-context-compatibility.tsx
```diff
@@ -125,7 +134,7 @@ export function SettingsContextCompatibility() {
         setDefaultWinePrefixBasePath("");
         setDefaultWinePrefixPath(userPreferences?.defaultWinePrefixPath ?? "");
       });
-  }, []);
+  }, [userPreferences?.defaultWinePrefixPath]);
```
Moyasee: Unrelated to this PR. Adding the dep re-runs `getDefaultWinePrefixSelectionPath()` and overwrites local wine prefix state every time that preference changes. Wine prefixes are not in scope here, so split it out.

src/locales/en/translation.json
```diff
@@ -514,6 +514,9 @@
     "run_with_gamemode_prefix": "Automatically run with",
     "gamemode_not_available_tooltip": "GameMode is not available in your PATH",
     "gamemode_disabled_due_to_global_setting_tooltip": "This option is disabled because GameMode is enabled globally",
+    "proton_logging": "Proton logging",
+    "proton_logging_description": "Write Proton debug logs for this game. Overrides the global setting.",
```
Moyasee: Never referenced, same for `enable_proton_logging_description` at line 1016:

```
$ git grep -n "proton_logging_description" -- src/
src/locales/en/translation.json:518:    "proton_logging_description": ...
src/locales/en/translation.json:1016:    "enable_proton_logging_description": ...
```

The desktop settings checkbox has no `secondaryText`, unlike its Big Picture counterpart. Either use them or drop them.

Separately, the new keys land in `en` only. `ru`, `es` and `pt-BR` ship in the same commit as the English strings in this repo, both here and under `src/big-picture/src/locales`.

src/big-picture/src/components/pages/game/game-settings-modal/compatibility-tab.tsx
```diff
@@ -345,6 +359,26 @@ export function GameCompatibilitySettingsTab({
             handleToggleMangohud(checked).catch(() => {});
           }}
         />
+
+        <Checkbox
+          id="game-compatibility-settings-proton-log"
+          label={t("Proton logging")}
```
Moyasee: `t()` does nothing for these strings. They live in the `exact` bundle, which is applied by the DOM observer in `i18n.tsx` rather than by i18next, so `t("Proton logging")` just returns its own key and the observer translates the rendered text afterwards. That is why the GameMode and MangoHud checkboxes right above pass plain string labels.

Same in `pages/settings/compatibility.tsx`, which additionally declares a `"settings"` namespace that Big Picture never registers.

Moyasee (commented):
Two files do not belong in this PR: `docs/PR_TEST.md` and the `constants.ts` `isStaging` change.

Beyond those, the env var feature does not work end to end in either UI. Big Picture never loads the saved value and wipes it on blur, the desktop textarea picks class names the SCSS does not define, and the Big Picture placeholder renders a literal backslash-n. On the logging side, a stored per-game `false` beats the global setting, which contradicts what the checkbox renders.

`shellQuote` is already applied to the new env values in `umu.ts`, so the textarea does not open a shell injection.

One note on the checklist: it says `npm run typecheck` and `npm run lint` passed, but `package.json` sets `engines.npm` to `"please-use-yarn"`, so neither could have run as written. The class name mismatch, the empty Big Picture textarea, and the placeholder are all things one manual pass over each settings screen would have surfaced.

Rest inline.

src/renderer/src/pages/settings/settings-behavior.scss
```diff
@@ -107,4 +107,35 @@
...
     gap: calc(globals.$spacing-unit * 1.25);
   }
+
+  &__env-vars {
+    display: flex;
+    flex-direction: column;
+    gap: calc(globals.$spacing-unit / 2);
+  }
+
+  &__env-vars-label {
```
Moyasee: Still does not match. Two more rules were added under `.settings-behavior` rather than renaming either side, so there are now four dead selectors instead of two:

```
$ git grep -rn "settings-context-compatibility__env-vars\|settings-context-compatibility__textarea" -- '*.scss'
(no matches)
```

The JSX uses `settings-context-compatibility__env-vars`, `__env-vars-label`, `__env-vars-help` and `__textarea`. The desktop textarea, its label and its help text all still render with browser defaults. Rename the four rules to `.settings-context-compatibility`, or point the JSX at `settings-behavior__*`.

src/locales/ru/translation.json
```diff
@@ -1008,6 +1008,11 @@
     "enable_auto_install": "Загружать обновления автоматически",
     "run_games_with_gamemode": "Запускать игры с GameMode по умолчанию",
     "run_games_with_mangohud": "Запускать игры с MangoHud по умолчанию",
+    "enable_proton_logging": "Включить логирование Proton",
+    "proton_log_disabled_due_to_global_setting_tooltip": "Эта опция отключена, так как логирование Proton включено глобально",
```
Moyasee: Wrong namespace. `en` has this key under `game_details`, the four new locales put it under `settings`:

```
en:              proton_log_disabled_due_to_global_setting_tooltip -> ['game_details']
ru/es/pt-BR/el:  proton_log_disabled_due_to_global_setting_tooltip -> ['settings']
```

`compatibility-section.tsx` reads it through `useTranslation("game_details")`, so none of the four translations are reachable and the tooltip falls back to the English `defaultValue`. Move it to `game_details` in all four.

While you are in here: `proton_logging` was not translated in any of them, and that is the label on the per-game checkbox this PR adds. `fallbackLng: "en"` keeps it from rendering the raw key, but it stays English next to a translated tooltip. Same four files.

src/locales/en/translation.json
Moyasee: Still unreferenced, along with `enable_proton_logging_description` at line 1016. Neither is read anywhere:

```
$ git grep -n "proton_logging_description" -- src/
src/locales/en/translation.json:518
src/locales/en/translation.json:1016
```

`enable_proton_logging_description` also did not get translated into the four new locales, so it is now a dead key that is English-only. Drop both, or wire them into the `secondaryText` the desktop checkboxes are missing.

src/big-picture/src/locales/en/translation.json
```diff
@@ -624,6 +624,13 @@
     "RPCS3 is already running": "RPCS3 is already running",
     "Run with GameMode": "Run with GameMode",
     "Run with MangoHud": "Run with MangoHud",
+    "Enable Proton logging": "Enable Proton logging",
+    "Write Proton debug logs for compatibility launches.": "Write Proton debug logs for compatibility launches.",
+    "Proton logging": "Proton logging",
+    "Write Proton debug logs for this game. Overrides the global setting.": "Write Proton debug logs for this game. Overrides the global setting.",
```
Moyasee, line 630: Stale after this revision. The component now renders `"Write Proton debug logs for this game."`, so this entry no longer matches any rendered text and the string that is rendered has no entry at all.

Separately, none of the seven new `exact` strings were added to the other bundles:

```
ru 0/5   es 0/5   pt-BR 0/5   fr 0/5   el 0/5
```

Since `translateExactText` falls back to the source text, Big Picture shows English for all of this in every locale except `en`.

src/big-picture/src/components/pages/game/game-settings-modal/compatibility-tab.tsx
```diff
@@ -345,6 +359,22 @@ export function GameCompatibilitySettingsTab({
...
           }}
         />
+
+        <Checkbox
+          id="game-compatibility-settings-proton-log"
+          label="Proton logging"
+          secondaryText={
+            globalProtonLogEnabled
+              ? "This option is disabled because Proton logging is enabled globally"
+              : "Write Proton debug logs for this game."
```
Moyasee: Dropping "Overrides the global setting." orphans the `exact` entry in `big-picture/src/locales/en/translation.json` and leaves this new wording with no entry. Update the bundle to match whichever string you keep.

src/big-picture/src/pages/settings/compatibility.tsx
```diff
@@ -387,6 +412,69 @@ export function CompatibilitySettingsSection({
...
+                        compatibilityEnvironmentVariables: nextValue,
+                      }));
+                    }}
+                    onBlur={() => {
+                      void updateCompatibilityPreferences({
+                        compatibilityEnvironmentVariables:
+                          form.compatibilityEnvironmentVariables || null,
+                      });
+                    }}
+                    placeholder={`PROTON_FSR4_UPGRADE=1\nMANGOHUD=1\n# Applies these variables to every compatibility launch.`}
```
Moyasee: The comment line here is now word for word the help paragraph rendered directly beneath it. The desktop placeholder still says `# One variable per line`, which is the part a user cannot infer from looking at the box. Put that back.

Moyasee (commented):
Most of it landed. The test file and the `isStaging` change are gone, the wine prefix dep is reverted, the Big Picture textarea now binds to `form` so it loads and no longer wipes on blur, `??` became `||` and moved up into `launchGame` with the duplicate db read dropped, and `PROTON_LOG` is first in `launchEnv` again. The `FocusItem` wrapper matches what `common/input` already does and `settings.tsx` provides the region, so that one is properly solved rather than patched.

Six left, all in the two areas that did not get a second pass: the desktop stylesheet and i18n. Inline.

src/main/helpers/launch-game.ts
```diff
@@ -249,7 +285,9 @@ const launchWindowsBinaryOnLinux = async (
   game: Game | undefined,
   launchOptions: string | null | undefined,
   useMangohud: boolean,
-  useGamemode: boolean
+  useGamemode: boolean,
+  protonLogEnabled: boolean,
```
Moyasee: `typescript:S107`, async arrow function has too many parameters (9), max 7.

`main` sits at exactly 7 here, so the two params added in this PR are what crossed it. An options object is the cheap fix and reads better at the call site than nine positional arguments, four of which are booleans in a row:

```ts
const launchWindowsBinaryOnLinux = async (options: {
  gameKey: string;
  objectId: string;
  parsedPath: string;
  game: Game | undefined;
  launchOptions: string | null | undefined;
  useMangohud: boolean;
  useGamemode: boolean;
  protonLogEnabled: boolean;
  compatibilityEnvironmentVariables: Record<string, string>;
}): Promise<boolean> => {
```

src/main/helpers/launch-game.ts
```diff
@@ -50,6 +50,40 @@ const ensureExecutablePermission = (executablePath: string) => {
...
 
+const parseCompatibilityEnvironmentVariables = (
+  rawVariables: string | null | undefined
+): Record<string, string> => {
+  if (!rawVariables) {
+    return {};
+  }
+
+  const environmentVariables: Record<string, string> = {};
+  const envVariableNameRegex = /^[A-Za-z_][A-Za-z0-9_]*$/;
```
Moyasee: `typescript:S6353`, use `\w` instead of `[A-Za-z0-9_]`.

Rather than just swapping the character class, note this regex already exists verbatim one file over:

```
$ git grep -n "A-Za-z0-9_" -- src/main
src/main/helpers/launch-game.ts:61:  const envVariableNameRegex = /^[A-Za-z_][A-Za-z0-9_]*$/;
src/main/helpers/resolve-launch-command.ts:5:const envVariableNameRegex = /^[A-Za-z_][A-Za-z0-9_]*$/;
```

Both are validating the same thing, an env var name, and they need to agree or the textarea and `%command%` prefixes will accept different sets of names. Export it from `resolve-launch-command.ts` and import it here. That clears the rule in one place and removes the drift risk.

src/main/services/umu.ts
```diff
@@ -234,7 +236,8 @@ export class Umu {
     ensureExecutablePermission(umuBinaryPath);
 
     const launchEnv = {
-      PROTON_LOG: "1",
+      PROTON_LOG: options?.protonLogEnabled ? "1" : "0",
+      ...(options?.compatibilityEnvironmentVariables ?? {}),
```
Moyasee: `typescript:S7744`, the empty object is useless. Spreading `undefined` is already a no-op, so the fallback does nothing:

```ts
...options?.compatibilityEnvironmentVariables,
```

The `? {...} : {}` ternaries below are a different shape and Sonar is right to leave those alone.

src/renderer/src/pages/game-details/modals/game-options-modal/compatibility-section.tsx
```diff
@@ -254,6 +265,38 @@ export function CompatibilitySettingsSection({
             <Tooltip id={mangohudTooltipId} />
           )}
         </div>
+
+        <div className="game-options-modal__mangohud-toggle">
```
Moyasee: `typescript:S3776`, cognitive complexity 18 in `CompatibilitySettingsSection`, max 15. This block is what pushed it over: a tooltip id ternary, a class name ternary, a `secondaryText` ternary and a two-term `&&` chain in the JSX.

This is now the third copy of the same row. GameMode, MangoHud and Proton logging differ only in label, availability flag, global flag and handler. Pulling that shape into one local component takes the complexity back under the limit and deletes about forty lines rather than moving them.

src/renderer/src/pages/settings/settings-behavior.scss
```diff
@@ -108,3 +108,34 @@
     gap: calc(globals.$spacing-unit * 1.25);
   }
 }
+
+.settings-context-compatibility__env-vars {
```
Moyasee: Rename is right. One follow-up: these four now sit at top level in a file named for a different component, and `settings-behavior.scss` is imported by three components. The rest of the folder keeps one block per file, `emulation/settings-context-emulation.scss` being the closest example. A `settings-context-compatibility.scss` imported from the component it belongs to would match.

Moyasee (commented):
All six from the last pass are done. Verified:

```
src/locales/*      proton_logging=game_details, tooltip=game_details in en/ru/es/pt-BR/el
                   proton_logging_description and enable_proton_logging_description gone
big-picture exact  en 7/7  ru 7/7  es 7/7  pt-BR 7/7  fr 7/7  el 7/7
```

The SCSS rename, the restored Big Picture secondaryText, and the placeholder are all correct too.

That leaves SonarCloud. The check is green, but only because every gate condition is a new-code rating and 38 minutes of debt does not move one:

```
status: OK
  OK new_reliability_rating 1 GT 1
  OK new_security_rating 1 GT 1
  OK new_maintainability_rating 1 GT 1
  OK new_duplicated_lines_density 0.0 GT 3
  OK new_security_hotspots_reviewed 100.0 LT 100
```

Four issues are open underneath it, all in code this PR touches. Inline. Three are one-liners, the fourth is a small extraction.

src/renderer/src/pages/game-details/modals/game-options-modal/compatibility-section.tsx
```diff
@@ -131,28 +212,14 @@ export function CompatibilitySettingsSection({
...
-            data-tooltip-id="winetricks-unavailable-tooltip"
-            data-tooltip-content={
-              showWinetricksUnavailableTooltip
-                ? t("winetricks_not_available_tooltip")
-                : undefined
-            }
+          <Button
+            type="button"
+            theme="outline"
+            onClick={onOpenWinetricks}
```
Moyasee: The winetricks tooltip is gone. The wrapper span carrying `data-tooltip-id="winetricks-unavailable-tooltip"` and the `<Tooltip id="winetricks-unavailable-tooltip" />` below it were both dropped, so when winetricks is missing from PATH the button is disabled with nothing explaining why.

The string is now orphaned as well. Zero code references, 33 locale files:

```
$ git grep -l "winetricks_not_available_tooltip" -- src/
(only src/locales/*/translation.json, 33 files)
```

Unrelated to Proton logging either way. Put it back.

src/renderer/src/pages/game-details/modals/game-options-modal/compatibility-section.tsx
```diff
@@ -121,7 +202,7 @@ export function CompatibilitySettingsSection({
                 <FileDirectoryIcon />
                 {t("select_executable")}
               </Button>
-              {game.winePrefixPath && (
+              {displayedWinePrefixPath && (
```
Moyasee: This was `game.winePrefixPath`. `game-options-modal.tsx:750` defines the new value as

```ts
const displayedWinePrefixPath =
  game.winePrefixPath ?? defaultHydraWinePrefixPath;
```

so Clear now renders whenever a path is shown, including the Hydra default that every Linux game gets. In that state `handleClearWinePrefixPath` calls `selectGameWinePrefix(shop, objectId, null)` on a prefix that is already null, so the button appears and does nothing.

Revert to `game.winePrefixPath`, which is the only case where there is an override to clear.

src/renderer/src/pages/game-details/modals/game-options-modal/compatibility-section.tsx
```diff
@@ -72,32 +119,66 @@ export function CompatibilitySettingsSection({
...
-      defaultValue: "Installed in Steam compatibilitytools.d",
+      defaultValue:
+        "Proton installation found in compatibilitytools.d directories.",
     }
   );
 
   return (
     <>
-      <div className="game-options-modal__wine-prefix">
+      <div className="game-options-modal__section">
```
Moyasee: This was `game-options-modal__wine-prefix`. The two are not interchangeable:

```scss
&__wine-prefix { gap: calc(globals.$spacing-unit * 2); }
&__section     { gap: calc(globals.$spacing-unit * 1.5); }
```

So the wine prefix block tightens up, and `__wine-prefix` goes dead, no `.tsx` references it anymore. Unrelated layout change, revert it.

src/renderer/src/pages/game-details/modals/game-options-modal/compatibility-section.tsx
```diff
@@ -30,18 +32,62 @@ interface CompatibilitySettingsSectionProps {
...
+function CompatibilityToggleRow({
+  label,
+  checked,
+  disabled,
+  tooltipId,
+  tooltipContent,
+  onChange,
+}: Readonly<CompatibilityToggleRowProps>) {
+  return (
+    <div className="game-options-modal__mangohud-toggle">
```
Moyasee: Hardcoding the MangoHud block name means the GameMode row now renders `__mangohud-toggle` and `__mangohud-label`, and `__gamemode-toggle` / `__gamemode-label` / `__gamemode-label--disabled` are left with no consumers.

Nothing looks different today because the two rule sets are byte for byte identical, which is exactly why this is easy to miss and easy to break later when one of them is edited.

Either take the block name as a prop, or rename both scss blocks to something neutral like `__toggle` and `__toggle-label` and delete the duplicate. The second is less code.

src/renderer/src/pages/game-details/modals/game-options-modal/compatibility-section.tsx
```diff
@@ -72,32 +119,66 @@ export function CompatibilitySettingsSection({
...
+
   const protonVersionAutoLabel = t("proton_version_auto", {
     ns: ["game_details", "settings"],
     defaultValue: "Auto (global default or umu default)",
   });
 
   const protonSourceUmuDefault = t("proton_source_umu_default", {
     ns: ["game_details", "settings"],
-    defaultValue: "umu default selection",
+    defaultValue: "Uses the default UMU-managed Proton version.",
```
Moyasee: These three `defaultValue` rewrites are dead weight. All of `proton_source_umu_default`, `proton_source_steam` and `proton_source_compatibility_tools` resolve in `game_details`, so the fallback never renders and the visible text is unchanged. If the wording should change, change it in `src/locales/en/translation.json` and the other three. Otherwise drop the hunk.

Moyasee (commented):
Sonar is clear. Checked the API rather than the badge, since the gate was passing while issues were open last time:

```
UNRESOLVED total: 0
RESOLVED total: 4
  compatibility-section.tsx  S3776  CLOSED FIXED
  launch-game.ts             S6353  CLOSED FIXED
  umu.ts                     S7744  CLOSED FIXED
  launch-game.ts             S107   CLOSED FIXED
```

Exporting `envVariableNameRegex` from `resolve-launch-command.ts` rather than just swapping in `\w` is the better call, the two validators can't drift now. Options object, the `?? {}` removal and the dedicated `settings-context-compatibility.scss` all look right.

The extraction itself is fine. What I want to flag is what rode along with it in d8a940654: five changes to `compatibility-section.tsx` that have nothing to do with Proton logging, two of which are user-visible regressions. Inline. Easiest path is probably to strip them back out of this PR and open a separate one for the wine prefix and winetricks bits if you want them.

src/renderer/src/pages/game-details/modals/game-options-modal/compatibility-section.tsx
```diff
@@ -161,99 +247,65 @@ export function CompatibilitySettingsSection({
...
 
-          {mangohudToggleDisabled && mangohudTooltipId && (
-            <Tooltip id={mangohudTooltipId} />
-          )}
-        </div>
+        <CompatibilityToggleRow
+          label={<span>{t("proton_logging")}</span>}
+          checked={protonLogEnabled || globalProtonLogEnabled}
+          disabled={protonLogToggleDisabled}
+          classNamePrefix="proton-log"
```
Moyasee: `gamemode` and `mangohud` resolve to real rules. `proton-log` does not exist:

```
$ git grep -rn "proton-log" -- '*.scss'
(no matches)

$ grep -E "&__[a-z-]+(-toggle|-label) \{" game-options-modal.scss
161:  &__mangohud-toggle {
168:  &__gamemode-toggle {
175:  &__mangohud-label {
186:  &__gamemode-label {
```

So this row now emits `__proton-log-toggle`, `__proton-log-label` and `__proton-log-label--disabled`, and none of them match anything. It loses the container's `display: flex` and `justify-content: space-between`, so it falls back to block layout and stops lining up with the two rows above it, and it loses the label's `inline-flex`.

The one that matters most is `--disabled` `text-decoration: underline dotted`. That underline is what tells the user the row is locked and worth hovering, and it is the only affordance pointing at the "enabled globally" tooltip. Without it the tooltip is undiscoverable, on the row this PR is here to add. Before this commit the row borrowed the mangohud classes and rendered correctly.

Adding a third `&__proton-log-*` block fixes it but leaves three byte-identical copies. Collapsing all three to `__toggle` and `__toggle-label` and dropping `classNamePrefix` is smaller and removes the duplication at the same time.

Moyasee (commented):
All five are reverted correctly. Winetricks tooltip is back, the Clear button is back on `game.winePrefixPath`, `__wine-prefix` is back on the container, and the `proton_source_*` defaultValues are back to what they were.

One thing came out of the `classNamePrefix` change, inline. Everything else in this PR looks settled from my side.

Moyasee (approved):
LGTM!

Moyasee:
Could you please fix the merge conflicts?

Cryptorial0:
@Moyasee done

src/main/helpers/launch-game.ts
```diff
@@ -379,16 +417,34 @@ const cleanupStaleCompatibilityProcesses = async (
...
+    parsedPath,
+    compatibilityContext,
+    launchOptions,
+    useMangohud,
+    useGamemode,
+    protonLogEnabled,
+    compatibilityEnvironmentVariables,
+  } = options;
   const { protonPath, winePrefixPath } = compatibilityContext;
+  await cleanupStaleCompatibilityProcesses(objectId, winePrefixPath);
```
Moyasee: This is merge residue. `fe031c7be` on main moved this call into `prepareLinuxCompatibilityForLaunch`, and the branch already has it there at 522, so both copies survived the resolution.

Ordering is the reason it matters: main calls it before `prepareCompatibilityPrefixForCloudSave` and before the pre-launch restore, deliberately. This one runs after both, and it SIGKILLs anything whose `STEAM_COMPAT_DATA_PATH` matches the prefix and whose name is wineserver, ends in `.exe`, or whose path contains wine. On Linux with a default prefix and cloud save v2 auto-sync on, that is the wineserver `Umu.preparePrefix` just used, killed instead of shut down, dropping unflushed `system.reg` and `user.reg` writes. I can't reproduce that from here without Linux and umu, but the ordering is plain in the source. Delete this line.

src/main/helpers/launch-game.ts
```diff
@@ -518,16 +577,34 @@ const runCommonRedistPreflight = async (shop: GameShop, objectId: string) => {
...
 const launchResolvedGame = async (
-  gameKey: string,
-  shop: GameShop,
-  objectId: string,
-  parsedPath: string,
-  compatibilityContext: LinuxCompatibilityLaunchContext | null,
-  launchOptions: string | null | undefined,
-  useMangohud: boolean,
-  useGamemode: boolean
+  options: LaunchResolvedGameOptions
```
Moyasee: Unformatted, `yarn format` fixes it. Also at 811, which came in with the merge commit. `lint` stays green because the repo uses `eslint-config-prettier` rather than `eslint-plugin-prettier` and there is no `format-check` job, so nothing catches it.

src/big-picture/src/locales/fr/translation.json
```diff
@@ -624,6 +624,13 @@
     "RPCS3 is already running": "RPCS3 est déjà en cours d'exécution",
     "Run with GameMode": "Exécuter avec GameMode",
     "Run with MangoHud": "Exécuter avec MangoHud",
+    "Enable Proton logging": "Activer la journalisation de Proton",
+    "Write Proton debug logs for compatibility launches.": "Écrire les journaux de débogage Proton pour les lancements de compatibilité.",
+    "Proton logging": "Journalisation de Proton",
+    "Write Proton debug logs for this game. Overrides the global setting.": "Écrire les journaux de débogage Proton pour ce jeu. Remplace le параметр global.",
```
Moyasee: `параметр` should be `paramètre`.

src/big-picture/src/locales/pt-BR/translation.json
```diff
@@ -451,6 +451,13 @@
     "RPCS3 is already running": "RPCS3 já está em execução",
     "Run with GameMode": "Executar com GameMode",
     "Run with MangoHud": "Executar com MangoHud",
+    "Enable Proton logging": "Ativar registro do Proton",
+    "Write Proton debug logs for compatibility launches.": "Escrever logs de depuração do Proton para inicializações de compatibilidade.",
+    "Proton logging": "Registro do Proton",
+    "Write Proton debug logs for this game. Overrides the global setting.": "Escrever logs de depuração do Proton para este jogo. Sobresscreve a configuração global.",
```
Moyasee: `Sobresscreve` -> `Sobrescreve`.

src/big-picture/src/pages/settings/compatibility.tsx
```diff
@@ -387,6 +412,69 @@ export function CompatibilitySettingsSection({
...
+              />
+
+              <FocusItem
+                id={COMPATIBILITY_ENV_VARS_INPUT_ID}
+                actions={{ primary: () => textareaRef.current?.focus() }}
+              >
+                <div className="compatibility-settings-section__env-vars-group">
+                  <label
+                    htmlFor="compatibility-environment-variables"
+                    className="compatibility-settings-section__env-vars-label"
```
Moyasee: Nit. `__env-vars-label` and `__env-vars-help` have no rules, `compatibility.scss` only defines `__env-vars-group` and `__env-vars-textarea`. Same thing that came up twice already against `settings-behavior.scss`.

Moyasee (changes requested):
The merge lost nothing from main, but it kept both sides of one hunk in `launch-game.ts` and the stale process sweep now runs twice. Rest is one-liners. Notes inline. Branch is 27 behind so it will want another merge before it lands, and the thread on the big picture locale is already fixed at head, just needs closing.

Moyasee (approved):
LGTM!

Moyasee (commented):
Approval stands, both of these are follow-ups rather than blockers. Tested the global variables end to end here, adding and removing HDR-related ones and checking they landed in the game, and Proton logging works from both the global setting and the game page.

Two things inline.

src/types/level.types.ts
```diff
@@ -182,6 +183,8 @@ export interface UserPreferences {
   globalTrackersUrl?: string;
   appendGlobalTrackersUrl?: boolean;
   defaultProtonPath?: string | null;
+  protonLogEnabled?: boolean;
+  compatibilityEnvironmentVariables?: string | null;
```
Moyasee, line 191: Worth a toggle alongside this so the variables can be switched off without losing what was typed. Right now clearing the textarea is the only way to disable them, and then you have to type it all back in to re-enable.

The shape already exists twice: trackers store `globalTrackers: string[]` next to `appendGlobalTrackers: boolean`, and this PR adds `protonLogEnabled` on the line right above. Something like `compatibilityEnvironmentVariablesEnabled`, checked at `launch-game.ts:808`, would match both.

`#` gives a per-line version of this today, which is nice, but there is no way to switch the whole set off at once.

src/renderer/src/pages/settings/settings-context-compatibility.tsx
```diff
@@ -370,6 +381,54 @@ export function SettingsContextCompatibility() {
...
+                <textarea
+                  id="compatibility-environment-variables"
+                  className="settings-context-compatibility__textarea"
+                  value={compatibilityEnvironmentVariables}
+                  onChange={(event) =>
+                    setCompatibilityEnvironmentVariables(event.target.value)
+                  }
+                  onBlur={() =>
+                    updateUserPreferences({
+                      compatibilityEnvironmentVariables:
```
Moyasee: Blank lines persist through here and come back into the textarea on reload. No runtime effect, the parser at `launch-game.ts:68` skips empty and `#` lines before anything reaches the process, so this is cosmetic only.

Raising it because the trackers feature does the opposite: `save-global-trackers.ts` filters and dedupes before storing, and the field rehydrates from `join("\n")`, so blanks disappear on save. Two similar-looking textareas behaving differently is the sort of thing people report as a bug.

Not a straight copy though. Trackers can normalize because they persist a `string[]`; this persists raw text, which is exactly what makes `#` comments work. If you do trim, drop only the fully blank lines and leave comment lines alone.

Cryptorial0: on it

src/big-picture/src/pages/settings/compatibility.tsx
```diff
@@ -387,6 +421,86 @@ export function CompatibilitySettingsSection({
...
+                focusId={COMPATIBILITY_PROTON_LOGGING_FOCUS_ID}
+                block
+                onChange={(checked) => {
+                  void updateCompatibilityPreferences({
+                    protonLogEnabled: checked,
+                  });
+                }}
+              />
+
+              <Checkbox
```
Moyasee, line 468: The new checkbox does not respond and the textarea below it cannot be typed into. The `items` memo reads six `form` fields and its dependency array at 533-549 lists three:

```
reads: defaultProtonPath, autoRunGamemode, autoRunMangohud,
       protonLogEnabled, compatibilityEnvironmentVariablesEnabled,
       compatibilityEnvironmentVariables
deps:  defaultProtonPath, autoRunGamemode, autoRunMangohud
```

`updateCompatibilityPreferences` is `useCallback(..., [])`, so nothing else invalidates the memo and the render closures stay pinned to a stale `form`:

```
initial:      checkbox.checked = false   textarea.value = ""
after click:  checkbox.checked = false   <-- expected true
after typing: textarea.value  = ""       <-- expected "PROTON_FSR4_UPGRADE=1"

with the three deps added:
after click:  checkbox.checked = true
after typing: textarea.value  = "PROTON_FSR4_UPGRADE=1"
```

Only `compatibilityEnvironmentVariablesEnabled` is new here. `protonLogEnabled` and `compatibilityEnvironmentVariables` were already missing at the commit I approved, so the big picture proton logging toggle is dead too and I should have caught it. Adding the three to 533-549 fixes all of it.

Minor, same block: this checkbox is missing the `disabled={!canUseBehaviorSection}` its three siblings carry. Only reachable in a dev build off Linux, so cosmetic.

src/locales/en/translation.json
```diff
@@ -1230,6 +1232,11 @@
     "enable_auto_install": "Download updates automatically",
     "run_games_with_gamemode": "Run games with GameMode by default",
     "run_games_with_mangohud": "Run games with MangoHud by default",
+    "enable_proton_logging": "Enable Proton logging",
+    "enable_compatibility_environment_variables": "Enable environment variables",
```
Moyasee, line 1241: `enable_compatibility_environment_variables` is in `en` only. ru, es and pt-BR have nothing, and `fallbackLng` is `"en"`.

Big picture is worse: the two new literals at `compatibility.tsx:450-451` are in no bundle at all.

```
"Enable environment variables"                 en 0  ru 0  es 0  pt-BR 0  fr 0  el 0
"Controls whether compatibility environment…"  en 0  ru 0  es 0  pt-BR 0  fr 0  el 0
```

`translateExactText` is `exactTranslations[language][sourceText] ?? sourceText`, so they render English everywhere and the `en` identity entry is needed as well. The base PR did this correctly for all seven of its literals, so it is just this commit that skipped it.

src/big-picture/src/pages/settings/compatibility.tsx
```diff
@@ -387,6 +421,86 @@ export function CompatibilitySettingsSection({
...
+                id={`${COMPATIBILITY_ENV_VARS_INPUT_ID}_enable`}
+                label="Enable environment variables"
+                secondaryText="Controls whether compatibility environment variables are injected."
+                checked={form.compatibilityEnvironmentVariablesEnabled}
+                focusId={`${COMPATIBILITY_ENV_VARS_INPUT_ID}_enable`}
+                block
+                onChange={(checked) => {
```
Moyasee, line 477: Once the toggle exists, the textarea should be disabled or dimmed while it is off. As it stands a user can turn the toggle off, type variables, watch them save, and get nothing at launch with no indication why. Same on the desktop side.

src/types/level.types.ts
```diff
@@ -182,6 +183,9 @@ export interface UserPreferences {
...
+  compatibilityEnvironmentVariablesEnabled?: boolean | null;
```
Moyasee: `boolean | null` where `protonLogEnabled` two lines up is plain `boolean`. Nothing writes null here, so the `| null` can go.

Moyasee (changes requested):
Withdrawing my approval on this, and part of it is my miss rather than the new commit's.

The toggle works on desktop, and the blank-line trimming is right, including keeping `#` lines, which was the thing I was worried about when I asked for it.

But in big picture the new checkbox and the environment variables textarea are both inert, and the dependency array behind that was already stale at the commit I approved, so the proton logging toggle there has been dead the whole time. That one is on me.

Separately the new strings landed in `en` only on desktop, and in no bundle at all in big picture.

The Windows build failed on a code-signing binary download rather than on anything here, so it just wants a re-run. Branch is behind main too.

PR #2645, opened 2026-08-05
feat(cloud-save): support multi-profile discovery and state-driven sync

Hachi-R:
## Summary

- discover and preserve multiple save profiles, including literal emulator folders and Steam IDs
- carry complete variant and directory coverage through merge and restore decisions
- centralize automatic synchronization planning around the observed local and remote state
- repeat automatic sync when the relevant state changes without duplicating completed work

## Impact

Games with multiple profiles no longer treat one Steam account as authoritative or silently omit other valid profiles. Empty/deleted profile folders and wildcard-derived identities remain distinguishable, while automatic sync uses one state-driven decision path.

## Automatic sync migration

- V2 becomes the default automatic sync mode for Steam games.
- A one-time atomic migration removes settings left by the alpha V2 rollout and disables legacy automatic sync.
- After that migration, users can still choose V1, V2, or leave both modes disabled.

## Validation

- Rust: 150 passed
- Node: 343 passed, 4 skipped; the only failure is the known Windows `EBUSY` cleanup hook in `notification-icon.test.ts`
- Targeted Node tests: 15 passed
- Node/Web typechecks: passed
- Native addon build: passed
- ESLint: 0 errors; 80 existing warnings

## PR chain

Depends on #2614 (`codex/cloud-saves-v2-big-picture`).


native/hydra-native/src/cloud_save/identity/mod.rs
```diff
@@ -179,31 +177,10 @@ pub fn build_rule_id(rule: &CloudSaveRule) -> String {
...
-fn account_matches(account: &KnownStoreAccount, captured: &str) -> Option<(String, String)> {
-    let validated = validated_steam_account(account)?;
-    ((captured == validated.0) || (captured == validated.1)).then_some(validated)
-}
-
 pub fn store_user_identity(
     store: &str,
     captured: Option<&str>,
-    context: &StoreUserContext,
+    _context: &StoreUserContext,
```
Moyasee: `store_user_identity` is the only consumer of `store_user_context`, and after this change it ignores it. That makes the whole chain dead: `ResolveSaveRulesInput.store_user_context` -> `collect_discovered_files` -> here. On the TS side it means `getCloudSaveGameContext` still reads and VDF-parses Steam's `loginusers.vdf` on every context build for a value nothing reads.

`toSteamAccount` stays live for validating legacy `steam-account` variants in `cloud-save-contract.ts`, so it is only the context plumbing that is dead.

Drop the parameter and the input field, or say in the PR body why it is being held for a later step.

src/main/services/cloud-save/automatic-sync-observation.ts
```diff
@@ -0,0 +1,135 @@
...
+  observationKey: string
+) => {
+  const key = gameKey(objectId, shop);
+  const previousObservationKey = latestObservations.get(key);
+  latestObservations.set(key, observationKey);
+
+  // A fingerprint may legitimately reappear after an intermediate state. For
+  // example: empty local files -> restored -> empty local files again. Once a
+  // different state has been observed, an older settled attempt must not
+  // suppress the new transition back to that fingerprint.
```
Moyasee: This should come out, with the reasoning moved to the commit message or the PR body.

The `if (previousObservationKey !== undefined && previousObservationKey !== observationKey)` block below reads as its own named function, something like `invalidateSettledAttemptOnStateChange`, which carries the same meaning in the code.

src/main/services/cloud-save/sync-game/planner.ts
```diff
@@ -0,0 +1,113 @@
...
+      execution: "none",
+      reason: "game-running",
+    };
+  }
+
+  const hasLocalChangesToApply =
+    restoreEntryCount > 0 || deleteLocalEntryCount > 0;
+
+  // An empty local snapshot is never proof that every remote save should be
+  // deleted. Full deletion is an explicit operation outside the sync flow.
```
Moyasee: Same here, move it to the commit message or PR body.

`const emptyLocalCannotDeleteRemote = !hasLocalFiles && hasRemoteSnapshot` in front of the `if` says it without the comment.

src/main/services/logger.ts
```diff
@@ -2,6 +2,13 @@ import { logsPath } from "@main/constants";
 import log from "electron-log";
 import path from "path";
 
+type DepthAwareTransport = log.Transport & { depth: number };
+
+// Main-process logs reach DevTools through electron-log's IPC transport. Its
+// default depth is too shallow for API payloads, turning array entries into
+// "[object]" before the renderer can inspect them.
+(log.transports.ipc as DepthAwareTransport).depth = 12;
```
Moyasee: Two things.

The comment should come out, explanation into the commit message.

More substantively, this sets the depth globally, so every main-process log now serializes 12 levels deep over IPC, not just the API payloads this PR wants to inspect. Combined with the response interceptor logging full bodies, that is a noticeable widening of what crosses to the renderer on every log call. Worth checking whether it can be scoped to the network logger instead.

src/main/services/cloud-save/automatic-sync-mode.ts
```diff
@@ -22,9 +22,21 @@ export const resolveStoredCloudSaveAutomaticSyncMode = (
 ) =>
   resolveCloudSaveAutomaticSyncMode({
     legacyEnabled,
-    v2Enabled: storedV2Enabled === true,
+    v2Enabled: storedV2Enabled ?? true,
```
Moyasee, line 25: This flips the V2 automatic sync default from off to on for any Steam game with no stored setting. Worth stating explicitly in the PR body, since it changes behavior on existing installs and not only for new ones.

Related, and the part I would call out in the commit message: `migrateCloudSaveAutomaticSyncDefaultsWithStore` deletes every steam key in the settings sublevel, not only the ones it is migrating. That is safe only because `persistCloudSaveAutomaticSyncMode` used to `del` on disable and so could never store `false`. This same PR changes it to always `put` the boolean. The completion flag means ordering works today, but the assumption is invisible at the call site.

Moyasee (commented):
Reviewed as part of the stack, findings inline.

The `hydra-api.ts` / `logger.ts` / `network-log-payload.ts` / `process-watcher.ts` / `game-running-state.ts` changes are unrelated to multi-profile discovery. Worth splitting out, mainly because `sanitizeNetworkLogPayload` is a real security fix that deserves to be visible on its own. The old `logger.error("401 - Current credentials:", this.userAuth, ...)` path was writing raw access and refresh tokens into the log file, and the new sanitizer redacts them by key. That gets lost in a 79-file cloud save PR.

Locales: this adds `en` and `pt-BR` only. Not something this PR introduces, `es` and `ru` are already at 6 `cloud_save_v2` keys against 195 in `en`, but the gap is worth one catch-up pass before the stack merges rather than per-PR.

Hachi-R:
Implemented the actionable review feedback in `936482511` and propagated it through #2646 and #2647.

### Addressed

- Removed the unused native `store_user_context` from the Rust snapshot-building input and discovery pipeline, and simplified `store_user_identity` accordingly. The existing default and opaque-folder variant IDs remain unchanged and are covered by exact-ID tests.
- Isolated `depth = 12` to a dedicated `networkLogger` instance. The main logger keeps electron-log's default IPC depth, while network logs still go to `network.txt` and retain the existing token/payload sanitization.
- Extracted `invalidateSettledGamePageAttemptOnObservationChange` and replaced the empty-local explanatory comment with the named `emptyLocalRequiresRemotePreservation` condition. These are readability-only changes.
- Updated the PR description to document the automatic-sync migration behavior.

### Intentionally not changed

- The TypeScript `storeUserContext` remains. Custom-path binding resolution still consumes the active and known store IDs to resolve `<storeUserId>` on the current device; removing it there would break custom-path resolution. Only the unused native copy was removed.
- The automatic-sync migration remains unchanged. Its behavior is intentional: V2 becomes the default for Steam games, the one-time atomic migration clears settings left by the alpha rollout and disables legacy sync, and users can subsequently choose V1, V2, or leave both disabled. Changing it here would alter the approved rollout semantics rather than fix a defect.
- No sync policy, multi-profile discovery, custom-path behavior, translations, or findings belonging to later PRs were changed, keeping this follow-up scoped to #2645.

Validation at the stack tip: 162/162 Rust tests passed; Node/Web typechecks, ESLint, Prettier, and `cargo fmt --check` passed. The Node suite passed 372 tests and skipped 4; its only failure is the known Windows `EBUSY` cleanup hook in `notification-icon.test.ts`.

I left the review threads unresolved so the reviewers can verify the changes.

native/hydra-native/src/cloud_save/identity/mod.rs
Moyasee: Fixed in 936482511.

Correcting one thing I got wrong above: only the discovery half was dead. `storeUserContext` is still read on the TS side by `storeUserIdsFromContext` in `custom-path.ts`, so `getSteamStoreUserContext` earns its keep.

One residual, outside the diff so I could not anchor it inline: `ResolveRestoreTargetsInput.store_user_context` at `native/hydra-native/src/cloud_save/restore/types.rs:38` is still declared and still never read. `resolve_restore_targets_inner` builds its `ResolveSaveRulesInput` from `home_dir` / `executable_path` / `wine_prefix_path` / `steam_path` only, and TS satisfies the required napi field by accident through the `...pathContext` spread at `resolve-remote-snapshot-targets.ts:120`. Same pattern as the one you just cut, restore side.

src/main/services/logger.ts
Moyasee: Fixed in 936482511, and the separate instance is a better answer than the scoping I suggested.

Correcting the second half of my comment: the ipc transport factory sets `level: logger.isDev ? 'silly' : false`, so that transport is off outside dev on both instances and the depth bump never reached production. The change is still the right one, it just fixes dev noise rather than a production cost.

Checked the split holds up. `Logger.create()` forwards `dependencies`, `isDev` and `transportFactories`, so `networkLog.transports.ipc` is constructed by the same factory and is defined under Electron. Dropping the `scope === "network"` branch from the default `resolvePathFn` leaves exactly one writer for `network.txt`.

Moyasee (commented):
Re-reviewed at 936482511. All five points from the previous pass are addressed, and the PR diff is comment-free again including the new revision.

`store_user_identity(store, captured)` losing the parameter, plus the removal from `BuildLocalGameSnapshotPipelineInput` and `collect_discovered_files`, is the right cut. `invalidateSettledGamePageAttemptOnObservationChange` and `emptyLocalRequiresRemotePreservation` both read better than the comments they replace. Follow-ups on the `identity/mod.rs` and `logger.ts` threads, including one residual dead field on the restore side and a correction to two things I got wrong the first time.

Validation at 936482511:

```
Rust:   155 passed; 0 failed
Node:   342 passed; 0 failed; 0 skipped
typecheck:node + typecheck:web: pass
prettier --check: pass
```

The numbers in the PR body are a revision stale (Rust 150, Node 338 with 4 skipped).

src/main/services/cloud-save/automatic-sync-observation.ts
```diff
@@ -0,0 +1,145 @@
...
+const latestObservations = new Map<string, string>();
+const attempts = new Map<string, AutomaticSyncAttempt>();
+
+const gameKey = (objectId: string, shop: GameShop) =>
+  JSON.stringify([shop, objectId]);
+const attemptKey = (objectId: string, shop: GameShop, trigger: string) =>
+  JSON.stringify([shop, objectId, trigger]);
+const byJson = (left: unknown, right: unknown) =>
+  JSON.stringify(left).localeCompare(JSON.stringify(right));
+const byText = (left: string, right: string) => left.localeCompare(right);
```
Moyasee: This is what Sonar asked for, but `localeCompare` treats canonically-equivalent strings as equal. On this machine:

```
"Café.sav" (NFC) .localeCompare("Café.sav" (NFD))  ->  0
```

Not a live bug, and I am not asking you to hold the merge for it. The key never leaves the process, the upstream arrays are already deterministically ordered, and V8's sort is stable so ties keep input order. It only bites if this fingerprint is ever persisted or compared across machines.

A code-unit comparator satisfies the rule without the locale dependency, if you want to close it off:

```ts
const byText = (left: string, right: string) =>
  left < right ? -1 : left > right ? 1 : 0;
```

src/main/services/game-running-state.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const gamesPlaytime = new Map<string, GamePlaytimeState>();
+
+export const setGamePlaytime = (
+  gameKey: string,
+  playtime: GamePlaytimeState
+) => {
+  gamesPlaytime.set(gameKey, playtime);
+};
```
Moyasee, line 27: This setter guards no invariant. `gamesPlaytime` is still exported mutable, and `process-watcher.ts`, `sign-out.ts` and `main-loop.ts` call `.delete()`, `.clear()`, `.get()`, `.has()`, `.entries()` and `.keys()` on it directly across ten sites, so the only thing routed through here is the write Sonar wanted to see.

Either move the deletes and the clear behind the module too, or drop the setter and mark the Sonar issue won't-fix. Half-encapsulation reads as if there is an invariant to protect when there is not.

src/main/services/network-log-payload.ts
```diff
@@ -0,0 +1,106 @@
...
+]);
+
+const normalizedKey = (key: string) => key.toLocaleLowerCase("en-US");
+
+const redactSensitiveUrlParameters = (value: string) => {
+  if (!/^https?:\/\//i.test(value)) return value;
+
+  try {
+    const parsed = new URL(value);
+    const queryKeys = new Set(parsed.searchParams.keys());
```
Moyasee, line 40: Worth noting that this is not only the style change Sonar asked for. The `Set` also dedups, so `?token=a&part=1&token=b` now collapses to a single redacted `token` instead of two. Better behavior and the new test pins it, but it deserves its own line in the commit message rather than riding along under "address Sonar findings".

Moyasee (approved):
Approving. Reviewed the whole PR again at 63d4c89f5, deeper this time on the native path resolution and the scanner, and I did not find anything blocking.

Everything from the previous two passes is addressed, and the Sonar gate is green again. Three issues stay open there, all `CODE_SMELL` rather than `BUG`, which is why the gate passes: cognitive complexity 90 on `mergeUserVariantSnapshots`, 23 on `planCloudSaveSync`, 17 in `sync-game-cloud-save.ts`. Not blocking, but `mergeUserVariantSnapshots` at 90 is genuinely hard to follow and this PR adds to it, so it is worth a follow-up before the stack lands.

Three small things inline, none of which need to hold the merge.

The deeper pass turned up no defects. Recording what I checked so it does not get re-checked later:

- `capture.rs` `component_offsets_from_end` counts from the end, so it is invariant to how a Windows drive prefix gets segmented, and `values.len() - 1` cannot underflow because `values.first()?` returns first.
- `resolve_rules.rs` clearing `unresolved_tokens` for a `<custom>` rule with a `preferredPath` is safe. The foreign-environment case returns before it, and a stale preferred path yields `confirmed-missing` coverage, which fails `provesDeletion`, so remote files are preserved rather than deleted.
- `isCloudSavePathWithinRoot` handles bare `..`, the `../` prefix, and the cross-drive case where `path.win32.relative` returns an absolute path. Lowercasing before the compare is needed since Node's `relative` is case-sensitive even on win32.
- The migration in `loadState()` is unguarded and pulls the whole games sublevel into memory, but it sits directly under `clearLegacyAchievementPersistence()` with the same shape, so it is not a new hazard.

Still outstanding from the earlier threads: the dead `ResolveRestoreTargetsInput.store_user_context` at `native/hydra-native/src/cloud_save/restore/types.rs:38`, and the PR body's validation numbers, now two revisions stale.

Validation at 63d4c89f5:

```
Rust:      155 passed; 0 failed
Node:      345 passed; 0 failed; 0 skipped
typecheck: node + web pass
prettier:  pass
eslint:    0 errors, 80 warnings (pre-existing)
Sonar:     gate SUCCESS
```

src/main/services/game-running-state.ts
Hachi-R: Addressed in 5195d1775. The mutable Map is now private, `gamesPlaytime` is exported as a `ReadonlyMap`, and set/delete/clear all go through the state module. `process-watcher`, sign-out, and the state tests were migrated without changing keys, values, or read behavior.

src/main/services/network-log-payload.ts
Hachi-R: Double-checked this one: there is no output behavior change from introducing the `Set`. `URLSearchParams.set(name, value)` already removes the additional values for that name, so the previous loop collapsed duplicate `token` parameters on its first call as well. The `Set` only avoids repeating the same `set` call; the duplicate-parameter test pins the existing serialized result. No code change was needed for this thread.

src/main/services/cloud-save/automatic-sync-observation.ts
Hachi-R: Addressed in 5195d1775. `byText` now uses explicit code-unit ordering with no locale dependency, and the observation tests include reversed NFC/NFD inputs to pin deterministic fingerprints.

PR #2646, opened 2026-08-05
fix(cloud-save): reconcile custom paths and restore targets

Hachi-R:
## Summary

- make custom path tracking and removal authoritative across local bindings and the remote snapshot
- reconcile unresolved remote identities with current local restore targets and file metadata
- preserve local files while untracking custom locations, including removal of the last tracked path
- improve pending-destination, loading, and error presentation for custom save operations

## Impact

Custom paths no longer remain inconsistently tracked across devices or appear missing when their resolved target already exists. Reconciliation keeps identity and physical destination separate, and destructive snapshot changes occur only through the explicit flows.

## Validation

- Rust: 159 passed
- Node: 367 passed, 4 skipped
- Node/Web typechecks: passed
- ESLint: 0 errors; existing warnings only
- Known non-blocking Windows failure: `EPERM` while cleaning the notification icon test fixture

## PR chain

Depends on #2645 (`codex/cloud-saves-v2-multi-profile-sync`).


src/main/services/cloud-save/custom-path-removal.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (proposal.files.length === 0) {
+    if (
+      proposal.customPathRawPaths.length > 0 ||
+      proposal.variants.length > 0
+    ) {
+      throw new Error("cloud_save_custom_path_removal_invalid_empty_snapshot");
+    }
+    await deleteSnapshot();
+    return "snapshot-deleted";
+  }
```
Moyasee: This throw is reachable, and when it fires the user cannot untrack the location at all.

`buildPrepareSnapshotPayload` only validates the converse (no file may reference an inactive path), and `mergedCustomPathRawPaths` in `analyze-cloud-save-state` is a plain union of the remote list and `bindings.ready`, independent of whether a path contributed files. So a snapshot can legitimately hold a custom path with zero files.

Repro: game with no automatic-rule files, custom paths A and B both tracked, B's folder emptied locally and synced. Snapshot is now `customPathRawPaths: [A, B]`, `files: [<from A>]`. Untracking A:

```
proposal.changed            === true
proposal.files              === []
proposal.customPathRawPaths === ["<custom>...B"]
-> rejects cloud_save_custom_path_removal_invalid_empty_snapshot
-> deleteSnapshot not called, updateSnapshot not called
```

`executeCloudSaveCustomPathUntracking` runs `publishRemoval()` before `removeBinding()`, so nothing is removed locally either, and `shouldRetryCloudSaveConflict` does not cover this error. The user gets `cloud_save_v2_custom_path_remove_error_description` ("The location is still being tracked") with no way forward.

Fix at either end: enforce that `customPathRawPaths` is a subset of the paths that actually contributed files at upload time, or handle the case here instead of asserting on it.

src/main/services/cloud-save/get-cloud-save-overview.ts
```diff
@@ -28,6 +29,11 @@ export const getCloudSaveOverview = async (
     ...(analysis.anchor?.unresolvedRemoteEntryIds ?? []),
     ...analysis.merge.unresolvedRemoteEntryIds,
   ]);
+  const unconfiguredCustomPathCount =
+    getUnconfiguredCloudSaveCustomPathCandidates(
+      analysis.remoteManifest?.files ?? [],
+      analysis.customPathBindings.ready.map(({ rawPath }) => rawPath)
+    ).length;
```
Moyasee, line 39: Passing only `ready` means a path that is tracked but currently in `bindings.unresolved` (unmounted drive, missing Wine prefix, anything classifying as `wine-prefix-unavailable`) counts as unconfigured. Two effects:

- `shouldSyncCloudSaveOnGamePage` returns false, so automatic sync on game-page open silently stops until the condition clears.
- The panel asks the user to configure a location they already configured, while the file browser separately renders that same path under `details.unresolvedCustomPaths` with its recoverable reason.

`[...ready, ...unresolved].map(({ rawPath }) => rawPath)` keeps the "never seen on this device" signal without swallowing the recoverable case.

src/main/services/cloud-save/sync-anchor.ts
```diff
@@ -100,7 +103,36 @@ export const getCloudSaveSyncAnchor = async (
   await cloudSaveSyncAnchorsSublevel
     .del(await getLegacyAnchorKey(shop, objectId))
     .catch(() => undefined);
-  return null;
+  if (!options.allowEnvironmentFallback) return null;
```
Moyasee, line 106: Checking intent on this one.

The fallback returns the most recently updated anchor for the game from any environment, and `analyze-cloud-save-state` calls it with `allowEnvironmentFallback: !activeRemoteSnapshot`. That anchor then feeds `buildRemoteSnapshotDeletionPlan`, so after a remote snapshot is deleted elsewhere, local files that are byte-identical to an anchor recorded under a different `environmentId` land in `unchangedAutomaticEntryIds` and are deleted with no conflict prompt.

Before this change the function returned `null` there and `executeRemoteSnapshotDeletionSync` never ran at all. Entry identity is environment-independent so the byte match is meaningful, but it is a real widening of when Hydra deletes local saves. Intentional?

src/main/services/cloud-save/sync-game-cloud-save.ts
```diff
@@ -461,27 +678,81 @@ const executeGameCloudSaveSync = async ({
...
+    if (remoteDeletionOutcome.kind === "retry") {
+      return executeGameCloudSaveSync({
+        objectId,
+        shop,
+        trigger,
+        emitProgress,
+        resolution,
+        suppliedContext,
+      });
+    }
```
Moyasee, line 716: No attempt counter here, unlike `runGameCloudSaveSync` which caps at 1. Each pass awaits the network so it will not blow the stack, but a flapping remote (snapshot present in the outer analyze, absent in the inner one, repeatedly) livelocks the sync. Threading the existing `attempt` through closes it.

src/main/services/cloud-save/reconcile-remote-target-observations.ts
```diff
@@ -0,0 +1,196 @@
...
+  }
+  return hash.digest("hex");
+};
+
+const pathKey = (value: string, context: LocalGameSnapshotContext) => {
+  let canonical = path.resolve(value);
+  try {
+    canonical = realpathSync.native(canonical);
+  } catch {
+    // A missing target has no filesystem identity yet; lexical identity is enough.
```
Moyasee: This one should come out too, reasoning into the commit message. If the intent needs to be visible in the code, lifting it to a named helper does it:

```ts
const canonicalOrLexical = (value: string) => {
  try {
    return realpathSync.native(value);
  } catch {
    return value;
  }
};
```

Moyasee (commented):
Findings inline. The one I would not merge past is on `custom-path-removal.ts`: untracking a custom location can throw and leave the user unable to remove it, with the location still tracked both remotely and on the device.

src/main/services/cloud-save/custom-path-removal.ts
Hachi-R: Fixed in e4393d31d. The snapshot is now deleted only when no files, variants, or declared custom paths remain. If another empty custom path is still tracked, we publish the empty updated snapshot instead. I added regressions for the two-path case and for rejecting an orphaned variant.

src/main/services/cloud-save/get-cloud-save-overview.ts
Hachi-R: Fixed in e4393d31d. The overview now treats both ready and unresolved bindings as already configured, so a recoverable unresolved location does not create a duplicate prompt or suppress automatic sync. The approval-policy regression covers both binding states.

src/main/services/cloud-save/sync-anchor.ts
Hachi-R: Kept intentionally. The cross-environment fallback is only used after the active remote snapshot disappears. Automatic deletion still requires the same composite identity plus matching hash and size; new or changed automatic files become conflicts, and custom-path entries are excluded from automatic removal. This preserves propagation of an explicit remote deletion across environment rotation without widening deletion to changed or custom saves.

src/main/services/cloud-save/sync-game-cloud-save.ts
Hachi-R: Fixed in e4393d31d. The external 409 retry and the internal state-change retry now share one attempt budget. The first state change re-runs analysis; a second consecutive change fails with cloud_save_sync_state_changed_twice instead of livelocking. The retry policy has a focused regression for attempts 0 and 1.

src/main/services/cloud-save/reconcile-remote-target-observations.ts
Hachi-R: Fixed in e4393d31d. I extracted canonicalOrLexicalPath, preserving path.resolve, realpathSync.native, and the lexical fallback while removing the explanatory catch comment. The reconciliation regressions pass.

Hachi-R:
### Addressed

- Merged the current #2645 branch through ba7b69163, preserving the #2646 reconciliation metadata and the deterministic byText ordering. This also removes the Sonar reliability bug caused by sort without a comparator.
- Empty custom-path snapshots are updated when another declared path remains; deletion only happens when files, variants, and declared paths are all empty. Orphaned variants remain rejected.
- Both ready and unresolved bindings count as configured in the overview.
- Remote-state changes and HTTP 409 conflicts now share a single retry budget; a second state change fails deterministically instead of livelocking.
- Extracted canonicalOrLexicalPath while preserving canonical and lexical path behavior.

### Intentionally unchanged

- Kept the cross-environment anchor fallback. It only applies after the active remote snapshot disappears; automatic deletion requires matching identity, hash, and size, changed/new files conflict, and custom-path entries are excluded.
- Did not refactor cognitive-complexity findings or add useMemo to the provider. Those are non-blocking smells and would widen the risk of this correctness-focused follow-up.
- Left all review threads open for reviewer confirmation.

### Validation

- Focused Node tests: 27/27 passed.
- Full Node suite: 375 passed, 4 skipped; the only failure was the known Windows EBUSY cleanup in notification-icon.test.ts.
- Rust: 159/159 passed.
- Native addon build passed.
- Node and Web typechecks passed.
- Prettier and cargo fmt checks passed.
- Targeted ESLint passed; pre-push global ESLint completed with 0 errors and 80 pre-existing warnings.
- Remote SonarCloud Code Analysis passed on e4393d31d.

src/main/services/cloud-save/custom-path-removal.ts
Moyasee: Confirmed fixed. Re-ran the original repro against the new logic:

```
original repro (two paths, one file-bearing)  ->  "snapshot-updated", deleteSnapshot not called
last tracked custom path                      ->  "snapshot-deleted"
```

Also checked the case the repro now falls through into: `validateSnapshotVariants` and `validateSnapshotFiles` both accept empty arrays, and the normal sync path can already publish a zero-file snapshot with tracked custom paths, since `hasLocalFiles` counts `customPathRawPaths`. So the empty updated snapshot is a shape the API already sees rather than a new one.

src/main/services/cloud-save/custom-path-approval.ts
```diff
@@ -115,11 +115,9 @@ const createPendingApproval = async (
       })
     ).ready.map(({ rawPath }) => rawPath)
```
Moyasee, line 116: This still passes `.ready` alone, while `get-cloud-save-overview.ts` was changed in this PR to pass `[...ready, ...unresolved]` to the same helper.

I think the difference is deliberate, since the approval prompt is exactly how a user gives an unresolvable binding a concrete folder, and folding `unresolved` in here would remove the only way out of that state. But nothing records the distinction, so the next person to touch `getUnconfiguredCloudSaveCustomPathCandidates` will reasonably conclude one of the two call sites is a bug. A line in the commit message, or a named argument at both sites, would settle it.

src/main/services/cloud-save/delete-game-cloud-save-targets.ts
```diff
@@ -0,0 +1,36 @@
...
+      ({ rawPath }) => rawPath
+    )
+  );
+  const pathApi = platform === "windows" ? path.win32 : path.posix;
+  const customRoots = [
+    ...customPathBindings.ready.map(({ path: customPath }) => customPath),
+    ...customPathBindings.unresolved.flatMap(({ pathHint }) =>
+      pathHint && pathApi.isAbsolute(pathHint) ? [pathHint] : []
+    ),
+  ];
```
Moyasee, line 27: `pathHint` goes into `isCloudSavePathWithinRoot` raw, while the other caller of that helper, `getInstallationOwnedCustomPathRawPaths`, resolves both sides through `realpath` first. The comparison itself is purely lexical, so a stored hint that is not already canonical will not match the canonical `absolutePath` the scanner produces.

The failure direction is the bad one: those files stop being recognized as custom-owned and get deleted by the flow whose own string promises "Files in custom locations were preserved and are still being tracked".

`customPathBindings.ready` entries are fine, `canonicalizeSelectedCloudSaveCustomPath` already normalized those. It is only the unresolved `pathHint` branch that skips it.

Moyasee (approved):
Approving. Reviewed the whole PR again at 0bd455a66 after the rebase.

All five findings from the first pass are resolved. The untrack hard-fail is properly fixed, not just moved: throwing now means `files === 0 && variants > 0`, which is genuinely inconsistent state, and the snapshot is deleted only once no declared custom paths remain either. Verification on that thread. The cross-environment anchor answer holds up against the code, so I am satisfied leaving it as is.

No comments added in either new commit, and the whole PR diff is comment-free.

Two small things inline, neither worth holding the merge for. Both are cases where two call sites that should agree now do not.

Sonar gate is green with five open `CODE_SMELL` issues. Four are cognitive complexity. The fifth, the unmemoized provider value at `src/big-picture/src/components/pages/game/cloud-save-v2/cloud-save-v2-context.tsx:574`, is pre-existing rather than introduced here, but worth knowing the desktop twin at `src/renderer/src/pages/game-details/cloud-save-v2/cloud-save-v2-context.tsx:804` does wrap its value in `useMemo`, so the two providers genuinely diverge.

Validation at 0bd455a66:

```
Rust:      164 passed; 0 failed
Node:      380 passed; 0 failed; 0 skipped
typecheck: node + web pass
prettier:  pass
CI:        build windows + ubuntu, lint, text-integrity, Sonar all SUCCESS
```

PR #2647, opened 2026-08-05
fix(cloud-save): polish background operations and restore resolution

Hachi-R:
## Summary

- keep background custom-path operations visible in the main modal and file browser
- show consistent operation progress while allowing non-destructive dialogs to close
- add the Cloud Save V2 feedback affordance
- validate restore-root candidates against the full manifest rule before accepting filesystem evidence

## Impact

Background work no longer appears idle after closing or reopening the file list. Restore resolution ignores incompatible legacy layouts instead of letting them redirect a valid wildcard-derived target, while preserving every existing file.

## Validation

- Rust: 162 passed
- Node: 372 passed, 4 skipped
- Node/Web typechecks: passed
- ESLint: 0 errors; 80 existing warnings
- Known non-blocking Windows failure: `EPERM` while cleaning the notification icon test fixture

## PR chain

Depends on #2646 (`codex/cloud-saves-v2-custom-path-reconciliation`).

### DIFF
--- a/src/renderer/src/pages/game-details/cloud-save-v2/cloud-save-v2-file-browser-modal.tsx
+++ b/src/renderer/src/pages/game-details/cloud-save-v2/cloud-save-v2-file-browser-modal.tsx
@@ -12,6 +12,7 @@ import { useTranslation } from "react-i18next";
 
 import type {
   CloudSaveState,
+  CloudSaveSyncProgressPayload,
   CloudSaveV2FileDetails,
   GameShop,
   SyncGameCloudSaveResult,
@@ -33,7 +34,10 @@ import {
 } from "./cloud-save-v2-file-tree";
 import { getCloudSaveFileBrowserOperationPolicy } from "./cloud-save-v2-file-browser-policy";
 import { CloudSaveV2FileTreeView } from "./cloud-save-v2-file-tree-view";
-import { hasCloudSaveDataToDelete } from "./cloud-save-presentation";
+import {
+  getCloudSaveOperationPresentation,
+  hasCloudSaveDataToDelete,
+} from "./cloud-save-presentation";
 
 const getCustomPathSelectionError = (error: unknown) => {
   const message = error instanceof Error ? error.message : "";
@@ -131,6 +135,7 @@ interface CloudSaveV2FileBrowserModalProps {
   hasError: boolean;
   isGameRunning: boolean;
   isSyncing: boolean;
+  progress: CloudSaveSyncProgressPayload | null;
   onRetry: () => void | Promise<void>;
   onSyncAfterCustomPathAdded: () => Promise<SyncGameCloudSaveResult>;
   onRemoveTrackedCustomPath: (rawPath: string) => Promise<void>;
@@ -290,6 +295,7 @@ export function CloudSaveV2FileBrowserModal({
   hasError,
   isGameRunning,
   isSyncing,
+  progress,
   onRetry,
   onSyncAfterCustomPathAdded,
   onRemoveTrackedCustomPath,
@@ -500,6 +506,14 @@ export function CloudSaveV2FileBrowserModal({
       isGameRunning,
       isSyncing,
     });
+  const activeOperation = isDeletingCloudSave
+    ? getCloudSaveOperationPresentation(null, "cloud_save_v2_deleting")
+    : isSyncing
+      ? getCloudSaveOperationPresentation(progress)
+      : null;
+  const activeOperationFileCount = activeOperation?.fileCount
+    ? t("cloud_save_v2_progress_file_count", activeOperation.fileCount)
+    : null;
   const hasSaveData = hasCloudSaveDataToDelete(details);
   const addCustomPathButton = (
     <Button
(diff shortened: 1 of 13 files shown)


Moyasee (commented):
Smallest of the three and I did not find anything blocking.

Moving `target_matches_rule` into `restore_root.rs` and gating every root candidate through `first_compatible_root` is well covered. `ignores_a_legacy_layout_when_resolving_intermediate_glob_segments`, `ignores_a_legacy_layout_inside_the_active_wine_prefix` and `resolves_each_profile_independently_when_legacy_layouts_exist` all assert the legacy file comes out byte-identical, which is the part that matters.

One thing I checked and cleared: the scan branch feeds `first_compatible_root` roots that have already been through `canonical_path`, so a symlink anywhere in the resolved path makes them fail the new lexical `root_matches_rule` check. The `complete_by_candidate` pass picks those up, so there is no behavior change on symlinked layouts.

Verified at the tip of the stack:

```
Rust:   167 passed; 0 failed
Node:   374 passed; 0 failed
typecheck:node + typecheck:web: pass
prettier --check: pass
```

If anyone runs these on macOS and sees 3 failures, they are all symlink artifacts rather than code. `custom-path.test.ts` trips `cloud_save_custom_path_protected` and `cloud_save_custom_path_symlink_ancestor` because `/tmp` resolves to `/private/tmp` and `/private` is in `protectedRoots("mac")`, and the Rust `preserves_literal_backslashes_in_filesystem_paths` trips because `canonical_path` resolves `/var/folders` to `/private/var/folders`. All pass with `TMPDIR` pointed outside `/private`. Not worth fixing given there are no macOS builds, just noting it so nobody chases it.

src/renderer/src/pages/game-details/cloud-save-v2/cloud-save-v2-file-browser-modal.tsx
```diff
@@ -479,16 +486,24 @@ export function CloudSaveV2FileBrowserModal({
...
+      isDeletingCloudSave,
+      isLoading,
+      isGameRunning,
+      isSyncing,
+    });
+  const activeOperation = isDeletingCloudSave
+    ? getCloudSaveOperationPresentation(null, "cloud_save_v2_deleting")
+    : isSyncing
+      ? getCloudSaveOperationPresentation(progress)
+      : null;
```
Moyasee, line 513: Sonar flags this one and the JSX ternary further down, and both are new in this PR. Not blocking, but it is the same rule you already cleared everywhere else in the stack: `planner.ts` and `automatic-sync.ts` in #2645, `analyze-cloud-save-state.ts` and `errorMessageKey` in #2646. Leaving these two means the stack lands with the rule enforced in four places and skipped in two.

Same shape as the fix you used before:

```ts
let activeOperation: CloudSaveOperationPresentation | null = null;
if (isDeletingCloudSave) {
  activeOperation = getCloudSaveOperationPresentation(null, "cloud_save_v2_deleting");
} else if (isSyncing) {
  activeOperation = getCloudSaveOperationPresentation(progress);
}
```

src/renderer/src/pages/game-details/cloud-save-v2/cloud-save-v2-file-browser-modal.tsx
```diff
@@ -556,9 +571,34 @@ export function CloudSaveV2FileBrowserModal({
...
+                          <span aria-hidden="true">·</span>
+                        )}
+                        {activeOperationFileCount && (
+                          <span className="cloud-save-v2__browser-operation-count">
+                            {activeOperationFileCount}
+                          </span>
+                        )}
+                      </span>
+                    </div>
+                  ) : !isConflict ? (
```
Moyasee, line 611: The other one. `activeOperation ? A : !isConflict ? B : null` is exactly the pattern the #2646 Sonar pass rewrote in `cloud-save-modal.tsx`, where `!hasExecutablePath ? x : mode === "skeleton" ? y : ...` became separate guarded blocks. Two `&&` blocks do the same job here:

```tsx
{activeOperation && ( ... )}
{!activeOperation && !isConflict && ( ... )}
```

Moyasee (approved):
Approving. Reviewed at c047898eb.

The rebase did not change the delta. I diffed the pre-rebase content against the current one and the only differences are hunk line offsets, so the review I did earlier still applies as written and I did not re-derive it from scratch.

The substance holds up. Moving `target_matches_rule` into `restore_root.rs` and routing every root candidate through `first_compatible_root` is well covered, and the three tests that matter all assert the legacy file comes out byte-identical rather than just checking the resolved target: `ignores_a_legacy_layout_when_resolving_intermediate_glob_segments`, `ignores_a_legacy_layout_inside_the_active_wine_prefix`, `resolves_each_profile_independently_when_legacy_layouts_exist`.

One thing I chased and cleared, recording it so nobody re-chases it: the scan branch feeds `first_compatible_root` roots that have already been through `canonical_path`, so a symlink anywhere in the resolved path makes them fail the new lexical `root_matches_rule` check. The `complete_by_candidate` pass picks those up, so symlinked layouts behave the same as before.

No comments added anywhere in the diff. One consistency note inline, not blocking.

Validation at c047898eb:

```
Rust:      167 passed; 0 failed
Node:      385 passed; 0 failed; 0 skipped
typecheck: node + web pass
prettier:  pass
CI:        build windows + ubuntu, lint, text-integrity, Sonar all SUCCESS
```

GitHub reports the branch as BEHIND, though the base tip is an ancestor of this head locally, so that looks stale rather than a real rebase requirement. Worth a glance before merging.

PR #2649, opened 2026-08-06
fix(i18n): enable Japanese and Slovenian translations

Cryptorial0:

**Fill in the PR content:**

Japanese (`ja`) and Slovenian (`sl`) translation files were present in `src/locales/`, but they weren't exported in `src/locales/index.ts`.

This PR imports and exports both locales so they show up in the language settings dropdown and work with OS language detection.


src/locales/index.ts
```diff
@@ -65,5 +67,8 @@ export default {
   uz,
   sv,
   lv,
+  ja,
+  sl,
```
Moyasee: `sl` should not be enabled yet. Slovenian has four plural categories and the file only supplies two, so counts 2 and 3 to 4 fall out of Slovenian entirely:

```
sl plural categories: [ 'one', 'two', 'few', 'other' ]
sl file supplies:     _zero, _one, _other   (all 8 plural bases, none have _two or _few)

count   1 | cat one   | "1 možnost prenosa"
count   2 | cat two   | "2 download options"
count   3 | cat few   | "3 download options"
count   5 | cat other | "5 možnosti prenosa"
```

`_two` and `_few` miss in `sl`, fall through to `en`, and `en` has no `_two`/`_few` either, so they land on the English `_other`. The same label flips language depending on the number. That is worse than the locale simply being absent.

The 8 bases that need `_two` and `_few` added:

```
game_details.download_options
settings.download_count
settings.found_download_option
settings.game_card.available
notifications.repack_count
settings.notification_messages.repack_count
user_profile.report_reason
settings.user_profile.report_reason
```

Separately, `sl` has 11 sections misnested under `settings` (`user_profile`, `library`, `modal`, `forms`, `system_tray`, `game_card`, `achievement`, `notifications_page`, `binary_not_found_modal`, `notification_messages`, `hydra_cloud_info`). 186 of those keys are dead duplicates of the correct top-level ones, so they are mostly harmless, but 2 are only in the nested copy and never resolve: `game_card.available_one` and `game_card.available_other`. Worth cleaning up in the same pass.

src/locales/index.ts
```diff
@@ -65,5 +67,8 @@ export default {
   uz,
   sv,
   lv,
+  ja,
```
Moyasee, line 70: `ja` is in good shape, no plural problem (Japanese is `other`-only) and no structural issues in the file.

It does need one more line though, in `src/shared/language-flags.ts`:

```
ja: "JP",
sl: "SI",
```

Without it `getLanguageFlagCountryCode` returns null and the Big Picture language picker falls back to the globe icon instead of a flag (`src/big-picture/src/pages/settings/language-flag.tsx:24`). Nothing crashes and the desktop dropdown only uses `nativeName` so it is unaffected, but every other enabled locale has an entry there.

Optionally `getDateLocale` in `src/shared/index.ts` too, though `lv` does not have one either so that is consistent with what is already merged.

Moyasee (commented):
The premise checks out. Both directories are on `main` and neither is imported here, so they never reach i18next's `resources` and cannot show up in either language picker or be resolved by `LanguageDetector`.

One thing worth getting ahead of, since it is the obvious reason to reject this: neither locale is anywhere near complete, but that is not a reason to hold it.

```
ja  1057 / 1648
sl  1247 / 1648
lv  1033 / 1648   <- already enabled
sv  1033 / 1648   <- already enabled
```

Both are more complete than two locales already in the dropdown, and anything missing falls back to English through `fallbackLng: "en"`. If the coverage here disqualifies them it disqualifies what already ships.

My suggestion is to split this. `ja` is basically ready and only needs one more line. `sl` has a real bug that this PR would turn on. Details inline.

Cryptorial0:
Applied the feedback (removed sl, fully registered ja). Just a heads up, the CI failed due to the current GitHub Actions outage, but the local typechecks and linting pass fine.

src/main/services/redist-manager.ts
```diff
@@ -0,0 +1,159 @@
...
+
+export interface RedistDownloadProgress {
+  packageName: string;
+  bytesDownloaded: number;
+  totalBytes: number;
+  percentage: number;
+  speed: number;
+}
+
+const REDIST_BASE_URL =
```
Moyasee: When this moves to its own PR, worth reconciling with what already exists. `main` has `src/main/services/common-redist-manager.ts` doing the same job against the same repo:

```
CommonRedistManager.downloadCommonRedist()   ->  https://github.com/hydralauncher/hydra-common-redist/raw/refs/heads/main/${redist}
CommonRedistManager.installCommonRedist()
CommonRedistManager.runPreflight()
```

`RedistManager` here is a second implementation with its own `https.get`, its own cache directory and its own install path pointed at the same `hydra-common-redist` repo. Two downloaders and two install paths for the same packages will drift. Either extend `CommonRedistManager` with the on-demand case or fold the shared download and install logic into one place.

src/locales/en/translation.json
```diff
@@ -529,6 +529,16 @@
     "winetricks_not_available_tooltip": "Winetricks is not available in your PATH",
     "winetricks_opened": "Winetricks opened",
     "winetricks_open_error": "Failed to open Winetricks",
+    "missing_redist_modal_title": "Missing Runtime Dependency Detected",
```
Moyasee: These 10 keys are English only. `ru`, `es` and `pt-BR` all return undefined for them, so the modal shows English inside a translated UI.

```
$ node -e "..." src/locales/{ru,es,pt-BR}/translation.json
ru     game_details.missing_redist_modal_title: undefined
es     game_details.missing_redist_modal_title: undefined
pt-BR  game_details.missing_redist_modal_title: undefined
```

Whichever PR ends up carrying the feature should add them to those three alongside `en`.

Moyasee (changes requested):
The locale part is right now. `sl` is out, `ja: "JP"` is in the flag map, and you added the `getDateLocale` branch on top, which I had only mentioned as optional. Typecheck, lint and prettier are all clean.

The problem is what else came with it. The commit is titled "address PR review feedback (remove sl, complete ja registration)" but it carries 739 lines of an unrelated feature: missing-DLL detection on Wine/Proton launch failure and auto-installing redistributables. 15 of the 17 files in this PR have nothing to do with locales.

```
src/main/services/redist-detector.ts               +142  (new)
src/main/services/redist-manager.ts                +159  (new)
src/main/events/library/install-game-redist.ts      +70  (new)
src/renderer/.../missing-redist-modal.tsx          +202  (new)
src/renderer/.../missing-redist-modal.scss          +74  (new)
src/main/helpers/launch-game.ts                     +17
src/preload/index.ts                                +15
src/renderer/src/app.tsx                            +24
src/main/services/umu.ts                            +12
src/locales/en/translation.json                     +10
```

That needs to come out and go in its own PR. A four-line locale registration gets reviewed as a four-line locale registration, and a launch-path change plus a new IPC channel plus a downloader would ride in without anyone looking at it. Two more things about it inline, since they will matter when it gets its own PR.

Once this is back to just the locale changes it is good to merge.

Moyasee (commented):
The extra files are gone and this is back to what it should be: three files, +5 -0, the `ja` registration in `src/locales/index.ts`, the `ja: "JP"` flag entry and the `getDateLocale` branch. Nothing else. Typecheck, lint, prettier and the CR check are all clean, and the `ja` file itself holds up: `other`-only plural category so no plural risk, no misnesting, and the one stale-looking string (`catalogue.clear_filters` carrying `{{filterCount}}` where `en` has none) renders fine because every call site still passes that variable.

One thing left, and it is not code. The title and body still say Japanese and Slovenian:

> fix(i18n): enable Japanese and Slovenian translations

Slovenian is not in this anymore. The title is what release note generation reads, so as written it would ship a changelog line claiming a locale that was not enabled. Could you retitle to just Japanese and drop the `sl` sentence from the body? I did not want to edit your PR out from under you.

Once that is updated I will clear my change request and this is good to merge.

Moyasee:
To be clear, retitling is only one way out. The other is to put `sl` back and fix the plurals, and then the title is accurate as it stands.

That is less work than my earlier comment made it sound. Only 4 keys are actually half-present, and those are the ones causing the flip:

```
game_details.download_options     has _zero,_one,_other   needs _two,_few
settings.download_count           has _zero,_one,_other   needs _two,_few
settings.found_download_option    has _zero,_one,_other   needs _two,_few
notifications.repack_count        has _one,_other         needs _two,_few
```

The other 14 plural bases in `en` are missing from `sl` entirely, so every count falls back to English together. That is just the normal untranslated state and it reads fine. It is only these 4, where `_one` and `_other` exist but `_two` and `_few` do not, that produce Slovenian at 1 and 5 and English at 2 and 3.

Ignore `user_profile.report_reason_other` if you go looking, by the way. It ends in `_other` but it is a literal key, the report reason "Other", called as ``t(`report_reason_${reason}`)``. Not a plural, leave it alone.

If you take that route it also wants `sl: "SI"` in `LANGUAGE_FLAG_MAP` alongside the `ja` one.

Worth a look while you are in there: `sl` has 11 sections nested under `settings` that should be top level (`user_profile`, `library`, `modal`, `forms`, `system_tray`, `game_card`, `achievement`, `notifications_page`, `binary_not_found_modal`, `notification_messages`, `hydra_cloud_info`). 186 of those keys are dead duplicates of the correct top-level ones so nothing breaks, but `game_card.available_one` and `game_card.available_other` only exist in the nested copy and never resolve. Not a blocker either way.

Either option is fine by me. Japanese-only with the title fixed is the quicker one if you would rather not touch the Slovenian file right now.

src/locales/sl/translation.json
```diff
@@ -1025,15 +793,21 @@
     "scan_games_complete_title": "Scanning for games finished successfully",
     "scan_games_complete_description": "Found {{count}} games without executable path set",
     "scan_games_no_results_title": "Scanning for games finished",
-    "scan_games_no_results_description": "No installed games were found"
+    "scan_games_no_results_description": "No installed games were found",
+    "repack_count_two": "{{count}} repacka dodana",
+    "repack_count_few": "{{count}} repacki dodani"
```
Moyasee, line 807: This brings the flip back, just inverted. `_two` and `_few` are Slovenian now but `_one` and `_other` are still the untranslated English strings at lines 780 and 781:

```
count 1 | one   | "1 repack added"
count 2 | two   | "2 repacka dodana"
count 3 | few   | "3 repacki dodani"
count 5 | other | "5 repacks added"
```

The Slovenian for the other two forms was already in the file, in the nested copy this commit deleted:

```
-      "repack_count_one": "{{count}} repack dodan",
-      "repack_count_other": "{{count}} repackov dodanih",
```

So it just needs moving up rather than rewriting:

```
"repack_count_one": "{{count}} repack dodan",
"repack_count_two": "{{count}} repacka dodana",
"repack_count_few": "{{count}} repacki dodani",
"repack_count_other": "{{count}} repackov dodanih",
```

src/locales/sl/translation.json
```diff
@@ -1170,7 +944,9 @@
     "user_reviews": "Reviews",
     "delete_review": "Delete Review",
     "loading_reviews": "Loading reviews...",
-    "wrapped_2025": "Wrapped 2025"
+    "wrapped_2025": "Wrapped 2025",
+    "report_reason_two": "Drugo",
+    "report_reason_few": "Drugo"
```
Moyasee: These two never resolve. `report_reason_other` is a literal key, the reason "Other", and the call site is:

```
report-profile.tsx:125   label: t(`report_reason_${reason}`)
```

where `reason` is one of `hate`, `sexual_content`, `violence`, `spam`, `other`. Nothing ever asks for `report_reason_two` or `report_reason_few`, so "Drugo" is parked where it cannot be reached.

Meanwhile line 901 still reads:

```
"report_reason_other": "Other",
```

That is the untranslated top-level copy, and "Drugo" came from the nested one this commit deleted. Drop these two keys and put the translation where it is actually looked up:

```
"report_reason_other": "Drugo",
```

While you are there, `report_reason_hate`, `report_reason_sexual_content`, `report_reason_violence` and `report_reason_label` are all still English at lines 897 to 901.

src/locales/sl/translation.json
```diff
@@ -1003,7 +767,11 @@
     "proton_version_auto": "Auto (global default or umu default)",
     "proton_source_umu_default": "umu default selection",
     "proton_source_steam": "Installed by Steam",
-    "proton_source_compatibility_tools": "Installed in Steam compatibilitytools.d"
+    "proton_source_compatibility_tools": "Installed in Steam compatibilitytools.d",
+    "download_count_two": "{{countFormatted}} možnosti prenosa",
+    "download_count_few": "{{countFormatted}} možnosti prenosa",
+    "found_download_option_two": "Najdeni {{countFormatted}} možnosti prenosa",
+    "found_download_option_few": "Najdene {{countFormatted}} možnosti prenosa"
```
Moyasee, line 783: Worth a second look from a native speaker, these two disagree on the adjective ending for the same noun:

```
count 2 | "Najdeni 2 možnosti prenosa"
count 3 | "Najdene 3 možnosti prenosa"
```

`_one` uses "Najdena" and `_other` uses "Najdenih", so the pattern is clearly deliberate elsewhere. I cannot judge which of "Najdeni" or "Najdene" is right here, just flagging that the dual and the few forms differ where the noun does not.

Moyasee (changes requested):
Good call taking this route, the title is accurate again. The plural work itself is right, all four bases are complete now and resolve correctly across the whole range:

```
game_details.download_options   1 "1 možnost prenosa"  2 "2 možnosti prenosa"  3 "3 možnosti prenosa"  5 "5 možnosti prenosa"
settings.found_download_option  1 "Najdena 1..."       2 "Najdeni 2..."        3 "Najdene 3..."        5 "Najdenih 5..."
```

The misnesting is fully gone (`settings` has no sub-objects left) and the file dropped from 1247 to 1045 keys as the 186 duplicates came out. Flag map and `getDateLocale` both have `ja` and `sl`. Typecheck, lint, prettier and the CR check are clean.

The problem is the direction the de-nesting went. For three key groups the nested copy held the Slovenian and the top-level copy held untranslated English, and the cleanup kept the English one and deleted the Slovenian. `game_card.available` got promoted correctly. The other two did not, and one of them reintroduces the exact flip this PR set out to fix. Both inline.

Also worth a rebase before merge, the branch is behind `main`.

src/locales/sl/translation.json
```diff
@@ -771,212 +741,6 @@
     "autoplay_trailers_on_game_page": "Samodejno predvajaj napovednike na strani igre",
```
Moyasee, line 741: Deleting these is right, they were unreachable. But most of them should move up rather than go away, and I undercounted badly when I called it three key groups earlier. It is 188.

The pattern is the one from the two threads you already fixed: a translated string sits at `settings.<namespace>.<key>` where nothing can reach it, while the live `<namespace>.<key>` still holds the English. That is not a fallback, the Slovenian file literally stores the English string, so a Slovenian user sees English.

```
216 nested leaves deleted here
   20  live key already has the same value      -> fine to drop
  188  live key still holds the English string  -> the loss
    7  live key already has different Slovenian -> fine to drop
    1  no matching key in en                    -> fine to drop

by namespace: user_profile 94, notifications_page 33, library 16,
              achievement 12, notifications 13, hydra_cloud 11, rest 9
```

So `"Prikazno ime"` goes out while `user_profile.display_name` keeps rendering "Display name", and the same for `library.play` and `notifications_page.title`.

Nothing regresses either way since the nested paths were dead, but promoting them takes `sl` from 628/1881 to 816/1881, 33.4% to 43.4%. Placeholders match on all 188 so it is mechanical.

Two of the eleven are named by meaning rather than by namespace and need mapping rather than a straight move: `notification_messages` goes to `notifications`, `hydra_cloud_info` goes to `hydra_cloud`.

One group to skip: `binary_not_found_modal`, the three keys there are semantically stale. The Slovenian still talks about Wine and Lutris where the current `en` says umu-run.

Moyasee (commented):
My changes-requested is stale, sorry about that. Both plural findings are fixed at head and all four forms resolve now. The "Najdeni 2 / Najdene 3" one turned out to be my mistake: `-i` is the nominative dual and `-e` the nominative plural for feminine adjectives, so the file was already right. Resolving those three threads.

One thing left, inline. Once that is in we can approve and merge this.

src/locales/sl/translation.json
Moyasee: Fixed, all four forms resolve now.

src/locales/sl/translation.json
Moyasee: Fixed, and the four sibling keys got translated too.

src/locales/sl/translation.json
Moyasee: My mistake, this is correct as written. Feminine adjectives take `-i` in the nominative dual and `-e` in the nominative plural, so "Najdeni 2" and "Najdene 3" is the right paradigm alongside "Najdena 1" and the genitive "Najdenih 5". Nothing to change.

src/locales/sl/translation.json
Moyasee: All 188 promoted, and you correctly left the three stale ones alone. `settings` has no sub-objects left and sl is at 43.2%.

Moyasee (approved):
LGTM!

PR #2652, opened 2026-08-06
feat(cloud-save): add legacy save management

Hachi-R:
## Summary

- keep Cloud Saves V1 available for emulated games while Steam games use Cloud Saves V2
- add a subscriber-only Legacy Saves archive for Steam games with existing V1 artifacts
- allow legacy artifacts to be exported as ZIP files and permanently deleted after confirmation
- show local and remote snapshot metadata during conflicts on desktop and Big Picture
- expose the shared Cloud Saves V2 panel in Steam game settings on Big Picture and align controller focus behavior

## Details

Legacy saves are listed newest first with their label, size, creation date, hostname, and download option when available. Exporting downloads the existing V1 artifact, converts it from TAR to ZIP, opens the native save dialog, and cleans temporary files on success, cancellation, or failure. Deletion continues to use the existing artifact endpoint and refreshes the list after the server confirms the operation.

The visibility policy is now explicit: Steam uses V2 and may expose the legacy archive on desktop, Launchbox/emulated games keep the active V1 experience, and Big Picture does not expose the legacy archive.

## Validation

- 46 targeted cloud-save, legacy-export, visibility, presentation, and Big Picture tests passed
- Node and Web typechecks passed
- pre-push ESLint completed with 0 errors (existing warnings remain)
- Prettier and `git diff --check` passed

## Stack

This PR is based on #2647 (`codex/cloud-saves-v2-steam-multi-account-restore`).


src/locales/en/translation.json
```diff
@@ -744,6 +742,14 @@
     "settings_category_customization": "Customization",
     "settings_category_launch": "Launch",
     "settings_category_hydra_cloud": "Hydra Cloud",
+    "settings_category_legacy_saves": "Legacy Saves",
```
Moyasee, line 744: The eight new `legacy_save*` / `settings_category_legacy_saves` keys only landed in `en` and `pt-BR`. `ru` and `es` don't have them, so those users get the English strings in the Legacy Saves panel, the delete confirmation, and the download toasts. Can you add them in the same commit? Same key path and nesting, `{{name}}` left as is.

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
```diff
@@ -844,9 +879,23 @@ export function GameOptionsModal({
     }
 
     setSelectedCategory(category);
-  }, [cloudSaveAccessAction, initialCategory, showHydraCloudModal, visible]);
+  }, [
+    cloudSaveAccessAction,
+    initialCategory,
+    showCloudSaveV2Settings,
+    showLegacyCloudSaveSettings,
```
Moyasee, line 897: Adding `showLegacyCloudSaveSettings` to these deps couples the effect to async state: it is derived from `artifacts.length`, and `artifacts` starts as `[]` and is filled in by the `/profile/games/artifacts` request. When that resolves for a Steam game with legacy artifacts the flag flips false -> true, the effect re-runs, and the last line unconditionally calls `setSelectedCategory(initialCategory ?? "general")`, throwing away whatever category the user picked in the meantime.

Precondition is opening the options modal while that request is still in flight, so the window is short but it is a plain round trip, and any later refetch that changes the count does the same thing.

The reset only needs to happen when the modal opens or `initialCategory` changes, plus when the currently selected category becomes unavailable. Splitting those apart, so the availability flags only bounce you out of a category you are actually sitting on, avoids re-applying `initialCategory` on every recompute.

Moyasee (commented):
The export path is nicely structured: pulling the ordered steps into `exportGameArtifactArchive` with injected dependencies makes the cancel/failure cleanup easy to follow, and the `finally` covers both.

Two things to fix before this goes in, both inline. Once these are addressed we can approve and merge.

src/big-picture/src/components/pages/game/cloud-save-v2/cloud-save-modal.tsx
```diff
@@ -159,206 +176,251 @@ export function BigPictureCloudSaveModal({
...
+      <section className="big-picture-cloud-save__toggle-card">
+        <div className="big-picture-cloud-save__copy">
+          <strong>
+            {t("cloud_save_v2_toggle_title", {
               status: t(
                 isAutomaticSyncEnabled
                   ? "cloud_save_v2_toggle_enabled"
                   : "cloud_save_v2_toggle_disabled"
               ),
             })}
```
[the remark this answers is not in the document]
Hachi-R: Addressed in 7fde1d227. I replaced the literal icon sizes with semantic constants (TOGGLE_ICON_SIZE, LOADING_ICON_SIZE, CLOUD_ACTION_ICON_SIZE, INLINE_STATUS_ICON_SIZE, NOTICE_ICON_SIZE, and SNAPSHOT_METADATA_ICON_SIZE) while preserving the existing visual values.

src/locales/en/translation.json
Hachi-R: Addressed in 7fde1d227. I added all eight Legacy Saves keys to both es and ru under the same game_details nesting, with localized copy and the {{name}} placeholder preserved in the deletion description.

src/renderer/src/pages/game-details/modals/game-options-modal.tsx
Hachi-R: Addressed in 7fde1d227. Category initialization is now limited to modal opens and initialCategory changes, while a separate availability effect only falls back when the currently selected category becomes unavailable. Artifact refreshes no longer reapply initialCategory, and paywall behavior remains tied to explicit access attempts. I also added six regression tests covering modal opening, category changes, selection preservation, and availability/access fallback. Validation completed with 45 targeted tests, Node and Web typechecks, and pre-push ESLint with 0 errors.

src/renderer/src/pages/game-details/modals/game-options-modal/legacy-save-card.tsx
```diff
@@ -0,0 +1,85 @@
...
+              {t("download")}
+            </>
+          )}
+        </Button>
+
+        <Button
+          type="button"
+          theme="danger"
+          className="legacy-saves-section__delete-button"
+          aria-label={t("delete_backup")}
```
Moyasee, line 76: Delete is enabled for every artifact here, frozen ones included. The legacy panel disables it (`cloud-sync-panel.tsx:352`, `disabled: disableActions || artifact.isFrozen`), so a pinned backup that can't be deleted there can be deleted from the archive, and the confirmation says it's permanent. For Steam games the archive is now the only surface these artifacts have, so it's the one place freezing should still hold.

`disabled={actionsDisabled || artifact.isFrozen}` on the delete button plus a frozen marker on the card covers it. `sortLegacySavesByNewest` also drops the frozen-first ordering the panel uses (`orderBy(artifacts, [(a) => !a.isFrozen], ["asc"])`), worth keeping so pinned saves stay at the top.

src/renderer/src/pages/game-details/cloud-sync/cloud-sync-panel.tsx
```diff
@@ -209,7 +209,7 @@ export function CloudSyncPanel({
       />
 
       <div className="cloud-sync-panel__section-header">
-        <h2>{t("cloud_save_legacy_title")}</h2>
+        <h2>{t("settings_category_hydra_cloud")}</h2>
```
Moyasee, line 212: With the heading now reading "Hydra Cloud", `cloud_save_legacy_description` ends on "Automatic synchronization is available in Hydra Cloud." and the automatic sync toggle is right below it. Same pairing in `src/big-picture/src/components/pages/game/game-settings-modal/cloud-tab.tsx:531`. Dropping that second sentence in the four locales fixes both spots.

`settings_category_hydra_cloud_legacy` and `cloud_save_legacy_title` have no code references left after the rename either, so they can come out of en/es/ru/pt-BR in the same pass.

Moyasee (changes requested):
Both earlier findings are fixed. The eight legacy keys are in all four locales now, and splitting the initialization effect from the availability clamp means an artifact refresh no longer throws away the category the user picked.

One thing to fix before merge, plus a copy nit that came in with the Hydra Cloud rename. Both inline.

src/renderer/src/pages/game-details/modals/game-options-modal/legacy-save-card.tsx
Hachi-R: Thanks for flagging this. Keeping deletion enabled for frozen artifacts in this archive is intentional. Frozen/pinned protects artifacts in the active backup workflow, while Legacy Saves is a retirement surface for old artifacts that users need to export or permanently remove after explicit confirmation. For the same reason, this view keeps its current newest-first ordering instead of adopting the active panel frozen-first presentation. No code change is planned for this thread.

src/renderer/src/pages/game-details/cloud-sync/cloud-sync-panel.tsx
Hachi-R: Addressed in e87e876b7. I removed the redundant automatic-sync sentence from cloud_save_legacy_description and deleted the unused settings_category_hydra_cloud_legacy and cloud_save_legacy_title keys in en, es, ru, and pt-BR. Validation passed for JSON parsing, exact copy and removed-key assertions, TypeScript/TSX reference checks, Prettier, and git diff --check. The pre-push Node/Web typechecks also passed, and ESLint completed with 0 errors.

src/renderer/src/pages/game-details/modals/game-options-modal/legacy-save-card.tsx
Moyasee: Fair enough, retiring the artifacts makes the freeze flag moot for this surface. Dropping it.

Moyasee (approved):
LGTM!

PR #2665, opened 2026-08-07
fix: keep the classics catalogue on PlayStation platforms

Moyasee:

**Fill in the PR content:**

Backport of #2660 to release/v4.0.1.

The Nintendo platforms are about to be imported into the prod catalogue. 4.0.x has no RetroArch, so those games would show up under Classics for people who have no way to run them, and the client cannot opt out on its own: the classics search sends `platforms: []`, which the API reads as every platform.

This pins the classics search to ps1, ps2 and ps3 and drops everything else from the platform filter list, on desktop and in Big Picture. The renderer is served per release branch from Cloudflare Pages, so merging into release/v4.0.1 reaches already-installed 4.0.1 clients on their next launch.

Verified against staging, which already has the new platforms imported. Searching "mario" with `platforms: []` returns count 70, all six Nintendo platforms. The same search with `platforms: ["ps1","ps2","ps3"]` returns count 4, PlayStation only, so the filtering happens server side and paging stays consistent.

One gap this does not close: `/catalogue/search/suggestions` returns no platform field, so Nintendo titles can still appear in the search box dropdown. Hiding those needs an API change.

Targets release/v4.0.1, not main. Main should keep showing whatever the API exposes.

The shared constants hunk sits in a different spot on this branch because `getGameExecutableFilters` landed later. The added line is identical to the other branches.


Hachi-R (changes requested):
[P2] Keep the Big Picture URL state aligned with the platform allowlist. The request filters values.platforms only while building the API payload, but the raw URL values still drive the search key, active-filter chips, and filter modal. Opening a Classics URL with an unsupported platform such as switch therefore displays that platform as active while silently querying all supported PlayStation platforms, and the modal cannot represent the stale selection. Sanitize the platform values when parsing the URL, or reconcile the URL before exposing values, so the UI, search key, and request payload agree. The core PS1-PS3 restriction otherwise looks good.

chubbygrannychaser (changes requested):
## Review

Requesting changes for a Classics URL-state regression. An unsupported platform retained in an existing Classics URL remains represented by the filter UI and search key, but the request drops it and falls back to all supported PlayStation platforms. Please normalize the Classics platform state/URL before rendering and searching, then add regression coverage.

Moyasee:
Sanitized at the parse step instead: `values.platforms` now drops anything outside the allowlist as it comes off the URL, so the search key, the chips and the filter modal read the same list the request sends. The payload no longer filters separately, it just falls back to the full allowlist when the selection is empty.

Did the same to the desktop `classicsPlatforms` memo, which had the same split between what the chips showed and what was requested.

Moyasee:
Both parts are in now.

Normalization: the platform values are sanitized the moment they come off the URL in Big Picture, so the search key, the chips, the filter modal and the request all read the same list. Desktop had the same split through redux rather than the URL and got the same treatment. A Classics URL carrying `switch` now shows no chip and queries PlayStation only.

Regression coverage: the rule was written out in four places, so it moved into `src/shared/classics-platforms.ts` behind `sanitizeClassicsPlatforms` and `resolveClassicsPlatformsForRequest`, with `src/shared/classics-platforms.test.ts` covering it. The case worth guarding is an entirely unsupported selection resolving to the allowlist rather than to an empty array, which the API reads as every platform.

```
▶ sanitizeClassicsPlatforms
  ✔ keeps the supported platforms
  ✔ drops platforms this client cannot run
  ✔ returns an empty list for missing values
▶ resolveClassicsPlatformsForRequest
  ✔ never asks for every platform when nothing is selected
  ✔ falls back to the allowlist when the selection is entirely unsupported
  ✔ keeps a supported subset
  ✔ does not hand out the shared allowlist array
```

Your inline note points at the desktop request builder while the URL path you describe is Big Picture. Both are covered.

chubbygrannychaser (commented):
## Re-review

The stale Classics URL-state issue is resolved with shared normalization helpers and focused regression tests. Required CI is still running, so I am leaving a current-head comment rather than approving prematurely.

Hachi-R (approved):
Re-reviewed the current head. The URL and filter-state mismatch is resolved: unsupported platforms are sanitized at the state boundary, desktop and Big Picture now share the same request normalization, and focused regression tests cover empty, unsupported, and mixed selections. I found no remaining code issues in this backport.

PR #2667, opened 2026-08-07
feat(cloud-save): implement legacy save export with progress tracking…

Hachi-R:
… and cancellation

**Fill in the PR content:**

### DIFF
--- a/src/main/events/cloud-save/export-game-artifact-operation.ts
+++ b/src/main/events/cloud-save/export-game-artifact-operation.ts
@@ -2,6 +2,7 @@ import type { LegacySaveExportResult } from "@types";
 import path from "node:path";
 
 export interface ExportGameArtifactDependencies {
+  signal?: AbortSignal;
   createTemporaryDirectory: () => Promise<string>;
   downloadTar: (destinationPath: string) => Promise<void>;
   extractTar: (tarPath: string, destinationPath: string) => Promise<void>;
@@ -31,20 +32,27 @@ export const sanitizeLegacySaveArchiveName = (value: string): string => {
 export const exportGameArtifactArchive = async (
   dependencies: ExportGameArtifactDependencies
 ): Promise<LegacySaveExportResult> => {
+  dependencies.signal?.throwIfAborted();
   const temporaryDirectory = await dependencies.createTemporaryDirectory();
   const tarPath = path.join(temporaryDirectory, "artifact.tar");
   const extractedPath = path.join(temporaryDirectory, "extracted");
   const zipPath = path.join(temporaryDirectory, "artifact.zip");
 
   try {
+    dependencies.signal?.throwIfAborted();
     await dependencies.downloadTar(tarPath);
+    dependencies.signal?.throwIfAborted();
     await dependencies.extractTar(tarPath, extractedPath);
+    dependencies.signal?.throwIfAborted();
     await dependencies.createZip(extractedPath, zipPath);
+    dependencies.signal?.throwIfAborted();
 
     const destinationPath = await dependencies.selectDestination();
     if (!destinationPath) return { status: "cancelled" };
 
+    dependencies.signal?.throwIfAborted();
     await dependencies.copyZip(zipPath, destinationPath);
+    dependencies.signal?.throwIfAborted();
     return { status: "saved", filePath: destinationPath };
   } finally {
     await dependencies.cleanupTemporaryDirectory(temporaryDirectory);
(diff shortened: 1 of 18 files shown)


src/main/events/cloud-save/export-game-artifact-operation.ts
```diff
@@ -31,19 +32,25 @@ export const sanitizeLegacySaveArchiveName = (value: string): string => {
...
+    dependencies.signal?.throwIfAborted();
     await dependencies.createZip(extractedPath, zipPath);
+    dependencies.signal?.throwIfAborted();
 
     const destinationPath = await dependencies.selectDestination();
     if (!destinationPath) return { status: "cancelled" };
 
+    dependencies.signal?.throwIfAborted();
     await dependencies.copyZip(zipPath, destinationPath);
     return { status: "saved", filePath: destinationPath };
```
Moyasee, line 55: This `throwIfAborted` runs after `copyZip` has already written the ZIP to the path the user picked. A cancel landing in that window throws, `exportGameArtifact` maps it to `cancelled`, and the user ends up with a complete file on disk and no success toast. The window is small for a local copy but not for a few GB going to an external drive.

Repro against this file:

```
  file written to: C:\Users\me\Downloads\save.zip
  returned result: null
  threw: AbortError - This operation was aborted
```

(`copyZip` aborts the controller right after recording the destination, everything else is the normal happy path.)

Dropping the line is enough. The check before `copyZip` already covers cancellation arriving early enough to matter.

src/main/events/cloud-save/game-artifact-export-coordinator.ts
```diff
@@ -0,0 +1,34 @@
...
+  start(senderId: number): AbortController | null {
+    if (this.activeExport) return null;
+
+    const controller = new AbortController();
+    this.activeExport = { senderId, controller };
+    return controller;
+  }
+
+  cancel(senderId: number): boolean {
+    if (!this.activeExport || this.activeExport.senderId !== senderId) {
```
Moyasee: Sonar flags this as `typescript:S6582`:

```ts
if (this.activeExport?.senderId !== senderId) return false;
```

Same behaviour, since with no active export the optional chain gives `undefined`, which never matches a sender id.

src/renderer/src/pages/game-details/modals/game-options-modal/legacy-save-card.tsx
```diff
@@ -28,6 +30,13 @@ export function LegacySaveCard({
     t("backup_from", {
       date: formatDate(artifact.createdAt),
     });
+  const downloadProgressLabel =
+    downloadProgress?.percentage !== null &&
+    downloadProgress?.percentage !== undefined
+      ? `${downloadProgress.percentage}%`
+      : downloadProgress
+        ? formatBytes(downloadProgress.downloadedBytes)
+        : "0%";
```
Moyasee: Nested ternary, and Sonar flags it as MAJOR (`typescript:S3358`). The same rule has already been enforced elsewhere on this stack, so worth pulling out:

```ts
const formatDownloadProgress = (progress: LegacySaveExportProgress | null) => {
  if (!progress) return "0%";
  if (progress.percentage === null) return formatBytes(progress.downloadedBytes);
  return `${progress.percentage}%`;
};
```

That also drops the `!== undefined` guard, which only exists because the optional chaining was doing double duty for "no progress yet" and "unknown total".

src/renderer/src/pages/game-details/modals/game-options-modal/legacy-save-card.tsx
```diff
@@ -59,7 +68,7 @@ export function LegacySaveCard({
                 className="legacy-saves-section__spinner"
                 size={16}
               />
-              {t("legacy_save_downloading")}
+              {downloadProgressLabel}
```
Moyasee, line 76: Replacing the label with a bare percentage orphans `legacy_save_downloading`. Nothing references it anymore, but it is still sitting in en, es, pt-BR and ru.

Either delete the key from all four, or keep it and put the percentage beside it. The second reads better: `0%` next to a spinner does not say what is happening, and since the percentage only tracks the download, extract, zip and copy all sit at `100%` with no explanation.

Moyasee (commented):
Cancellation holds up end to end. I checked the zip kill against the real 7z binary and the extract pipeline against a 400 MB tar, and both unwind with an AbortError, the child process dies, and the temp cleanup still runs on every path.

Four inline comments. One is a narrow behaviour bug where a late cancel throws away an export that already finished; the other three are cleanup, including the two issues Sonar raised on this diff.

Outside the diff: the PR has no label yet, and release note generation categorizes from it, so it needs one of feature/fix/documentation/translation. The title also got truncated with an ellipsis, and the tail ("… and cancellation") ended up as the first line of the body.

Once these are fixed we can approve and merge this.

PR #2670, opened 2026-08-09
feat: match custom games to a Steam catalogue entry

t0mbi:
## Summary
- Adds an optional Steam catalogue match to custom games: search suggestions appear while typing a custom game's title, in both the Add Custom Game modal and the rename flow.
- Picking a match auto-applies that Steam game's public catalogue artwork (icon/logo/hero/cover) and records the AppID as `matchedSteamObjectId` on the game.
- Also seeds `trackingExecutablePaths` from the community-maintained known-executable list for the matched AppID, so the process watcher can recognize the game even when launched through a different wrapper executable.

This lays the groundwork for extending catalogue-only features (shop details, achievements, stats, reviews, cloud save) to matched custom games in follow-up PRs.

## Test plan
- [ ] Add a custom game and type a title matching a real Steam game — confirm suggestions appear and picking one applies real Steam artwork
- [ ] Rename an existing custom game's title from Properties > General — confirm the same matching flow works there
- [ ] Confirm an unmatched custom game is unaffected (still uses its own icon/gradient hero)


Moyasee:
Could you please address the findings from Greptile, and fix the Sonar findings? Greptile is at 2/5 and Sonar reports 2 new issues.

Hachi-R:
Thanks for the work on this. We are closing this direction because Hydra already allows users to set any executable on an existing Steam catalogue game. Creating a custom game and then matching it back to Steam gives the same game two competing identities: a custom identity for launching and a Steam identity for metadata, achievements, and cloud saves. That makes ownership and synchronization ambiguous and requires special cases throughout the launcher. If the goal is to launch a Steam catalogue game through a different executable, the existing flow already supports selecting that game and setting the desired executable in its settings. We want to keep one catalogue identity instead of introducing matchedSteamObjectId across the application.

PR #2672, opened 2026-08-09
fix: allow creating a Steam shortcut for custom games from the General tab

t0mbi:
## Summary
The right-click context menu has always let a custom game create a Steam shortcut (it only checks for an executable path), but the Properties > General tab's shortcut button was unconditionally hidden for `shop === "custom"`, making the two entry points inconsistent.

This makes the General tab use the same executable-path-only gate as the context menu.

## Test plan
- [ ] Add a custom game with an executable path set — confirm "Create Steam Shortcut" now appears in Properties > General, matching the context menu
- [ ] Confirm the button still behaves correctly for regular catalogue games (unchanged)

### DIFF
--- a/src/renderer/src/pages/game-details/modals/game-options-modal.tsx
+++ b/src/renderer/src/pages/game-details/modals/game-options-modal.tsx
@@ -297,12 +297,10 @@ export function GameOptionsModal({
   }, [visible]);
 
   useEffect(() => {
-    if (game.shop !== "custom") {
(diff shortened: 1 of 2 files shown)


Moyasee:
Could you please address the findings from Greptile? It is at 4/5 with 1 open comment.

Hachi-R:
Closing this due to inactivity and the lack of a response. Feel free to reopen this PR whenever you want to continue working on it.

PR #2675, opened 2026-08-09
feat: add a custom game by dragging its executable onto the window

t0mbi:
## Summary
Drag an executable file (.exe on Windows, or the platform-appropriate extensions) from Explorer/Finder anywhere onto the app window to open the Add Custom Game modal pre-filled with that path and a title guessed from the filename — same result as picking the file via the Browse button.

A drop overlay shows while dragging a file over the window; dropping an unsupported file type shows an error toast instead of doing nothing.

Requires exposing `webUtils.getPathForFile` through the preload bridge, since Electron no longer exposes real filesystem paths on dropped `File` objects directly. The window-level `dragover`/`drop` handlers also stop Electron's default behavior of navigating to `file://<path>` when something is dropped on it.

## Test plan
- [ ] Drag a `.exe` onto the app window from anywhere in the UI — confirm the drop overlay appears and the Add Custom Game modal opens pre-filled
- [ ] Drag an unsupported file type — confirm an error toast appears and nothing else happens
- [ ] Confirm the window doesn't navigate away/break when a file is dropped
- [ ] Confirm the regular "+" button / Browse flow still works unaffected

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -72,6 +72,8 @@
     "custom_game_modal_success": "Custom game added successfully",
     "custom_game_modal_failed": "Failed to add custom game",
     "custom_game_modal_executable": "Executable",
+    "custom_game_modal_drop_hint": "Drop to add as a custom game",
+    "custom_game_modal_unsupported_file": "This file type can't be added as a game executable",
     "edit_game_modal": "Customize Assets",
     "edit_game_modal_description": "Customize game assets and details",
     "edit_game_modal_title": "Title",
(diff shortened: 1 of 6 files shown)


Moyasee:
Could you please address the findings from Greptile? It is at 4/5 with 2 open comments.

Hachi-R:
Thanks for the PR — dragging an executable in to skip the file picker is a good idea.

This isn't the shape we want for the feature, though. The drop is window-wide and always opens **Add custom game**, so dropping onto an existing library title (e.g. Terraria) creates a second custom entry instead of binding that game's executable. We'd rather not ship that as the interaction.

Closing for now. Happy to revisit a more targeted version later.

PR #2676, opened 2026-08-09
feat: extend achievement sync and automatic cloud save to matched custom games

t0mbi:
Depends on #2670 — this branch is stacked on top of it (the diff here will shrink to just this PR's commit once #2670 merges).

## Summary
Custom games matched to a Steam catalogue entry (#2670) were still fully excluded from `AchievementWatcherManager`'s remote sync and automatic cloud save on open/close, since both were gated behind a blanket `shop === "custom"` check. This routes achievement sync and automatic cloud save through the matched Steam identity for matched custom games, while:

- unmatched custom games are unaffected (still skipped, as before)
- Hydra-account library registration (`remoteId` / `trackGamePlaytime` / `createGame`) is deliberately left untouched for all custom games, matched or not — that's account-library bookkeeping that doesn't apply here
- the periodic remote playtime sync in `onTickGame` is unchanged for the same reason

Automatic cloud save still requires the user's own active Hydra Cloud subscription, same as it always has — this only removes the blanket exclusion for custom games, it doesn't add a new access path.

## Test plan
- [ ] Add a custom game, match it to a Steam game with achievements, launch and play it — confirm achievements sync
- [ ] With an active Hydra Cloud subscription, confirm automatic cloud save on close/open runs for a matched custom game
- [ ] Confirm an unmatched custom game still has no achievement/cloud-save sync (unchanged)
- [ ] Confirm regular catalogue games are unaffected


Moyasee:
Could you please address the findings from Greptile, and fix the Sonar findings? Greptile is at 0/5 with 5 open comments and Sonar reports 2 new issues.

Hachi-R:
Closing this together with #2670. Hydra already allows users to set any executable on an existing Steam catalogue game. Matching a custom game back to Steam makes the custom identity responsible for launching while the Steam identity becomes responsible for achievements and cloud saves. That split is ambiguous and can associate synchronization data with the wrong game. Achievements and cloud saves should remain attached to the existing catalogue identity, with the user configuring the executable they want on that game.

PR #2678, opened 2026-08-09
fix: launch games that require elevation on Windows

t0mbi:
## Summary
Games spawned with launch options (or wrapper commands like gamemode/mangohud) go through `child_process.spawn()` directly instead of the `shell.openPath()` fallback used when there are no launch options. `spawn()` can't show the UAC elevation prompt, so any game whose executable manifest declares `requestedExecutionLevel="requireAdministrator"` fails to launch with `spawn ... EACCES` the moment launch options are involved — Windows maps `ERROR_ELEVATION_REQUIRED` to `EACCES`. The same game launches fine from a desktop shortcut because Explorer elevates it automatically.

When `spawn()` fails with `EACCES` on Windows, this now retries through PowerShell's `Start-Process -Verb RunAs`, which shows the same UAC prompt Explorer would and passes launch arguments through `-ArgumentList` (base64-encoded via `-EncodedCommand` to avoid quoting issues with paths that contain spaces). This mirrors the elevation behavior the no-args `shell.openPath()` path already had.

Found this while testing a game that needs a launch flag (`-savetouserdir`) and also requires admin rights — it launched fine via a manual shortcut but failed from inside Hydra with exactly this error.

## Test plan
- [ ] Add a game whose executable requires admin rights (manifest `requireAdministrator`) with a launch option set — confirm it now shows the UAC prompt and launches instead of failing with EACCES
- [ ] Confirm a game with launch options that does *not* require elevation still launches normally (no UAC prompt, no regression)
- [ ] Confirm a game with no launch options still launches via the existing `shell.openPath()` path unchanged
- [ ] Confirm closing the game still works correctly afterward (this PR doesn't change the returned pid used for elevation-launched games, matching the existing null-pid convention already used by the `shell.openPath()` fallback)

### DIFF
--- a/src/main/helpers/launch-game.ts
+++ b/src/main/helpers/launch-game.ts
@@ -62,7 +62,12 @@ const ensureExecutablePermission = (executablePath: string) => {
   }
 };
 
-const quotePowerShellString = (value: string) => `'${value.replace(/'/g, "''")}'`;
+const quotePowerShellString = (value: string) => `'${value.replaceAll("'", "''")}'`;
+
+const getTrustedPowerShellPath = () => {
+  const systemRoot = process.env.SystemRoot || process.env.windir || "C:\\Windows";
+  return path.join(systemRoot, "System32", "WindowsPowerShell", "v1.0", "powershell.exe");
+};
 
 const buildElevatedStartProcessCommand = (
   command: string,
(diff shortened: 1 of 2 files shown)


t0mbi:
Closing this one — the elevation-retry mechanism this PR is built around doesn't actually work reliably. Tested it live against a real admin-required installer and couldn't get it to consistently trigger the UAC prompt through a spawned child process, despite the same mechanism working fine when invoked directly outside of Electron's child_process. Rather than merge something that doesn't hold up, I'm pulling it. May revisit with a different approach later, but nothing here is ready.

PR #2680, opened 2026-08-09
fix: ignore Python virtualenvs in git, prettier and packaging

BenWolst:

**Fill in the PR content:**

A Python virtualenv created at the repository root — the natural location for one, since `requirements.txt` lives there — is currently matched by no ignore list in the project. That breaks two things.

`electron-builder` fails outright:

    ⨯ unable to copy, file is symlinked outside the package  source=.venv/bin/python
    ⨯ Cannot copy file (python) symlinked to file (python3.13) outside the package
      as that violates asar security integrity  failedTask=build

and `yarn format-check` exits 2 after walking `site-packages`:

    [warn] .venv/lib/python3.x/site-packages/setuptools/tests/indexes/test_links_priority/external.html
    Error occurred when checking code style in the above file.

This adds `.venv*` / `venv` to `.gitignore` and `.prettierignore`, and excludes it from the `electron-builder` `files` glob. Six added lines, no behaviour change, and it also prevents a virtualenv being committed by accident.

`.eslintignore` is deliberately left alone — eslint skips dot-directories by default, so it is unaffected.


electron-builder.yml
```diff
@@ -14,6 +14,8 @@ files:
   - "!src/*"
   - "!native/hydra-native/target"
   - "!native/hydra-native/target/**"
+  - "!.venv*"
+  - "!.venv*/**"
```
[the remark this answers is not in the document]
BenWolst: Confirmed and fixed in 8855e3b.

Reproduced it first to be sure: created a non-dot `venv/` at the repo root and ran `electron-builder --mac --dir`, which failed exactly as described:

```
⨯ unable to copy, file is symlinked outside the package  source=venv/bin/python
Error: Cannot copy file (python) symlinked to file (python3.12) outside the package
  as that violates asar security integrity
```

`.gitignore` and `.prettierignore` already covered the non-dot case; the electron-builder glob was the one place that did not, so the three files disagreed. Added `!venv` / `!venv/**` and re-ran the same packaging command with `venv/` still present — it now completes successfully.

.prettierignore
```diff
@@ -1,6 +1,8 @@
 out
 dist
 seeds
+.venv*
+venv
```
Moyasee, line 5: The body says eslint is unaffected because it skips dot-directories. That holds for `.venv` but not for `venv`, and `yarn lint` is `eslint . --fix` with `.husky/pre-push` running it:

```
.../venv/lib/python3.12/site-packages/tool.js
x 2 problems      EXIT=1
```

So a non-dot venv fails the push gate and `--fix` rewrites files inside site-packages. Adding `venv` to `.eslintignore` takes it back to one.

electron-builder.yml
```diff
@@ -14,6 +14,10 @@ files:
   - "!src/*"
   - "!native/hydra-native/target"
   - "!native/hydra-native/target/**"
+  - "!.venv*"
```
Moyasee, line 17: These four are root-anchored while the `.gitignore` and `.prettierignore` entries aren't, so `python_rpc/.venv` is caught by those two and still breaks `yarn build:linux`. That directory has its own `setup.py`, so it is the likely place for one. `!**/.venv*` / `!**/.venv*/**` / `!**/venv` / `!**/venv/**` matches the existing `!**/.vscode/*` style.

Moyasee (commented):
Checked whether any of this was already covered and none of it is, so all three files need the addition. Two gaps inline. Once those are in we can approve and merge this.

Hachi-R (changes requested):
Two blocking gaps remain.

`.eslintignore:L1` — P2: The PR explicitly supports a non-dot `venv/`, but ESLint still scans it. Because `yarn lint` runs `eslint . --fix`, JavaScript fixtures inside `site-packages` can fail the pre-push gate or be rewritten. Add `venv/` to `.eslintignore`.

`electron-builder.yml:L17` — P2: These exclusions only cover virtualenvs at the repository root. A nested environment such as `python_rpc/.venv` is still packaged and can trigger the same external-symlink failure. Use recursive exclusions such as `!**/.venv*`, `!**/.venv*/**`, `!**/venv`, and `!**/venv/**`.

The branch is also 64 commits behind `main`, so it should be updated before rerunning the checks.

Hachi-R:
Closing this for now because the requested ignore-pattern fixes are still pending and the branch needs to be updated with `main`. Feel free to reopen this PR whenever you want to address them or continue iterating here.

PR #2681, opened 2026-08-09
fix: only apply Windows-only cx_Freeze options on Windows

BenWolst:

**Fill in the PR content:**

`python_rpc/setup.py` passes two Windows-only cx_Freeze options unconditionally.

**The icon has a visible effect on Linux.** cx_Freeze only *embeds* an icon into the executable on Windows — its Windows freezer calls `AddIcon(target_path, exe.icon)`. On every other platform the generic implementation is used instead, and its own comment describes what it does:

```python
def _add_resources(self, exe: Executable) -> None:
    """Add resources for an executable, platform dependent."""
    # Copy icon into application. (Overridden on Windows)
```

It copies the file into the build output. So the Linux build currently ships a 110 KB `icon.ico` inside `hydra-python-rpc/`, which is then packaged out through `extraResources`. Nothing reads it.

**`include_msvcr`** bundles the MSVC runtime and is silently ignored off Windows.

This guards both behind `sys.platform == "win32"`, matching the existing `get_windows_openssl_includes()` pattern already in the file. Windows builds are unchanged.

Verified by rebuilding before and after: the stray `icon.ico` disappears from the output directory, and the resulting binary still starts and completes its handshake (`{"event":"ready","protocolVersion":1}`).

### DIFF
--- a/python_rpc/setup.py
+++ b/python_rpc/setup.py
@@ -28,9 +28,11 @@ def get_windows_openssl_includes():
 build_exe_options = {
     "packages": ["libtorrent"],
     "build_exe": "hydra-python-rpc",
-    "include_msvcr": True,
     "include_files": get_windows_openssl_includes(),
 }
+
+if sys.platform == "win32":
+    build_exe_options["include_msvcr"] = True
 
 setup(
     name="hydra-python-rpc",
@@ -41,7 +43,9 @@ setup(
         Executable(
             "python_rpc/main.py",
             target_name="hydra-python-rpc",
-            icon="build/icon.ico",
+            # cx_Freeze only embeds the icon on Windows; elsewhere it copies
+            # the .ico into the build output, shipping a file nothing reads.
+            icon="build/icon.ico" if sys.platform == "win32" else None,
         )
     ],
 )


python_rpc/setup.py
```diff
@@ -41,7 +43,9 @@ def get_windows_openssl_includes():
         Executable(
             "python_rpc/main.py",
             target_name="hydra-python-rpc",
-            icon="build/icon.ico",
+            # cx_Freeze only embeds the icon on Windows; elsewhere it copies
```
Moyasee, line 46: Comment can come out. It is already verbatim in both the commit body and the PR description.

python_rpc/setup.py
```diff
@@ -28,10 +28,12 @@ def get_windows_openssl_includes():
 build_exe_options = {
     "packages": ["libtorrent"],
     "build_exe": "hydra-python-rpc",
-    "include_msvcr": True,
     "include_files": get_windows_openssl_includes(),
 }
 
+if sys.platform == "win32":
```
Moyasee, line 34: This guard has no effect. `freezer.py:99` is already `self.include_msvcr = IS_WINDOWS and bool(include_msvcr)`, so cx_Freeze does the platform check itself.

Dropping it would leave the diff as just the icon fix, which is the part that does something: `_add_resources` is only overridden in `WinFreezer`, so `LinuxFreezer` uses the base implementation and copies the icon into the output. Built against the pinned 7.2.3:

```
Only in BEFORE: icon.ico   (110226 bytes)
main binary sha256: identical both builds
```

`extraResources` ships that directory, so it lands in every Linux artifact and nothing reads it.

Moyasee (commented):
The icon change is worth having, so this isn't wasted. Two notes inline. Once those are sorted we can approve and merge this.

Hachi-R:
Closing this because the main change currently has no effect in the application flow; only the icon fix is useful. Please split out the icon fix or rework the main change so it has a verifiable runtime effect. Feel free to reopen the PR after that is addressed.

PR #2685, opened 2026-08-10
fix: detect game executable after installer finishes running

t0mbi:
## Summary
The only auto-detection of a game's executable ran once, immediately after extraction, scanning the raw downloaded/extracted folder — before any installer had a chance to place files anywhere. For installer-based repacks (FitGirl, DODI, etc.), that scan finds nothing, and nothing ever re-checks afterward: the installer is launched fire-and-forget (resolved on process *spawn*, not exit), so the game is never recognized as installed even after a fully successful install. Users are stuck seeing "Install"/"Download" prompts indefinitely and have to set the executable path manually to recover.

## Changes
- The direct-spawn and wine-fallback installer launch paths in `open-game-installer.ts` now observe real process exit, not just spawn.
- On installer exit, a new `rescanAndBindExecutableAfterInstall()` searches the plausible install destinations — the download folder again, the Wine prefix's `drive_c` on Linux, and `Program Files`/`Program Files (x86)`/`%LOCALAPPDATA%\Programs` on Windows — using the same known-executable-name matching already used by the extraction-time scan (`GameExecutables` + `findGameExecutableInFolder`'s ranking logic), and binds `executablePath` via the existing `updateGameExecutablePath()` helper if a match is found.
- Program Files scanning tolerates permission-restricted subfolders (`WindowsApps`, several Windows Defender folders, etc. — confirmed live while testing) by walking manually and catching errors per-directory, rather than using `fs.readdir`'s `recursive` option, which aborts the *entire* walk on the first unreadable subdirectory it hits.
- Also fixes a related but separate bug found in the same area: the Downloads page's "Install" button is driven by a check that only asks "does this folder still contain a `setup.exe`" — since installers don't delete themselves, that button would say "Install" forever regardless of whether the game had actually been detected as installed. It now also checks `executablePath`.
- The renderer needs no changes: it already listens for the `on-library-batch-complete` event (used elsewhere for the same purpose) and refreshes automatically once a match is found.

**Note**: an earlier version of this PR also attempted to fix Windows UAC elevation for installers requiring admin rights (as a separate PR, #2678). That didn't hold up under real-world testing and has been dropped from both — installers requiring elevation still rely on the existing `shell.openPath()` fallback, unchanged by this PR.

## Test plan
- [ ] Install an installer-based repack that doesn't require elevation — confirm "Play" appears automatically once the installer finishes, without manually setting the executable path
- [ ] Confirm a portable-style repack (no installer) is unaffected — still detected at extraction time as before
- [ ] Confirm manually setting the executable path while an installer is still running isn't clobbered by the post-exit scan (the scan re-checks `executablePath` immediately before writing)
- [ ] Confirm the Downloads page correctly switches away from "Install" once a game is detected as installed
- [ ] Confirm an installer that fails to launch, or is cancelled by the user, doesn't crash or hang anything — the scan just finds nothing and logs, no user-facing error

### DIFF
--- a/src/main/events/library/get-game-installer-action-type.ts
+++ b/src/main/events/library/get-game-installer-action-type.ts
@@ -3,7 +3,7 @@ import fs from "node:fs";
 
 import { getDownloadsPath } from "../helpers/get-downloads-path";
 import { registerEvent } from "../register-event";
-import { downloadsSublevel, levelKeys } from "@main/level";
+import { downloadsSublevel, gamesSublevel, levelKeys } from "@main/level";
 import { GameShop } from "@types";
 
 const getGameInstallerActionType = async (
@@ -15,6 +15,14 @@ const getGameInstallerActionType = async (
   const download = await downloadsSublevel.get(downloadKey);
 
   if (!download?.folderName) return "open-folder";
+
+  // Installers don't delete themselves after running, so a setup.exe check
+  // alone would offer "Install" forever, even for a game that's already
+  // been detected as installed (see rescanAndBindExecutableAfterInstall in
+  // open-game-installer.ts). Once executablePath is set, there's nothing
+  // left to install here.
+  const game = await gamesSublevel.get(downloadKey);
+  if (game?.executablePath) return "open-folder";
 
   const gamePath = path.join(
     download.downloadPath ?? (await getDownloadsPath()),
--- a/src/main/events/library/get-game-installer-action-type.ts
+++ b/src/main/events/library/get-game-installer-action-type.ts
@@ -16,13 +16,10 @@ const getGameInstallerActionType = async (
 
   if (!download?.folderName) return "open-folder";
 
-  // Installers don't delete themselves after running, so a setup.exe check
-  // alone would offer "Install" forever, even for a game that's already
-  // been detected as installed (see rescanAndBindExecutableAfterInstall in
-  // open-game-installer.ts). Once executablePath is set, there's nothing
-  // left to install here.
   const game = await gamesSublevel.get(downloadKey);
-  if (game?.executablePath) return "open-folder";
+  if (game?.executablePath && fs.existsSync(game.executablePath)) {
+    return "open-folder";
+  }
 
   const gamePath = path.join(
     download.downloadPath ?? (await getDownloadsPath()),
--- a/src/main/events/library/open-game-installer.ts
+++ b/src/main/events/library/open-game-installer.ts
@@ -5,11 +5,14 @@ import { spawn } from "node:child_process";
 
 import { getDownloadsPath } from "../helpers/get-downloads-path";
 import { updateGameExecutablePath } from "@main/helpers/update-executable-path";
+import { withGameRecordLock } from "@main/helpers/game-record-lock";
 import {
   rankExecutableCandidates,
+  type ExecutableSearchScope,
   type KnownGameExecutable,
 } from "@main/helpers/game-executable-ranking";
 import { registerEvent } from "../register-event";
+import { getInstallLocationScanDirectories } from "./scan-installed-games";
 import { downloadsSublevel, gamesSublevel, levelKeys } from "@main/level";
 import { GameShop } from "@types";
 import {
@@ -21,23 +24,6 @@ import {
   runAutomaticCloudSaveSync,
 } from "@main/services";
 
-// After an installer-based repack finishes installing, nothing else in the
-// app ever re-checks for the resulting executable -- the only auto-detection
-// pass runs once, right after extraction, before the installer has placed
-// any files. This scans the plausible install destinations once the
-// installer process actually exits, so the game is recognized as installed
-// without the user having to set the executable path manually.
-//
-// Program Files on Windows contains a long tail of folders locked down to
-// TrustedInstaller/System that throw EPERM even for elevated processes --
-// WindowsApps, various Windows Defender folders, etc. There's no reliable
-// list of which ones; fs.readdir's `recursive` option also aborts the
-// *entire* walk on the first unreadable subdirectory it hits, so scanning
-// Program Files as one big recursive tree means a single restricted folder
-// anywhere in the tree can silently block finding an otherwise perfectly
-// accessible install elsewhere under it. This walks manually instead,
-// catching permission errors per-directory and just skipping that subtree,
-// so one inaccessible folder never takes out the rest of the scan.
 const MAX_INSTALL_SCAN_DEPTH = 6;
 
 const collectAccessibleFilePaths = async (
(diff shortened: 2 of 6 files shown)


src/main/events/library/open-game-installer.ts
```diff
@@ -4,12 +4,175 @@ import fs from "node:fs";
...
+};
+
+const findGameExecutableResilient = async (
+  folderPath: string,
+  executables: KnownGameExecutable[]
+): Promise<string | null> => {
+  if (executables.length === 0) return null;
+
+  const relativeFilePaths = await collectAccessibleFilePaths(folderPath);
+  const match = rankExecutableCandidates(relativeFilePaths, executables);
```
Moyasee: `rankExecutableCandidates` takes a third `scope` arg and this call omits it, so it defaults to `"installation"`. `scan-installed-games.ts:262` passes `"library"` for exactly this situation, because `"library"` returns null when the candidates span more than one installation while `"installation"` commits to a pick. Since the rescan fires on any installer exit, including a cancel, a cancelled install can bind a stranger's exe.

Same synthetic Program Files through the two scopes:

```
candidate folder 1 (download folder): null
candidate folder 2 (Program Files) : /var/.../Program Files/AcmeChat/launcher.exe
same input, library scope          : null
```

The overlap isn't theoretical. From `assets.hydralauncher.gg/game-executables.json`, 769 of 9532 basenames are claimed by more than one game: `game.exe` by 223, `nw.exe` by 38, `dosbox.exe` by 20, `launcher.exe` and `client.exe` by 6 each. Those all collide with ordinary software in Program Files.

Pass `"library"` for the shared roots. The download folder can stay `"installation"`.

src/main/events/library/open-game-installer.ts
```diff
@@ -4,12 +4,175 @@ import fs from "node:fs";
...
+    );
+
+    const candidateFolders = [downloadFolderPath];
+
+    if (process.platform === "linux" && winePrefixPath) {
+      candidateFolders.push(path.join(winePrefixPath, "drive_c"));
+    }
+
+    if (process.platform === "win32") {
+      candidateFolders.push(
```
Moyasee: These aren't the directories installer-based repacks target. `scan-installed-games.ts:28-34` already has the real defaults (`C:\Games`, `D:\Games`, steamapps\common) and they default to `C:\Games\<Title>`, so this scans three large trees they rarely use and misses the one they do.

Narrowing to that set also fixes the cost. The walk is `Promise.all` over every entry at every level, and ranking is synchronous on the main process:

```
files=66,870   walkMs=213   maxConcurrentReaddir=1873
files=150,000  syncRankMs=51.4   (126.2 with a tie-break)
```

src/main/events/library/get-game-installer-action-type.ts
```diff
@@ -16,6 +16,14 @@ const getGameInstallerActionType = async (
 
   if (!download?.folderName) return "open-folder";
 
+  // Installers don't delete themselves after running, so a setup.exe check
+  // alone would offer "Install" forever, even for a game that's already
+  // been detected as installed (see rescanAndBindExecutableAfterInstall in
+  // open-game-installer.ts). Once executablePath is set, there's nothing
+  // left to install here.
+  const game = await gamesSublevel.get(downloadKey);
+  if (game?.executablePath) return "open-folder";
```
Moyasee: `executablePath` is only cleared by `delete-game-folder.ts:63`, `remove-game-from-library.ts:45` and the subscription-gated `executable-path-guard.ts:44`. Uninstall through the game's own uninstaller, or delete the folder in Explorer while keeping the download, and the row says "Open folder" permanently. That is also the escape hatch when the scan picks wrong. `game?.executablePath && fs.existsSync(game.executablePath)`.

src/main/events/library/open-game-installer.ts
```diff
@@ -4,12 +4,175 @@ import fs from "node:fs";
...
+  GameExecutables,
+  logger,
+  Umu,
+  WindowManager,
+  Wine,
+  runAutomaticCloudSaveSync,
+} from "@main/services";
 
-const launchInstallerWithWine = async (filePath: string): Promise<boolean> => {
+// After an installer-based repack finishes installing, nothing else in the
```
Moyasee: Please drop the added comments here and at 57 and 143, plus `get-game-installer-action-type.ts:19-23` and `umu.ts:450-452`. The PR body already says all of it and that is where reviewers read it.

`umu.ts:449` is the one that genuinely reads opaque without its comment. `const alreadyResolvedAsLaunched = settled;` or a rename of `settled` makes the branch explain itself.

src/main/events/library/open-game-installer.ts
```diff
@@ -4,12 +4,175 @@ import fs from "node:fs";
...
+      // Re-check under the lock of "hasn't been set since we started" --
+      // the user may have set it manually while the installer was running.
+      const latestGame = await gamesSublevel.get(gameKey);
+      if (!latestGame || latestGame.executablePath) return;
+
+      logger.info(
+        `[openGameInstaller] Auto-detected executable after installer exit for ${objectId}: ${foundExePath}`
+      );
+
+      await gamesSublevel.put(gameKey, {
```
Moyasee: Non-blocking. `GameFilesManager.searchAndBindExecutable()` at `game-files-manager.ts:413-451` is this same sequence, and it also calls `createDesktopShortcutForGame`. So detection at extraction time creates a shortcut and detection after an installer doesn't. Reusing it drops about 60 lines.

src/main/events/library/open-game-installer.ts
```diff
@@ -4,12 +4,175 @@ import fs from "node:fs";
...
+// list of which ones; fs.readdir's `recursive` option also aborts the
+// *entire* walk on the first unreadable subdirectory it hits, so scanning
+// Program Files as one big recursive tree means a single restricted folder
+// anywhere in the tree can silently block finding an otherwise perfectly
+// accessible install elsewhere under it. This walks manually instead,
+// catching permission errors per-directory and just skipping that subtree,
+// so one inaccessible folder never takes out the rest of the scan.
+const MAX_INSTALL_SCAN_DEPTH = 6;
+
+const collectAccessibleFilePaths = async (
```
Moyasee, line 29: Non-blocking. The EPERM-tolerant walk is better than `fs.readdir({recursive:true})` and the extraction-time scan wants it too. Worth putting behind an option on `findGameExecutableInFolder` instead of a private copy in an event handler.

Moyasee (changes requested):
The rescan is the right idea, but as written it can bind an executable that isn't the game, and once it does there is no way back to the Install action. Notes inline. Leaving greptile's three open ones alone since they still stand.

PR #2687, opened 2026-08-10
feat: enable crash reporter and log process lifecycle events

Moyasee:
We had a user report where Hydra disappears with no error window, nothing useful in the logs, and no `hydra.exe` left in Task Manager. Trying to diagnose it, I found we have no way to see what happened:

- `crashReporter.start()` is never called anywhere, so Crashpad never initializes and no minidump is ever written.
- `log.errorHandler.startCatching({ showDialog: false })` in `src/main/services/logger.ts` catches `uncaughtException` and `unhandledRejection`, logs them, and returns. Nothing rethrows or exits. So JS faults can't kill the app, and never surface a dialog either.

Between those two, a silent exit leaves nothing behind at all.

This starts the crash reporter with `uploadToServer: false`. Dumps stay in the Crashpad folder under userData and nothing is transmitted anywhere. The path gets logged once at startup so we can point people at it.

It also logs three lifecycle events Electron already emits and we were discarding. `child-process-gone` and `render-process-gone` carry a reason and exit code when the GPU, a utility process, or a renderer dies. `will-quit` fires on any clean shutdown.

That last one is the useful bit for triage, because it splits a silent disappearance three ways:

| observed | meaning |
| --- | --- |
| quit line, no dump | something called `app.quit()` |
| dump, no quit line | native crash, and the dump names the module |
| neither | process was killed from outside |

Verified locally. A forced native crash writes a dump and no quit line, and the stack symbolizes off `symbols.electronjs.org` down to file and line. A deliberate `app.quit()` logs the quit line and writes no dump.

Dumps can be read with:

```
minidump-stackwalk --brief --symbols-url https://symbols.electronjs.org <dump>
```

Frames inside `hydra_native.node` will show as module plus offset rather than function names, since we don't keep a PDB per release. Module attribution still works, which is enough to tell whether a crash came from Electron, `classic_level.node`, or our own addon. Worth revisiting if dumps start pointing at the addon.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,4 +1,11 @@
-import { app, BrowserWindow, net, powerMonitor, protocol } from "electron";
+import {
+  app,
+  BrowserWindow,
+  crashReporter,
+  net,
+  powerMonitor,
+  protocol,
+} from "electron";
 import updater from "electron-updater";
 import i18n from "i18next";
 import path from "node:path";
@@ -21,6 +28,10 @@ import { launchGame, openClassicsGame } from "./helpers";
 import { refreshPortableShortcutLauncher } from "./helpers/shortcut-launch";
 import { lookupCachedPlatform } from "./events/library/get-library";
 import { loadState } from "./main";
+
+crashReporter.start({
+  uploadToServer: false,
+});
 
 const { autoUpdater } = updater;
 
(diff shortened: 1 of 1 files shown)


Hachi-R (approved):
No blocking issues found. Crashpad starts before app readiness, and the lifecycle listeners preserve useful process-exit diagnostics. Typecheck and CI checks pass.

chubbygrannychaser (approved):

PR #2691, opened 2026-08-10
fix: native addon crashing on load for some Windows machines

### LINKED ISSUE #2688: Hydra 4.1.0 crashes when adding a game to library — BEX64 / 0xC0000409 / ucrtbase.dll
Environment
Hydra version: 4.1.0.0
OS: Windows 11 24H2
Windows build: 26100.8875
Architecture: Windows x64
Node native dependency: classic-level 2.0.0
Image dependency: sharp 0.34.5
Native addon: hydra-native.node
hydra-native.node size: 10,714,112 bytes
hydra-native.node SHA-256: 927A74A7DE42F4782CDA3676F21BA81488E8EED0B7B1CEF85FE73481052EAF7A
app.asar SHA-256: 9089B64E7F1109BFA50016A8F54E9AA6446F5408DB2DF258F9F3D068C0EF9464
Problem

Since updating to Hydra 4.1.0, Hydra consistently crashes when a game is added to the user's library.

The problem happens with both:

An existing account.
A completely new account.

Therefore, it does not appear to be specific to the user's account data.

Reproduction steps
Install Hydra 4.1.0.
Start Hydra normally.
Log into an existing account, or create a new account.
Add a game to the library.
Hydra immediately closes/crashes.
Additional testing

I also tested with a completely fresh local database.

I closed Hydra and renamed:

%APPDATA%\hydralauncher\hydra-db

to:

hydra-db.backup

Hydra then created a new database.

After logging into a new account and adding a game to the library, Hydra crashed again in exactly the same way.

Therefore, the crash is reproducible with:

a new account;
a new local database;
the same Hydra 4.1.0 installation.

Clearing the cache/database does not resolve the problem.

Windows Error Reporting

Windows Event Viewer reports:

Event: BEX64
Application: Hydra.exe
Application version: 4.1.0.0
Faulting module: ucrtbase.dll
ucrtbase.dll version: 10.0.26100.8875
Exception code: 0xc0000409
Exception parameter: 0x7

The complete Report.wer generated by Windows is attached to this issue.

Native modules

The WER report shows that Hydra has several native components loaded, including:

classic_level.node
sharp-win32-x64.node
libvips-cpp-8.17.3.dll
libvips-42.dll
hydra-native.node
VCRUNTIME140.dll

I am not claiming that any specific module is the cause, but the crash appears to be a native crash rather than a normal JavaScript/Electron exception.

GPU test

I also tested Hydra with:

Hydra.exe --disable-gpu

In that mode Hydra did not display a window, so this test did not provide evidence that GPU acceleration is the cause.

Expected behavior

Adding a game to the library should add the game normally without terminating Hydra.

Actual behavior

Hydra terminates immediately after adding the game to the library.

Possible area to investigate

The crash appears strongly correlated with the library-add/synchronization flow introduced or changed in 4.1.0.

In particular, it may be worth investigating the code path that adds the game to the remote profile and then persists it locally.

The issue is reproducible even with a newly created local database and a newly created account.

Workaround

At the moment, the only reliable workaround I have found is using the previous Hydra version instead of 4.1.0.

Additional information

I can provide the complete Windows Report.wer file and additional logs if needed.

Thank you for looking into this.

Moyasee:
Closes #2688.

**Fill in the PR content:**

Some Windows users on 4.1.0 lose the app the instant they add their first game. No error dialog, nothing in the logs, the window is just gone. Windows Error Reporting records an abort rather than memory corruption: exception `0xC0000409`, parameter `7` (`FAST_FAIL_FATAL_APP_EXIT`), faulting module `ucrtbase.dll`.

The cause came in without anyone picking it. reqwest 0.13 changed its default TLS backend from native-tls to rustls, and reqwest's `rustls` feature selects the aws-lc-rs provider, so `reqwest = { version = "0.13", features = ["stream"] }` quietly pulls `aws-lc-sys` into the addon. AWS-LC registers a `.CRT$XCU` constructor that runs `OPENSSL_cpuid_setup` during `LoadLibrary`, so it executes inside the `require()` of `hydra-native.node`, before any of our code gets a chance to run. On some machines that init aborts.

The reason it presents as a library bug is the load timing. `watchProcesses` returns early while the library is empty, before it ever calls into the addon, and nothing else in a normal startup touches it, so on a fresh profile the addon is never mapped into the process at all. Adding the first game lets the next 2 second tick through, the worker thread requires the addon, and the constructor finally runs. A native abort on a worker thread takes the whole process down with it, which is why nothing reaches `error.txt` and why it reads as an instant close rather than a crash.

The fix moves reqwest to `rustls-no-provider` and installs ring as the process default provider. Same stack on Windows and Linux, no OpenSSL to package on the Linux side, and no C crypto carrying a static initializer. rustls needs the provider installed before any client is built, and the manifest cache builds its own client separately from the blob client, so both call sites install it.

Verification, on top of the reporter confirming that a build with the crypto removed stops the crash while the shipped 4.1.0 addon reproduces it every time:

- the built binary contains no `AWS-LC` or `OPENSSL_ia32cap` strings, and `aws-lc-sys` is gone from the resolved graph
- `getSaveRulesForGame` for Steam 440 returns the same `ruleSourceRevision` and the same rules payload as the aws-lc build
- blob download over HTTPS reaches the same size check as before, so handshake, certificate validation against the OS trust store, and body transfer all still work
- 162 native tests pass, including the ones that fetch the real manifest over HTTPS

One behavioural difference worth flagging: ring supports a narrower set of suites than aws-lc-rs, most visibly no post-quantum key exchange. Nothing we connect to requires it and servers negotiate down, but it is a real difference rather than a drop-in swap.

Kept deliberately out of scope, happy to open these separately: `catch_unwind` on the napi exports so a native panic surfaces as a JS error instead of killing the app, trimming `System::new_all()` in `list_processes` down to the fields we actually read, and recovering from a crash that leaves an orphaned process holding the single instance lock, which currently makes every later launch exit silently.

### DIFF
--- a/native/hydra-native/Cargo.lock
+++ b/native/hydra-native/Cargo.lock
@@ -48,29 +48,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
-name = "aws-lc-rs"
-version = "1.17.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00bdb5da18dac48ca2cc7cd4a98e533e8635a58e2361d13a1a4ee3888e0d72f1"
-dependencies = [
- "aws-lc-sys",
- "zeroize",
-]
-
-[[package]]
-name = "aws-lc-sys"
-version = "0.43.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43103168cc76fe62678a375e722fc9cb3a0146159ac5828bc4f0dfd755c2224c"
-dependencies = [
- "cc",
- "cmake",
- "dunce",
- "fs_extra",
- "pkg-config",
-]
-
-[[package]]
 name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -146,8 +123,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5add81bb678e6cb321aff7fa0dc7689ad82b112dbc032cea19f91d6b8e3582b9"
 dependencies = [
  "find-msvc-tools",
- "jobserver",
- "libc",
  "shlex",
 ]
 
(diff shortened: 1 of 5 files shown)


Hachi-R (approved):
No blocking issues found. The ring provider is installed before both reqwest client construction paths, `aws-lc-sys` is removed from the resolved graph, and the compile and CI checks pass.

Moyasee (commented):
.

PR #2697, opened 2026-08-12
Feat/achievements metadata export

galinhaman22:

**Fill in the PR content:**

Generates missing achievement metadata for emulators and broadens where achievement files are found.

**Features**

- Generates `achievements.json` and downloads icons into a game's emulator settings folder when missing, on launch and after download.
- Finds settings folders across flat, Unreal and Unity layouts, and when nested one level deeper, whichever executable is selected.
- Finds unlocks nested under the app id folder and inside the game's own settings folder, with no hardcoded id list.
- Adds an ALI213 parser for `Achievements.Bin`.

**Fixes**

- Emulator profile folders are user defined, not always `Player`, and every profile is watched.
- Game directory scan now resolves executables added on Linux.
- `achievements.json` points at the images folder that actually exists.
- Reset deletes every collected file, survives missing ones, and re-detects a restored file without a restart.
- Startup no longer reports the same achievements twice.


galinhaman22:
@greptile review again

Hachi-R (changes requested):
## Findings

- `src/main/services/achievements/metadata-export/generate-achievement-metadata.ts:L61` — P1: `writeFile` follows a pre-existing `achievements.json.tmp` symlink inside downloaded game content, so a crafted repack can point it outside the install and make Hydra truncate any writable target automatically on launch or after download. Validate the canonical `steam_settings` path against the canonical game root, reject symlinked path components, and create the temporary file with randomized exclusive/no-follow semantics before renaming it.

- `src/main/services/achievements/metadata-export/index.ts:L52` — P2: `withTimeout` only stops awaiting `generateAchievementMetadata`; the underlying scan, fetch, and file writes continue after `finishExport` removes the controller, so closing the game cannot stop them and a later export can race the same files. Pass the abort signal through generation and keep the controller active until the underlying operation settles.

- `src/main/services/achievements/metadata-export/index.ts:L50` — P2: Launcher status is sent to whichever global launcher window is current, without `gameKey`. Concurrent exports for different games, including the background export in `game-files-manager.ts`, can overwrite the active game status with stale progress or `complete`. Include `gameKey` in the payload and ignore events that do not match the window `shop` and `objectId`, or bind updates to the originating window.

Hachi-R (changes requested):
## Finding

- `src/main/services/achievements/metadata-export/find-steam-settings-directories.ts:L106` — P1: `isDirectory` uses `fs.stat`, so a symlinked `steam_settings` directory is still accepted. The randomized exclusive temp file fixes the reported `achievements.json.tmp` attack, but both `writeAchievementsFile` and `downloadAchievementIcons` can still write through a crafted directory symlink outside the game install. Resolve the game root and every candidate with `realpath`, require each candidate to remain inside that root, reject symlinked path components, and repeat the containment check immediately before writing.

## CI

- `lint`, text integrity, and SonarCloud pass.
- Windows and Ubuntu builds fail during dependency downloads with `socket hang up` and `EOF`, after compilation; those failures appear unrelated to this change.

Moyasee:
Could you please fix the failing builds? Both the Windows and Ubuntu jobs are red.

Hachi-R (changes requested):
## Finding

- `src/main/services/achievements/metadata-export/download-achievement-icons.ts:L60` — P1: `directory` comes from the game-supplied `achievements.json`, so a value such as `../../../.ssh/authorized_keys/icon.png` reaches `fs.promises.mkdir` before containment is checked and can create arbitrary directories outside the game root. Reject absolute and traversal paths before `mkdir`, and validate/create icon directories relative to the resolved `steam_settings` directory rather than the broader game root.

Hachi-R (approved):

PR #2698, opened 2026-08-12
fix(shortcuts): honor create-shortcut preference on Linux

Sociable-Fish:
Gates auto-created desktop shortcuts on Linux with the existing createStartMenuShortcut user preference. Hoists the preference check before icon download so Windows and Linux skip unnecessary work when disabled, and removes the duplicated Windows-only internal check. Shows the toggle in settings and Big Picture on both Windows and Linux. macOS behavior is unchanged.

**Fill in the PR content:**


src/big-picture/src/pages/settings/downloads-behavior-section.tsx
```diff
@@ -113,7 +114,7 @@ export function DownloadsBehaviorSection({
       },
     ];
 
-    if (isWindows) {
+    if (isWindows || isLinux) {
```
Moyasee, line 117: Showing the checkbox on Linux makes it the last item in this section, but `getLastDownloadsBehaviorItemFocusId` in `src/big-picture/src/pages/settings/settings-navigation.ts:193` still only knows about Windows. Downloads Sources feeds that helper into the `up` override for its sync / delete-all buttons (`downloads-sources-section.tsx:127`), so on Linux going up out of the sources actions lands on "Delete archive files after extraction" and skips right past the new checkbox. Going down into sources works, since that direction is derived from `items`, so it is a one-way trap on a controller.

Rendering this section and calling the helper on both branches:

```
--- main ---
platform=win32
  behavior items rendered : 5
  actual last item        : downloads-behavior-create-start-menu-shortcut
  sources section "up"    : downloads-behavior-create-start-menu-shortcut
  reachable going up      : yes
platform=linux
  behavior items rendered : 4
  actual last item        : downloads-behavior-delete-archive-files-after-extraction-by-default
  sources section "up"    : downloads-behavior-delete-archive-files-after-extraction-by-default
  reachable going up      : yes
--- pr ---
platform=win32
  behavior items rendered : 5
  actual last item        : downloads-behavior-create-start-menu-shortcut
  sources section "up"    : downloads-behavior-create-start-menu-shortcut
  reachable going up      : yes
platform=linux
  behavior items rendered : 5
  actual last item        : downloads-behavior-create-start-menu-shortcut
  sources section "up"    : downloads-behavior-delete-archive-files-after-extraction-by-default
  reachable going up      : NO - skipped
```

The helper needs the same predicate this file uses. Simplest fix is changing it to take `showsShortcutItem: boolean` and passing `isWindows || isLinux` from `downloads-sources-section.tsx`, which also stops the two lists drifting apart again next time a platform is added.

src/renderer/src/pages/settings/settings-behavior.tsx
```diff
@@ -312,7 +312,8 @@ export function SettingsBehavior() {
         }
       />
 
-      {window.electron.platform === "win32" && (
+      {(window.electron.platform === "win32" ||
+        window.electron.platform === "linux") && (
         <CheckboxField
           label={t("create_shortcuts_on_download")}
```
Moyasee, line 318: `create_shortcuts_on_download` is "Create desktop and Start Menu shortcuts when a game finishes downloading", which is not what Linux users get. `createDesktopShortcutForGame` only writes a desktop entry on the non-Windows branch, and `create-game-shortcut.ts:206` throws "Start Menu shortcuts are only available on Windows." for the manual path.

Either reword the shared key so it does not name a Windows-only location, or add a Linux variant and pick per platform. Same label is rendered by `settings-context-downloads.tsx:260`, so whichever way you go it needs to cover both. If you add a key, en / ru / es / pt-BR should all get it in this commit.

Moyasee (changes requested):
Main-process change looks right to me, and hoisting the preference read above the icon download is a nice bonus.

Two things to fix on the UI side before this can go in, both Linux-only. Details inline.

One unrelated thing: the PR still has no label, and it needs one of feature / fix / documentation / translation for the release notes to categorize it.

Moyasee (approved):
LGTM!

PR #2700, opened 2026-08-13
feat: souvenirs for achievements

Moyasee:
Supersedes #1820.

When an achievement unlocks, Hydra grabs the screen, uploads it to the souvenirs bucket, and the image shows up next to that achievement in the game's achievement list and in a new Souvenirs tab on the profile.

#1820 has been open since October and the merge base is January. `mergeAchievements` was rewritten since then, achievements moved out of `gameAchievementsSublevel` into `AchievementMemoryStore`, and settings grew the category sidebar, so `AchievementImageService` was writing to a sublevel that no longer exists and the branch's whole settings tab was dead weight. Rebasing 40 commits across 1700 upstream ones was not going to produce anything reviewable, so this is a fresh branch and the pieces that survived were rewritten against what is on main. The settings redesign, the bespoke fullscreen modal, and the helper extraction from `mergeAchievements` are all gone.

### Cloud only

`GET /users/:id/achievements` returns `null` for the whole payload unless the profile owner has an active subscription, and the presigned upload needs one too. Capturing for free users would write images to disk that could never surface anywhere, so the toggle opens the Hydra Cloud modal instead and no capture runs.

### Where the capture happens

In `mergeAchievements`, between publishing the notification and syncing with the API, gated on new unlocks plus `publishNotification`. That second condition matters: without it, the first sync of a game with 200 already-unlocked achievements would fire 200 screen grabs and 200 uploads. A failed capture logs and moves on rather than taking the sync down with it.

Screenshots are kept under `userData/Screenshots/<game>/<achievement>.jpeg`, resized to fit 1280x720 at quality 80, and trimmed to the 50 most recent. There is a button in settings to open the folder.

### Reading them back

`getUnlockedAchievements` only fetches souvenirs when the caller asks for fresh data. The watcher calls it with cached data every couple of seconds, and hitting the API on that path would have been a request every tick per running game.

### Linux

Off for now. `desktopCapturer` under Wayland goes through xdg-desktop-portal, which would pop a permission dialog mid-game the first time. Worth doing properly in a follow-up rather than shipping something that interrupts play.

### Not covered

`MAX_RECENT_ACHIEVEMENTS` caps that endpoint at 12, so the Souvenirs tab can only ever show twelve across every game. It reads like a full gallery and is not one. Either the tab gets relabeled as recent souvenirs or the API needs a paginated endpoint, and I would rather settle that here than guess.

---


src/main/services/screenshot.ts
```diff
@@ -0,0 +1,177 @@
...
+
+const getPrimaryScreenSource = async () => {
+  const sources = await desktopCapturer.getSources({
+    types: ["screen"],
+    thumbnailSize: { width: 1920, height: 1080 },
+  });
+
+  if (!sources.length) return null;
+
+  const primaryDisplayId = String(screen.getPrimaryDisplay().id);
```
Hachi-R: `src/main/services/screenshot.ts:L43` — P1: `captureGameScreenshot` always selects the primary display and never verifies that the tracked game owns the foreground window. A game on another monitor, or a delayed unlock while the user is alt-tabbed, uploads unrelated desktop content as a public souvenir. Pass the tracked process/window into capture, select its display, and skip capture unless foreground ownership can be verified.

src/main/services/emulators/emulator-souvenir-config.ts
```diff
@@ -0,0 +1,155 @@
...
+export const enableRetroArchAchievementScreenshots = (
+  executablePath: string
+) => {
+  const configPath = findRetroArchConfig(executablePath);
+
+  if (!configPath) return;
+
+  try {
+    const content = fs.readFileSync(configPath, "utf8");
+    const updated = setCfgValue(content, "cheevos_auto_screenshot", "true");
```
Hachi-R: `src/main/services/emulators/emulator-souvenir-config.ts:L134` — P2: This permanently enables `cheevos_auto_screenshot`, and imported source screenshots are never removed. RetroArch keeps writing achievement screenshots after the Hydra option is disabled, so its directory grows without the 50-file cleanup applied to Hydra copies. Use a session-scoped append config and temporary screenshot directory, then restore or remove both when the session ends.

src/main/services/emulators/emulator-session-tracker.ts
```diff
@@ -85,6 +89,17 @@ export const startEmulatorSession = async ({
     session.heartbeat.unref?.();
   }
 
+  if (game.shop === "launchbox") {
+    void startEmulatorSouvenirWatcher({
```
Hachi-R, line 104: `src/main/services/emulators/emulator-session-tracker.ts:L93` — P1: Watcher startup is detached from session lifetime. If the emulator exits while `startEmulatorSouvenirWatcher` is awaiting subscription/preferences, `finalizeEmulatorSession` runs before the interval is registered; the watcher then starts afterward and survives forever with stale game identity, potentially attaching later unlocks to the wrong game. Await startup or recheck that the same session is still registered immediately before storing the watcher.

src/big-picture/src/pages/profile/use-profile-data.ts
```diff
@@ -583,3 +584,36 @@ export function useRecentAchievements(
...
+  useEffect(() => {
+    if (!targetUserId || !hasActiveSubscription) {
+      setSouvenirs([]);
+      return;
+    }
+
+    let isMounted = true;
+
+    globalThis.window.electron.hydraApi
+      .get<ProfileAchievement[]>(`/users/${targetUserId}/achievements`, {
```
Hachi-R: `src/big-picture/src/pages/profile/use-profile-data.ts:L603` — P1: `/users/:id/achievements` is capped at 12 recent achievements, but this response backs the Souvenirs collection on both profile surfaces. Once a user has more than 12 souvenirs, older images become inaccessible even though they still exist. Add a dedicated paginated souvenirs endpoint and paginate the gallery before shipping it as a collection.

Hachi-R (changes requested):
Four blockers remain before this can merge: screenshot targeting can upload unrelated desktop content, emulator configuration changes persist outside Hydra, watcher startup can outlive a failed session, and the profile gallery is backed by a twelve-item endpoint. CI is green, but these runtime behaviors need to be fixed.

Moyasee:
@greptile

Moyasee:
@greptile

Moyasee:
@greptile

Moyasee:
@greptile

Moyasee:
@greptile

src/main/services/achievements/grouped-souvenir-retry-policy.ts
```diff
@@ -0,0 +1,52 @@
...
+    responseData?.errorCode ?? responseData?.code ?? responseData?.message;
+  if (
+    error.response.status === 409 &&
+    responseCode === SOUVENIR_CONFLICT_CODE
+  ) {
+    return true;
+  }
+
+  const responseClientId = responseData?.clientId;
+  return responseClientId === clientId;
```
Hachi-R: `src/main/services/achievements/grouped-souvenir-retry-policy.ts:L51` — P2: The current API filter serializes `HttpError` as `{ message }` and drops `details.clientId`, so terminal `409` responses such as `achievements/souvenir-upload-deleted` and `achievements/souvenir-upload-length-mismatch` reach this comparison without a client ID and remain pending forever. Classify the known terminal `409` codes directly, or preserve `HttpError.details` in the API response.

src/main/services/emulators/emulator-souvenir-config.ts
```diff
@@ -0,0 +1,349 @@
...
+      updatedBackups,
+      { valueEncoding: "json" }
+    );
+
+    const updated = setRetroArchSouvenirConfigValues(
+      content,
+      screenshotDirectory
+    );
+
+    if (updated !== content) fs.writeFileSync(configPath, updated, "utf8");
```
Hachi-R: `src/main/services/emulators/emulator-souvenir-config.ts:L212` — P2: This still writes `cheevos_auto_screenshot` and `screenshot_directory` into the main `retroarch.cfg`, while source cleanup only runs inside Hydra's watcher. If Hydra exits, crashes, loses the subscription, or disables the option while RetroArch is still running, RetroArch keeps producing files in `hydra-souvenirs` with no watcher removing them and may rewrite the enabled values on exit. Use a per-session append config and temporary directory, then remove both when the emulator session ends.

src/big-picture/src/pages/profile/use-profile-data.ts
```diff
@@ -583,3 +592,161 @@ export function useRecentAchievements(
...
+      setTotal(0);
+      setHiddenReason(null);
+      setIsLoading(false);
+      setLoadedTargetUserId(null);
+      setIsLoadingMore(false);
+      isLoadingMoreRef.current = false;
+      return;
+    }
+
+    const requestGeneration = requestGenerationRef.current + 1;
```
Hachi-R, line 625: `src/big-picture/src/pages/profile/use-profile-data.ts:L625` — P2: Switching from one profile to another leaves the previous profile's `souvenirs`, `total`, and `hiddenReason` in state until the new request resolves. `ProfileSouvenirs` renders whenever `souvenirs.length` is nonzero, so the old user's images and actions appear under the new profile during a slow request. Clear the page state when `targetUserId` changes, or render the data only when `loadedTargetUserId === targetUserId`.

Hachi-R (changes requested):
Requesting changes for the three runtime issues in the inline comments. GitHub Actions and the Sonar quality gate are green.

src/main/services/emulators/emulator-souvenir-config.ts
Moyasee: Fixed in a5f4eb9b2. RetroArch now receives a per-session --appendconfig file with a temporary screenshot directory and config_save_on_exit disabled. Hydra no longer writes souvenir settings to the main retroarch.cfg. The watcher uses only the session directory, which is removed after a failed launch, when the session ends, when souvenirs are disabled, and during a normal app shutdown. Existing legacy backups are still restored for users who ran an earlier build.

src/main/services/achievements/grouped-souvenir-retry-policy.ts
Moyasee: Fixed in a5f4eb9b2. The retry policy now treats achievements/souvenir-upload-deleted and achievements/souvenir-upload-length-mismatch as terminal 409 responses even when no client ID is present. I also added regression coverage for both responses.

src/big-picture/src/pages/profile/use-profile-data.ts
Moyasee: Fixed in a5f4eb9b2. Big Picture now clears the souvenir state as soon as the target user changes and gates the returned items by the loaded user ID, so the previous profile cannot render or expose actions while the next profile is loading.

Hachi-R (approved):
Rechecked the three runtime issues from the previous round. The fixes are in place, and the two follow-up commits do not introduce a blocker. CI and the Sonar quality gate are green.

src/main/services/emulators/emulator-souvenir-watcher.ts
```diff
@@ -0,0 +1,670 @@
...
+        game,
+        processId,
+        achievements,
+        souvenirs: newSouvenirs,
+        seenByDirectory,
+        clientId,
+      });
+      if (!screenshotPath) return;
+
+      let groupedAchievements = achievements;
```
Hachi-R, line 502: `src/main/services/emulators/emulator-souvenir-watcher.ts:L482` — P2: Reconciliation can add an achievement whose RetroArch screenshot was not present in `newSouvenirs` yet. That later file remains unseen, so the next poll queues a second souvenir for the same achievement group and leaves a terminal conflict requiring cleanup. Track reconciled achievement IDs for the session, or rescan and remove matching source files before releasing `isProcessing`.

Hachi-R (approved):
The delayed RetroArch screenshot case is now covered by the handled achievement set, and the regression test matches the runtime path. No blocker remains. CI and the Sonar quality gate are green.

PR #2704, opened 2026-08-14
fix(catalogue): guard the remaining repack.uris call sites

sotik11:
## Summary

Follow-up to #2451 and continuation of #2514. That PR was closed after its final review; the branch could not be reopened after the last push (GitHub kept the PR head stale and refused reopen), so this PR carries the complete fix. It guards every remaining unguarded `repack.uris` read in the download flow.

`RepacksModal`'s availability guard renders a malformed repack as "offline", but the row stays clickable — opening `DownloadSettingsModal`, which iterates `repack.uris` and crashes the renderer the same way #2451 prevented. The desktop and Big Picture download paths also share the same hoster-availability helpers, so one malformed row breaks availability enrichment for the whole list.

Guarded call sites:

- `game-details.tsx` — `selectRepackUri`
- `download-settings-modal.tsx` — `selectedUri` and `downloadOptions` memos
- `big-picture/src/helpers/downloaders.ts` — `getDownloaderAvailabilityOptions`
- `shared/index.ts` — `collectHosterAvailabilityUris` and `applyHosterAvailability` (the piece flagged in the final review of #2514)

## The shared helpers (addresses @Hachi-R's review on #2514)

The `getDownloaderAvailabilityOptions` guard alone is never reached for a repack whose `uris` is missing. `big-picture/src/hooks/queries/use-game-download-options.hook.ts` first calls `fetchHosterAvailability` → `collectHosterAvailabilityUris`, which iterates `repack.uris` with no array check; that throw is caught by `fetchDownloadOptions`, and `setNoDownloadOptionsState` then clears every repack — so one malformed row removes all Big Picture download options. The desktop `game-details.context` takes the same shared path (`fetchHosterAvailability` / `applyHosterAvailability`).

Guarding both shared helpers normalizes the shape once at the chokepoint: it fixes both consumers and makes the downstream `getDownloaderAvailabilityOptions` guard reachable. I traced the full `fetch → transform → consumer` path for both consumers; these two helpers were the only remaining unguarded reads.

## Fix

Same defensive shape as #2451 — `Array.isArray(repack.uris)` before every touch. In `selectRepackUri` the non-null assertion is dropped and the undefined case routes to a toast (`invalid_repack`) instead of sending an `undefined` uri into the download path. `invalid_repack` is added to every locale that already ships the sibling `downloader` key (en/ru/es/pt-BR/uk/pl/de/fr).

## Test plan

- `yarn typecheck` (node + web) — green
- Local build: a repack whose `uris` is malformed shows as offline with every downloader greyed out — no renderer crash, and Big Picture keeps the other repacks' download options instead of clearing all of them.

Closes the loop on #2514.


Hachi-R (approved):
Looks good. This closes the loop from #2514: the shared hoster-availability helpers were the remaining crash path, and guarding them once fixes both desktop and Big Picture consumers.

`selectRepackUri` dropping the non-null assertion and returning `{ ok: false, error: "invalid_repack" }` is the right call-site check. Toast wiring through `t(response.error)` in `game_details` matches the new key.

Nit: `invalid_repack` is only in en/ru/es/pt-BR/uk/pl/de/fr. Other locales fall back to English. Fine for merge; a follow-up can fill the rest.

PR #2705, opened 2026-08-14
fix: stop hanging big picture on a failed shop details fetch

cybx:
Follow-up from https://github.com/hydralauncher/hydra/pull/2639#discussion_r3722832843, taking the gate and the retry as the design.

Stacked on #2639 — the first three commits are that PR; the two to review here are the last ones.

- `6c91af8` drops the loading gate to plain `isLoading`. Every consumer downstream is null-safe after #2639, so a library game still opens with the API down instead of sitting on Loading forever.
- `b96159f` tracks the fetch failure in the hook and renders an error state with a Retry button when `!game && !shopDetails && shop !== "custom"` — the catalogue case that would otherwise render an empty shell ("Download Game" title, catalogue actions, silent Add to Library). Library and custom stay on the normal page, and the `handleAddToLibrary` guard is untouched; loosening it would add a library row titled "Download Game".

The failure flag covers all three ways into the state from the thread: the swallowed `.catch` on `getGameShopDetails`, plus `getGameStats` and `getGameAssets` rejections that previously escaped through the `finally`. Retry re-runs `refreshGameDetails` with the loading state, and the flag resets when a loading fetch starts, so retrying shows Loading rather than the stale error.

### DIFF
--- a/src/big-picture/src/components/pages/game/controller-support/index.tsx
+++ b/src/big-picture/src/components/pages/game/controller-support/index.tsx
@@ -11,7 +11,7 @@ import PlayStationLogo from "@renderer/assets/PlayStation Logo Wordmark.svg?reac
 
 export interface ControllerSupportBoxProps {
   shop?: GameShop;
-  shopDetails: ShopDetails;
+  shopDetails: ShopDetails | null;
   focusId?: string;
   focusNavigationOverrides?: FocusOverrides;
   focusNavigationOrder?: number;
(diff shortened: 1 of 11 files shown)


Hachi-R:
Closing this.

A failed shop-details fetch no longer hangs Big Picture: `useGameDetails` on `main` already sets `buildFallbackShopDetails` when the catalogue payload is null, so `!shopDetails` does not pin the page on Loading.

This branch is a conflicted stack of #2639 plus a Retry screen. #2639 is being asked to rebase on its own. If you still want Retry for catalogue-only failures (empty "Download Game" shell when the API is down), open a new PR on current `main` with in-flight request isolation — do not restack #2639.

PR #2713, opened 2026-08-18
fix: replace console statements with logger in main process

protonnegativo:

## Summary

Replace stray `console.log`, `console.error`, and `console.warn` calls with the project's structured `logger` utility (`electron-log`) in the main process, as required by [`.cursorrules`](https://github.com/hydralauncher/hydra/blob/main/.cursorrules#L3-L14).

## Problem

Several main process files bypass the project's logging infrastructure by using raw `console.*` statements. This means:

- Log output is not written to the structured log files (`logs.txt`, `error.txt`, etc.)
- One instance (`game-files-manager.ts`) fires `console.log` on **every extraction progress tick**, flooding stdout with thousands of lines
- Error/warning messages in `system-path.ts` and `delete-temp-file.ts` are invisible to the log file system

## Changes

| File | Change |
|------|--------|
| `get-available-drives.ts` | Remove 3 debug `console.log` calls, replace `console.error` → `logger.error` |
| `game-files-manager.ts` | Remove debug `console.log` that spams on every extraction tick |
| `system-path.ts` | Replace `console.error` → `logger.error` (already imports `logger`) |
| `delete-temp-file.ts` | Replace `console.warn` → `logger.warn` |

**4 files changed, 7 insertions, 12 deletions.**

## Why

The project's `.cursorrules` explicitly states:

> **Always use `logger` instead of `console` for logging** in both main and renderer processes

These changes enforce that guideline with zero behavior change — logs are simply routed to the proper `electron-log` infrastructure instead of raw stdout/stderr.

### DIFF
--- a/src/main/events/library/get-available-drives.ts
+++ b/src/main/events/library/get-available-drives.ts
@@ -1,4 +1,5 @@
 import fs from "node:fs/promises";
+import { logger } from "@main/services";
 import { registerEvent } from "../register-event";
 
 interface DriveInfo {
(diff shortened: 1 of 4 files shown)


Hachi-R (approved):
LGTM. These four files are the last `console.*` calls in `src/main`. The real win is dropping the per-tick `handleProgress` log during extraction; converting that to `logger` would just move the spam into `logs.txt`.

The `getAvailableDrives` debug logs are leftover noise — deleting them is better than routing them through the logger. `system-path` / `delete-temp-file` error paths now hit `electron-log`, which matches the rest of main.

Sonar Quality Gate is red on new-code reliability. Not a merge blocker — those catch paths already returned `[]` / `""` before this PR.

PR #2736, opened 2026-08-25
feat(souvenirs): handle collection limit

Moyasee:

**Fill in the PR content:**

## Summary

- handle the 1,000-souvenir API error without blocking achievement synchronization
- show a neutral collection-limit notice on the owner profile
- add English, Spanish, Brazilian Portuguese, and Russian copy

## Testing

- yarn typecheck
- ESLint on changed files
- focused souvenir retry and message tests, 12 passed
- full test suite, 498 passed and 2 cloud-save custom-path tests failed under external worktree path checks


Moyasee:
@greptile

src/renderer/src/context/user-profile/user-profile.context.tsx
```diff
@@ -369,6 +375,7 @@ export function UserProfileContextProvider({
       current.filter((souvenir) => getSouvenirKey(souvenir.id) !== key)
     );
     setSouvenirsTotal((current) => Math.max(0, current - 1));
+    setHasReachedSouvenirLimit(false);
```
Hachi-R: `src/renderer/src/context/user-profile/user-profile.context.tsx:L378` — P2: `hasReachedLimit` is true for any collection at or above 1,000, so a pre-existing user with 1,001+ souvenirs remains capped after deleting one, but this forces the UI to say the limit is cleared. Re-fetch the server flag after deletion (or return it from the delete endpoint) instead of setting `false` unconditionally.

Hachi-R (changes requested):
The deletion flow can clear the limit notice while the server still rejects new souvenirs.

Hachi-R (approved):
Previous P2 is fixed. No new findings.

PR #2739, opened 2026-08-28
feat: scan additional drive letters (E-I) for Games and SteamLibrary …

luis128-cpu:
…folders

## What
Adds drive letters E: through I: to the default `SCAN_DIRECTORIES` list used
by the installed-games scanner, for both:
- `\Games`
- `\SteamLibrary\steamapps\common`

## Why
The current defaults only check `C:` and `D:`. Users with more than two
drives/partitions dedicated to games (common with multi-SSD or SSD+HDD
setups) won't get their libraries auto-detected on those extra drives
unless they add each folder manually every time.

## Testing
- Ran the project locally and confirmed no TypeScript errors were
  introduced by this change.

### DIFF
--- a/src/main/events/library/scan-installed-games.ts
+++ b/src/main/events/library/scan-installed-games.ts
@@ -28,9 +28,19 @@ import type { Game, GameShop } from "@types";
 const SCAN_DIRECTORIES = [
   String.raw`C:\Games`,
   String.raw`D:\Games`,
+  String.raw`E:\Games`,
+  String.raw`F:\Games`,
+  String.raw`G:\Games`,
+  String.raw`H:\Games`,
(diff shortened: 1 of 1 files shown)


Hachi-R:
Closing this.

Hardcoding `E:\Games` through `I:\Games` (and the same for `SteamLibrary`) does not scale past `J:` and is the wrong layer. Steam libraries are already discovered dynamically via `getSteamLibraryFolders()` / `libraryfolders.vdf`. Extra `\Games` folders should come from mounted drives (`get-available-drives`), not a longer letter list.

If you still want non-Steam `?:\Games` coverage, reopen with drive enumeration instead of more hardcoded paths.

PR #2742, opened 2026-09-01
feat: add option to hide adult content in library

LeJeunek:
[prdescription.md](https://github.com/user-attachments/files/31670101/prdescription.md)


LeJeunek:
Thanks for the review — both addressed in 688a8c2.

*Missing metadata bypasses filtering:* Fixed. getLibrary now queues a
background classification for any Steam game still missing a contentWarning
when hideAdultContent is enabled: throttled to 2 concurrent requests,
capped at 20 games per call, and deduped against in-flight requests. It
drains the backlog gradually across normal library refreshes instead of
bursting requests at Steam's unauthenticated storefront API, and costs
nothing for users who haven't turned the preference on.
shouldHideGameForAdultContent itself is intentionally unchanged — a missing
warning should never be treated as "adult" — but the window where that's
true is now transient instead of permanent.

While in there, I also noticed get-game-shop-details.ts was only persisting
a warning when content_descriptors.ids.length > 0, so a genuinely clean
game looked identical to "not yet checked" and would keep getting
re-queried indefinitely. Fixed to persist an explicit "none" as well, and
extracted the shared persistContentWarning helper so both the reactive
(Game Details opened) and proactive (library backfill) paths write through
the same code.

*Adult age remains a literal:* Extracted to STEAM_ADULT_ONLY_MINIMUM_AGE,
with a comment noting it's informational (the minimumAge field) rather than
the actual filtering threshold — shouldHideGameForAdultContent keys off
level, not age.

All checks re-verified locally (test, typecheck, lint, format-check).

LeJeunek:
Thanks for the review — both addressed in 688a8c2.

Missing metadata bypasses filtering: Fixed. getLibrary now queues a background classification for any Steam game still missing a contentWarning when hideAdultContent is enabled: throttled to 2 concurrent requests, capped at 20 games per call, and deduped against in-flight requests. It drains the backlog gradually across normal library refreshes instead of bursting requests at Steam's unauthenticated storefront API, and costs nothing for users who haven't turned the preference on. shouldHideGameForAdultContent itself is intentionally unchanged — a missing warning should never be treated as "adult" — but the window where that's true is now transient instead of permanent.

While in there, I also noticed get-game-shop-details.ts was only persisting a warning when content_descriptors.ids.length > 0, so a genuinely clean game looked identical to "not yet checked" and would keep getting re-queried indefinitely. Fixed to persist an explicit "none" as well, and extracted the shared persistContentWarning helper so both the reactive (Game Details opened) and proactive (library backfill) paths write through the same code.

Adult age remains a literal: Extracted to STEAM_ADULT_ONLY_MINIMUM_AGE, with a comment noting it's informational (the minimumAge field) rather than the actual filtering threshold — shouldHideGameForAdultContent keys off level, not age.

All checks re-verified locally (test, typecheck, lint, format-check).

PR #2745, opened 2026-09-02
Feat/review discoverability and author banners

galinhaman22:

**Fill in the PR content:**

Custom banner support for reviews and replies, reworked reviews callout, and custom banner tweaks on Profile and Friends window.


src/renderer/src/pages/game-details/use-review-prompt.ts
```diff
@@ -0,0 +1,87 @@
...
+  shop?: GameShop;
+  objectId?: string;
+  playTimeInMilliseconds: number;
+  userDetailsId?: string;
+  isGameInLibrary: boolean;
+  hasUserReviewed: boolean;
+  isCheckingUserReview: boolean;
+}
+
+const REVIEW_PROMPT_DISMISSED_STORAGE_KEY = "reviewPromptDismissedAt";
```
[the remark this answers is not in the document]
galinhaman22: Intentional, not a regression. The requirement for this branch is that "Maybe later" hides the discovery prompt for every eligible game for a week, then it comes back everywhere — the per-game key was the previous behaviour and was changed deliberately. Keeping the single global timestamp.

src/renderer/src/pages/game-details/review-thread.tsx
```diff
@@ -309,14 +311,23 @@ export function ReviewThread({
...
+              <button
+                className="game-details__reply-action-link"
+                onClick={() => handleReplyTo("")}
+                aria-disabled={!canReply}
+                data-tooltip-id={canReply ? undefined : replyTooltipId}
+                data-tooltip-place="top"
+                data-tooltip-content={t("reply_requires_playtime", {
+                  count: REPLY_MIN_PLAYTIME_IN_HOURS,
+                })}
+              >
```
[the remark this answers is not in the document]
galinhaman22: The reply link only renders when `showInlineReply` is true, and that requires `Boolean(userDetailsId)`, so a signed-out user never sees this button or its tooltip. When it does render the user is signed in by construction, so the only way `canReply` is false is insufficient playtime and no existing review — which is exactly what the tooltip says. No change needed here.

galinhaman22:
@greptile

src/renderer/src/pages/game-details/review-thread.tsx
```diff
@@ -249,23 +261,17 @@ export function ReviewThread({
...
       });
 
-      if (response) {
-        setReplies((prev) => mergeReplies(prev, [response]));
-        setTotalCount((prev) => prev + 1);
-        setHidden(false);
-        setExpanded(true);
-      }
+      setHidden(false);
+      await fetchReplies(0, true);
```
Moyasee: P2: Posting a reply replaces every loaded reply with only the first page. Threads with more than 10 replies lose already-loaded pages, and the new reply may not be visible if it falls outside page zero. Merge the POST result or fetch the created reply without replacing the existing list.

src/renderer/src/hooks/use-processed-image.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    let cancelled = false;
+
+    setProcessedImageUrl(
+      processedImageCache.get(getCacheKey(imageUrl, { width, height })) ??
+        imageUrl
+    );
+
+    requestProcessedImage(imageUrl, { width, height }).then((resolved) => {
```
Moyasee, line 108: P2: Every mounted review and reply starts image processing immediately. One page can launch roughly 120 concurrent downloads and native resizes, including animated banners, which can cause heavy CPU and network use. Add a shared concurrency limit or defer processing until each card is visible.

src/renderer/src/pages/game-details/review-thread.tsx
```diff
@@ -249,23 +261,28 @@ export function ReviewThread({
...
-        setTotalCount((prev) => prev + 1);
-        setHidden(false);
-        setExpanded(true);
-      }
+      const params = new URLSearchParams({
+        take: REPLIES_TAKE.toString(),
+        skip: totalCount.toString(),
+      });
+
+      const response = await electron.hydraApi.get<
```
Moyasee, line 302: P2: If the reply POST succeeds but the follow-up GET fails, the UI reports submission failure and leaves the composer open. Retrying creates a duplicate reply. Treat the POST as the submission boundary and handle refresh failure separately.

Moyasee (approved):
lgtm

src/renderer/src/pages/game-details/review-thread.tsx
```diff
@@ -249,30 +261,44 @@ export function ReviewThread({
...
+        { answers: GameReviewAnswer[]; totalCount: number } | undefined
+      >(`${baseUrl}?${params.toString()}`, { needsAuth: false });
 
       if (response) {
-        setReplies((prev) => mergeReplies(prev, [response]));
-        setTotalCount((prev) => prev + 1);
-        setHidden(false);
-        setExpanded(true);
+        setTotalCount(response.totalCount);
+        setReplies((prev) => mergeReplies(prev, response.answers));
```
Moyasee, line 308: P2: Posting in a partially loaded thread appends the new tail reply after the loaded preview. Later Load more pages are appended after that reply because mergeReplies preserves existing order, so the new reply moves into the middle of long threads. Sort merged replies by createdAt, or insert newly loaded middle pages before known tail replies.

src/renderer/src/pages/game-details/use-user-review-status.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function useUserReviewStatus({
+  shop,
+  objectId,
+  userDetailsId,
+}: Readonly<UseUserReviewStatusParams>) {
+  const [hasUserReviewed, setHasUserReviewed] = useState(false);
+  const [isCheckingUserReview, setIsCheckingUserReview] = useState(true);
+
+  useEffect(() => {
```
Moyasee, line 29: P2: When userDetailsId, shop, or objectId changes, this hook returns the previous completed status for one render because the effect resets isCheckingUserReview only after rendering. On sign-in, the review form can briefly appear before discovering that the user already reviewed the game. Track the identity associated with the result and report checking whenever it differs from the current identity.

Moyasee (approved):
LGTM

PR #2747, opened 2026-09-02
Feat/cloud gift

Hachi-R:

**Fill in the PR content:**


src/renderer/src/pages/shared-modals/cloud-gift-notification-modal.tsx
```diff
@@ -0,0 +1,551 @@
...
+                      count: gift.durationMonths,
+                    })}
+                  </h2>
+
+                  {gift.message && (
+                    <div
+                      ref={messageRef}
+                      id={messageId}
+                      className="cloud-gift-notification-modal__message-card"
+                      dangerouslySetInnerHTML={{ __html: gift.message }}
```
Moyasee: [P1] `gift.message` is sender-controlled but is inserted as raw HTML. The launcher CSP allows inline handlers, and this page exposes privileged `window.electron` methods, so a malicious gift message can execute code in the renderer. Pass the message through `sanitizeHtml` before assigning `__html`, and make the same change in `src/big-picture/src/components/cloud-gift-notification-modal.tsx:558`.

src/renderer/src/app.tsx
```diff
@@ -665,6 +666,8 @@ export function App() {
         feature={hydraCloudFeature || undefined}
       />
 
+      <CloudGiftNotificationModal />
```
Moyasee, line 669: [P2] The main renderer and Big Picture each mount an independent gift modal. Opening Big Picture hides the main window but does not unload it, so both can cache the same pending gift; accepting or dismissing it in Big Picture leaves a stale modal that reappears when the user returns to the main window. Gate polling to the active window or broadcast gift acceptance and dismissal so every mounted modal clears its state; the second mount is at `src/big-picture/src/app.tsx:163`.

src/renderer/src/pages/shared-modals/cloud-gift-notification-modal.tsx
```diff
@@ -0,0 +1,523 @@
...
+        .get<CloudGiftDetails>(`/cloud-gifts/${giftId}`, {
+          needsAuth: true,
+        })
+        .then((giftDetails) => {
+          if (giftDetails.status === "PENDING_ACCEPTANCE") {
+            setGift(giftDetails);
+            return;
+          }
+
+          setNotification(null);
```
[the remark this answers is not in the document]
Moyasee: [P1] This is still reproducible on `64b0766`. `findPendingGift` can overlap with `onOpenGiftModal`, but it neither records nor checks `activeGiftIdRef`. If its `setNotification(item)` runs between the explicit request setting its notification and gift details, the modal displays the explicit gift while `acceptGift` uses the auto-discovered notification ID. Guard the automatic request with the same gift or request token before resolving this thread.

src/renderer/src/app.tsx
Moyasee: [P2] The new broadcast fixes acceptance, but not dismissal. `dismissCurrentGift` still clears only local state, so choosing "Decide later" in Big Picture leaves the cached main-window modal to reappear when Big Picture closes. Broadcast dismissal too, or prevent the hidden window from polling.

Moyasee: [P2] Still present on 8e8f2a9. This commit only changes request guards; dismissCurrentGift at renderer line 263 and Big Picture line 284 still clears local state only. Choosing "Decide later" in Big Picture leaves the hidden main window modal cached, so it reappears after returning. Broadcast dismissal or stop the hidden window from polling.

src/renderer/src/pages/shared-modals/cloud-gift-notification-modal.tsx
```diff
@@ -0,0 +1,576 @@
...
+
+        const giftDetails = await window.electron.hydraApi
+          .get<CloudGiftDetails>(`/cloud-gifts/${giftId}`, {
+            needsAuth: true,
+          })
+          .catch(() => null);
+
+        if (
+          giftDetails?.status === CLOUD_GIFT_STATUS_PENDING_ACCEPTANCE &&
+          !activeGiftIdRef.current
```
Moyasee: [P2] L168: This guard does not recheck dismissedGiftIdsRef after the request. If a manual open resolves and the user chooses "Decide later" while this automatic lookup is still pending, dismissCurrentGift records the ID and clears activeGiftIdRef; the stale response then passes this condition and immediately reopens the same gift. Recheck dismissedGiftIdsRef.current.has(giftId) here before setting either state value; the Big Picture copy has the same race at line 178.

Moyasee: [P2] L168: This guard does not recheck `dismissedGiftIdsRef` after the request. If the user dismisses a manually opened gift while this lookup is pending, the stale response sees a cleared `activeGiftIdRef` and immediately reopens the same gift. Recheck the dismissed ID before setting `gift` or `notification`; apply the same fix to the Big Picture copy.

src/renderer/src/app.tsx
Moyasee, line 669: [P2] L669: The main renderer and Big Picture mount separate modal instances, but `dismissCurrentGift` only clears the instance where the user acts. Choosing "Decide later" in Big Picture leaves the hidden main-window modal cached, so it reappears when the user returns. Broadcast dismissal to every app window or stop the hidden window from polling.

src/renderer/src/pages/shared-modals/cloud-gift-notification-modal.scss
```diff
@@ -0,0 +1,360 @@
...
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+  }
+
+  &__accept {
+    min-height: 40px;
+    padding: 10px 16px;
+    color: #fff;
```
Moyasee, line 225: [P2] L225: White text on this bright cyan gradient does not meet minimum contrast, so the accept label is difficult to read for low-vision users. Use a dark text color or darken the gradient to meet WCAG AA, and apply the same fix to the Big Picture button.

src/big-picture/src/components/cloud-gift-notification-modal.tsx
```diff
@@ -0,0 +1,692 @@
...
+    ease: [scaleEase, "linear" as const],
+  },
+  y: {
+    duration: OPEN_ANIMATION_DURATION,
+    times: RING_MOVE_TIMES,
+    ease: ["linear" as const, figmaFirmSpring, "linear" as const],
+  },
+});
+
+export function CloudGiftNotificationModal() {
```
Moyasee, line 119: [P2] L119: This component duplicates the renderer modal controller and most visual logic. Sonar now fails the quality gate with 11.2% duplication, and each race fix has required matching edits in both files. Extract the shared gift state and request handling into one hook, plus common animation helpers where practical, while keeping Big Picture navigation code here.

Moyasee (approved):
lgtm

PR #2752, opened 2026-09-03
refactor(cloud-gift): simplify Cloud Gift notification modal and upda…

Hachi-R:


Moyasee (commented):
P2 src/renderer/src/pages/shared-modals/cloud-gift-notification-modal.tsx:L317: Removing the keydown effect also removes its cleanup that restored `previouslyFocusedElement`, so every remaining close path now leaves keyboard focus on an element that has been unmounted. Keep a visibility-scoped effect that captures the previous focus and restores it during cleanup, without registering the Escape handler. Add a focused test for closing through accept or the sender link.

Moyasee (commented):
P1 src/big-picture/src/components/cloud-gift-notification-modal.tsx:L371: Removing `press.b` does not disable back input. `triggerBackAction` now skips this modal registration, invokes a lower screen handler or `navigate(-1)`, and leaves the full-screen gift overlay mounted over the changed route. Keep a no-op `press.b` action while the modal is visible so controller B and Escape are consumed without dismissing it. Add coverage that back input neither closes the gift nor changes the underlying route.

Moyasee (commented):
P1 src/big-picture/src/components/cloud-gift-notification-modal.tsx:L374: The no-op `press.b` handler does not reliably block back input because this global component creates its screen-action registration once at app mount. Any route mounted later sits above it in `NavigationScreenActionsService.stack`; for example, `game-achievements` registers its own `press.b`, so B or Escape can navigate the obscured route while the gift stays open. Register this action only while the gift is visible, such as through a child component mounted with the modal, so it is pushed above existing page handlers. Add a regression test where a later page-level B action is not called while the gift modal is open.

Moyasee (approved):
LGTM

PR #2756, opened 2026-09-04
Feat: Support Big Picture deeplink

Wiwok:

**Fill in the PR content:**
This PR adds a deeplink for Big Picture mode. This allows software such as the Moonlight/Sunshine combo to launch Hydra directly in Big Picture mode.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -291,11 +291,7 @@ const handleDeepLinkPath = (uri?: string) => {
     }
 
     if (url.host === "bigpicture") {
-      if (WindowManager.mainWindow) {
-        WindowManager.openBigPictureWindow();
-      } else {
-        WindowManager.createMainWindow({ forceBigPicture: true });
-      }
+      WindowManager.openBigPictureWindow();
       return;
     }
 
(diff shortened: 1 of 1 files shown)


Hachi-R (changes requested):
Requesting changes, and we should not merge this as-is.

Hydra already starts in Big Picture with `--big-picture`. First instance reads `process.argv.includes("--big-picture")` and calls `createMainWindow({ forceBigPicture: true })`; a second instance does the same on `commandLine` and opens the Big Picture window. Sunshine/Moonlight can launch `Hydra.exe --big-picture` (or the Linux equivalent) without a protocol handler.

The deeplink also does not match that path on cold start, which is the case this PR is for. `hydralauncher://run` skips the main window; `hydralauncher://bigpicture` does not. `createMainWindow()` is still called without `forceBigPicture`, so `ready-to-show` `show()`s the desktop UI and races `handleDeepLinkPath` → `openBigPictureWindow()`. Second-instance handling is closer, but it does not make the first-launch path correct.

If a `hydralauncher://bigpicture` link is still needed for hosts that can only open URLs, reopen with cold start wired like `--big-picture` (`forceBigPicture: true`), and keep `handleDeepLinkPath` for a running instance / macOS `open-url`.

Hachi-R:
Closing this. Starting Hydra in Big Picture is already supported with `--big-picture` (same flag on a second instance if the app is already running).

If you still need a `hydralauncher://bigpicture` deeplink — for something that can only open a URL, not a process argument — please reopen with the cold-start issues from the review fixed so the deeplink uses `forceBigPicture` the same way as the CLI flag.

PR #2760, opened 2026-09-04
Feat/4.1.2 gift retrocompatibility

Hachi-R:
gift retrocompatibility


src/renderer/src/pages/shared-modals/cloud-gift-notification-modal.tsx
```diff
@@ -0,0 +1,451 @@
...
+      }
+
+      setGift(null);
+      setNotification(requestedNotification);
+
+      void window.electron.hydraApi
+        .get<CloudGiftDetails>(`/cloud-gifts/${giftId}`, {
+          needsAuth: true,
+        })
+        .then((giftDetails) => {
```
chubbygrannychaser, line 137: This response is not tied to the currently selected notification. If request A is still pending when the user opens B, A can resolve here and set A's details while `notification` remains B; clicking Accept then posts to B's ID. The stale non-pending/failure paths also dismiss B or open A's external page. Track the active gift ID (or cancel/ignore prior requests) and only update/redirect when it still matches.

chubbygrannychaser (changes requested):
## Summary

The Cloud Gift modal does not guard asynchronous detail responses against a newer modal-open event. This can present one gift while accepting another.

## Critical Issues (Must Fix)

- Verify the requested gift remains current before every response-side state change, redirect, or dismissal.

## Validation

- `yarn typecheck:node` — passed
- `yarn typecheck:web` — passed
- `yarn format-check` — passed

src/renderer/src/pages/shared-modals/cloud-gift-notification-modal.tsx
```diff
@@ -0,0 +1,459 @@
...
+
+    const giftId = notification.variables[CLOUD_GIFT_ID_VARIABLE];
+    setIsAccepting(true);
+
+    try {
+      await window.electron.hydraApi.post(`/cloud-gifts/${giftId}/accept`, {
+        needsAuth: true,
+      });
+
+      dismissCurrentGift();
```
Moyasee, line 195: [P2] Guard acceptance completion against a newer gift. Start accepting gift A, select Decide later while the request is pending, then open gift B. When A finishes, this unconditional dismissal closes B and opens A's follow-up page. The same happens in src/big-picture/src/components/cloud-gift-notification-modal.tsx:248. Check that activeGiftIdRef.current still matches the accepted gift before dismissing or redirecting, or prevent dismissal/replacement while acceptance is pending. Add a regression test for this sequence.

src/renderer/src/pages/settings/settings-account.tsx
```diff
@@ -260,6 +267,26 @@ export function SettingsAccount() {
...
+        render={({ field }) => (
+          <section className="settings-account__section">
+            <h3>{t("cloud_gifts")}</h3>
+            <CheckboxField
+              checked={field.value ?? true}
+              disabled={isSubmitting}
+              label={t("allow_cloud_gifts")}
+              onChange={(event) => {
+                field.onChange(event.target.checked);
+                void handleSubmit(onSubmit)();
```
Moyasee, line 282: [P2] Restore the gift preference when saving fails. The checkbox changes before the save completes, but a rejected patchUser leaves the new value displayed and produces an unhandled rejection with no error feedback. For example, disabling gifts while offline leaves the checkbox off even though the saved preference still allows gifts. Catch the failure, restore the last saved value, and show an error toast. Add coverage for a rejected preference save.

Moyasee (approved):
LGTM. No P0 or P1 findings. The earlier stale-detail response issue is fixed. Two non-blocking P2 findings are attached.

Web and node typechecks passed. React/jsdom checks against the shared modal code confirmed the stale-detail guards and reproduced both P2 issues with mocked IPC. These flows have no regression tests in this PR. Packaging checks are still running; SonarCloud is failing.

PR #2767, opened 2026-09-06
docs: propose optional per-game shader cache

Polabiel:
Proposal only, no code changes. Motivated by Ready or Not (1144200) on Fedora + GTX 1650 + UMU-Proton: 5-20 min black screen at first launch while VKD3D compiles hundreds of PSOs (healthy process, growing proton log, swapchain created). Proposes an opt-in per-game shader cache (persist DXVK/VKD3D/NVIDIA caches, optionally fetch a verified cache at install) with stale-cache fallback, disk/trust safeguards. Details in docs/proposals/optional-shader-cache.md.


Hachi-R:
Closing this.

A docs-only proposal does not belong as a PR. There is no `docs/proposals/` convention in this repo, and the write-up is not something we can merge or act on as-is (it also assumes prefix cleanup that does not exist). If this is worth doing, open an issue with a concrete implementation, not a design doc on `main`.

PR #2790, opened 2026-09-16
fix(steam): restore actively imported games

Hachi-R:
## Problem

A game already added through Hydra can keep `source: hydra` after Steam imports it. If the user removes it locally, a later Steam sync leaves it hidden even though the Steam import is active.

## Change

- Treat `hasActiveSteamImport` as authoritative when resolving local deletion state.
- Apply the same rule to full-library and single-game merges.
- Preserve the existing Hydra/Steam playtime merge and source provenance.

## Validation

- Library source/deletion and playtime tests: 12 passed.
- Typecheck: passed.
- Prettier: passed.
- ESLint: 0 errors; existing warnings only.

Stacked on #2789.

PR #2791, opened 2026-09-16
fix(steam): offer reconnect for stale sessions

Hachi-R:
## Problem

When the local Steam session expires or belongs to another account, sync asks the user to sign in again but Settings still offers the same failing Sync action. A startup failure can also happen before Settings mounts, so renderer-only state is insufficient.

## Change

- Keep a process-local `requiresReconnect` flag in the Steam sync state.
- Surface Reconnect and the existing session message instead of Sync.
- Preserve the flag across renderer mounts and clear it after successful OAuth or Hydra logout.
- Treat account mismatch like an expired session because both require local Steam authentication.

## Validation

- Entire Steam service suite: 117 passed.
- Typecheck: passed.
- Prettier: passed.
- ESLint: 0 errors; existing warnings only.

Stacked on #2790.


chubbygrannychaser (changes requested):
Changes requested on head `ea96dd5117a53b23de509b22b0badecac7a37a05`.

Blocking finding:

- `src/renderer/src/pages/settings/settings-steam.tsx:168-173`: `tryAutoStart` copies a persisted `{ status: "idle", requiresReconnect: true }` state, then calls `shouldAutoStartSteamSync` with only `localState.status === "idle"`. The helper therefore returns `true` for a connected account with no prior successful sync, even when the process-local state explicitly requires reconnect. A startup `steam-session-required` or `steam-account-mismatch` failure can be retried automatically whenever Settings mounts or remounts, instead of preserving the failure state and presenting only Reconnect. Exclude `requiresReconnect` from the auto-start eligibility, or pass that state into the helper and add a regression test covering this startup/remount sequence.

Validation completed:

- Focused Steam test set: 128 passed.
- `node --import ./scripts/register-ts-node.mjs --test src/shared/steam-session.test.ts`: 2 passed.
- `corepack yarn typecheck:node`: passed.
- `corepack yarn typecheck:web`: passed.
- Changed-file ESLint: 0 errors, 2 existing warnings in `src/shared/index.ts`.
- Changed-file Prettier check: passed.
- `git diff --check`: passed.

The remaining OAuth success reset, Hydra logout reset, renderer state handoff, UI Reconnect action, and stacked scope were inspected and showed no additional blocking issue.

chubbygrannychaser (changes requested):
Changes requested on current head `a60b9336269d2679327b4efe81f69a0a2e13566a`.

Blocking finding:

- `src/main/services/steam-integration/steam-store-session.ts:161-168`: the embedded OAuth success handler calls `notifySteamConnected()` directly, so it never reaches the new `handleDeepLinkPath()` branch in `src/main/index.ts:329-337`, where `steamSyncOrchestrator.clearReconnectRequired()` is called. After a sync fails with `steam-session-required` or `steam-account-mismatch`, the main-process state remains `{ status: "idle", requiresReconnect: true }` even after successful embedded reconnect. The renderer clears its local state, then `tryAutoStart()` reads the stale main-process state and declines auto-start, leaving the account stuck showing Reconnect instead of Sync. Clear the reconnect flag in the shared embedded success path, or route embedded success through the same reset logic, and add a regression test for embedded OAuth success after a reconnect-required failure.

The prior auto-start regression is fixed: `requiresReconnect` is now passed to `shouldAutoStartSteamSync`, and the focused test covers the retry block.

Validation completed on this head:

- Focused Steam tests: 10 passed, 0 failed.
- `corepack yarn typecheck:node`: passed.
- `corepack yarn typecheck:web`: passed.
- Changed-file ESLint: 0 errors, 2 existing warnings in `src/shared/index.ts`.
- Changed-file Prettier check: passed.
- `git diff --check`: passed.
- Hosted `build` and `lint` checks were still pending when reviewed; `text-integrity` passed and `build-production` was skipped.

chubbygrannychaser (approved):
Re-reviewed the full PR at current head `09086474692747b2e6c792153c5be860ff9beaf6`.

The two prior blocking findings are fixed:

- Auto-start now receives and honors `requiresReconnect`, with regression coverage preventing retry after `steam-session-required` or `steam-account-mismatch`.
- Embedded-window OAuth success and external deep-link success both use `completeSteamOpenIdConnection`, which clears the reconnect requirement before notifying the renderer. The ordering regression test passes.

No additional blocking correctness, security, lifecycle, or data-integrity issues found.

Validation:
- Focused Steam tests: 22 passed, 0 failed.
- `corepack yarn typecheck:node`: passed.
- `corepack yarn typecheck:web`: passed.
- Changed-file ESLint: 0 errors; 2 existing warnings in `src/shared/index.ts`.
- `git diff --check`: passed.
- Hosted `build` and `lint` checks were pending at review time; `text-integrity` passed and `build-production` was skipped.

PR #2794, opened 2026-09-16
fix(steam): clear achievements after disconnect

Hachi-R:
## Problem

Deleting an active Steam integration removes imported achievement data on the API, but the launcher keeps the merged Steam unlocks in memory. An open game page can continue showing deleted Steam achievements until the cache is rebuilt.

## Change

- Invalidate achievement memory only for games with an active Steam import after destructive disconnect.
- Preserve cached achievements when the user disconnects without deleting the imported snapshot.
- Add targeted coverage for per-game cache invalidation.

## Validation

- Achievement memory tests: 8 passed.
- Node typecheck: passed.
- Full pre-push typecheck: passed.
- ESLint: 0 errors; existing warnings only.
- Full suite: 652 passed; 2 existing Cloud Save custom-path tests failed as on the stack base.

Stacked on `feat/steam-connection`.


chubbygrannychaser (approved):
## Review summary

Reviewed the complete diff against `feat/steam-connection` at head `18a27929d7f3a482543941257fed633b0e5a6fba`.

The per-game achievement cache invalidation is correctly scoped to destructive Steam disconnects and uses the same `(shop, objectId)` identity as the cache and local library records. The non-destructive disconnect path preserves the imported snapshot and its cache as intended. I found no blocking correctness or regression issues.

## Validation

- Focused achievement memory tests: 8 passed.
- Node typecheck: passed.
- `git diff --check`: passed.

Hosted Ubuntu, Windows, and lint checks were still pending at review time.

PR #2795, opened 2026-09-16
feat(steam): add per-game playtime opt-out

Hachi-R:
## Problem
Steam imports can use executables outside a Steam library. Users need a per-game choice to prevent Hydra from counting playtime for those games without changing other games.

## Changes
- adds a per-game playtime checkbox to the Game Settings modal for active Steam imports
- persists the preference locally per computer
- separates Hydra playtime counting from Steam exit-sync eligibility
- keeps automatic protection for executables inside Steam libraries
- ignores the local preference when the Steam import is no longer active

## Validation
- 22 focused Steam playtime and exit-sync tests pass
- node and web typechecks pass
- ESLint passes with existing warnings
- Prettier and `git diff --check` pass

Stacked on #2794.


chubbygrannychaser (changes requested):
## Summary

REQUEST_CHANGES on current head `435b3543d9f9f2cf14a69c3126a9e743f0710c57`.

The focused playtime and exit-sync tests pass, but the current lifecycle and persistence paths still have two blocking regressions.

## Blocking findings

1. **`src/main/services/process-watcher.ts:424-427`**

   The `/profile/games/steam/:objectId` lookup is still performed only inside `isSteamLibraryExecutablePath(...)`. If a user has changed a Steam game's executable to a path outside the registered Steam library, `game.hasActiveSteamImport` is never refreshed. After the Steam import is disconnected, a cached `true` therefore remains `true`, so `resolveSteamSessionPlaytimePolicy` continues to disable Hydra counting and the session never becomes eligible for the intended disconnected behavior. The UI also keeps showing the import-only setting from the stale persisted flag. Please resolve the current Steam import status for every Steam game session, or explicitly clear/revalidate the cached flag before applying the policy to non-library paths, and add an integration regression test for a cached active import followed by a disconnected non-library launch.

2. **`src/main/events/library/set-game-hydra-playtime-disabled.ts:15-20`**

   This handler reads the full game record and then writes it back with `{ ...game, disableHydraPlaytimeTracking: disabled }`. That read-modify-write is concurrent with `persistGamePlaytime` in `process-watcher.ts:262-276`, which uses the same full-record pattern. Toggling the checkbox while a game is running can therefore lose a playtime tick or lose the newly selected setting, depending on which write finishes last. Please serialize this update with the playtime persistence path or use an atomic field-level update, and add a concurrency/persistence test proving that toggling the setting preserves playtime and pending unsynced deltas.

## Validation

- Focused tests: `corepack yarn node --import ./scripts/register-ts-node.mjs --test src/main/services/steam-integration/steam-playtime.test.ts src/main/services/steam-integration/steam-game-exit-sync.test.ts src/main/services/game-running-state.test.ts` passed (22 tests).
- Typecheck: `corepack yarn typecheck` passed.
- Changed-file ESLint: passed with 18 pre-existing warnings and 0 errors.
- Changed-file Prettier check: passed.
- `git diff --check`: passed.
- Hosted checks were still in progress at review time: build on Ubuntu and Windows, lint, and Greptile Review. Text Integrity was successful; build-production was skipped.

src/main/services/process-watcher.ts
```diff
@@ -441,10 +445,19 @@ async function onOpenGame(game: Game, matchedPath: string) {
...
       game = { ...currentGame, hasActiveSteamImport };
       await gamesSublevel.put(gameKey, game);
-      countHydraPlaytime = !hasActiveSteamImport;
     }
   }
 
+  if (game.shop === "steam") {
+    ({ countHydraPlaytime, syncSteamOnExit } =
+      resolveSteamSessionPlaytimePolicy({
+        hasActiveSteamImport: game.hasActiveSteamImport === true,
```
chubbygrannychaser, line 452: This policy uses the cached `game.hasActiveSteamImport` value, but the server refresh above runs only inside the `isSteamLibraryExecutablePath(...)` branch. If this imported Steam game is launched from a non-library path and Steam is later disconnected, the cached value can remain true; with `disableHydraPlaytimeTracking` enabled, every subsequent session will keep `countHydraPlaytime` false indefinitely. Resolve the active-import status independently of the executable-path branch, or explicitly clear the cached flag on disconnect, and add a regression test covering a non-library path after a server-side disconnect.

chubbygrannychaser (changes requested):
## Summary

REQUEST_CHANGES for current head `435b3543d9f9f2cf14a69c3126a9e743f0710c57`.

The per-game setting is persisted and the Hydra and Steam playtime counters are separated, but the disconnect path has a blocking stale-state case. See the inline finding.

## Validation

- Focused Node tests: 22 passed, 0 failed.
- Node typecheck: passed.
- Web typecheck: passed.
- ESLint on changed JS, TS, and TSX files: 0 errors, 18 pre-existing warnings.
- `git diff --check`: passed.

chubbygrannychaser (approved):
## Summary

APPROVE for current head `423377a4eeb46efbb6d6cd0cd28170c8771b57b2`.

The two blockers from the prior review are resolved. `onOpenGame` now revalidates the active Steam import for every Steam session, including launches outside registered Steam library paths, before applying the playtime policy. Playtime persistence, Steam-session status persistence, and the setting IPC now use the same per-game serialized read-modify-write updater, preventing concurrent writes from losing playtime or pending unsynced deltas.

## Validation

- Focused Node tests: 24 passed, 0 failed, including the new Steam disconnect and concurrent game-record update regressions.
- Node typecheck: passed.
- Web typecheck: passed.
- ESLint on changed JavaScript, TypeScript, and TSX files: 0 errors; 18 existing warnings.
- `git diff --check`: passed.
- Hosted checks were still in progress at review time for Ubuntu build, Windows build, and lint. Text Integrity was successful; build-production was skipped.

PR #2796, opened 2026-09-16
fix(achievements): preserve Steam unlocks on reset

Hachi-R:
## Problem
The Game Settings reset copy said Steam achievements would be re-imported, while the reset also cleared the local achievement cache before the server confirmed the operation.

## Changes
- explains that Steam manages its achievements and Hydra cannot delete them
- states that only Hydra achievements and their souvenirs are reset for Steam games
- applies the same behavior in desktop and Big Picture settings
- clears unlocked achievement memory only after the reset API succeeds, then reloads the preserved Steam unlocks
- serializes the local count update through the shared game-record updater

## Validation
- 12 focused achievement cache and merge tests pass
- node and web typechecks pass
- changed-file ESLint passes
- Prettier and `git diff --check` pass

Requires hydralauncher/hydra-api#601. Stacked on #2795.


chubbygrannychaser (changes requested):
## Summary
REQUEST_CHANGES for current head `e7a63489ca63142d923a54f7138d9ea92d8fedf0`.

## Blocking finding

- `src/main/events/library/reset-game-achievements.ts:49-71` — This desktop change assumes the Hydra API delete endpoint already preserves Steam-backed unlocks, but that behavior is only implemented in the still-open `hydralauncher/hydra-api#601`. Until that API change is deployed, `DELETE /profile/games/achievements/:id` still deletes all achievements for the game, so the new Steam copy and visibility text is false and the reset will still remove Steam achievements remotely. Please make this PR depend on and ship against hydra-api#601 (or otherwise gate the client behavior on the deployed API contract), then verify the endpoint behavior before merging.

## Validation

- `git diff --check` passed.
- Focused ESLint passed for the three changed TypeScript/TSX files.
- Prettier check passed for all seven changed files.
- `yarn typecheck:node` passed.
- `yarn typecheck:web` passed.
- `hydralauncher/hydra-api#601` is currently open, not merged; its changed endpoint is the implementation that preserves Steam unlocks.

chubbygrannychaser (approved):
Reviewed the exact head `ceeb6579a7d03815689eb28d2cc915252afde943`.

No blocking issues remain. The reset handler now preserves the Steam achievement cache, clears Hydra-owned state after the API reset succeeds, and calls `syncAndGetUnlockedAchievements` so preserved Steam unlocks are read again before the update event is published. The Steam-specific confirmation text also matches that behavior.

Validation completed:
- `yarn eslint` on all changed TypeScript and TSX files
- Prettier checks on all changed code and locale files
- `yarn typecheck:node`
- `yarn typecheck:web`
- `git diff --check`

The documented API dependency is PR #601 at head `c5edc997686e157e5efe74e63e43ac90f0e33ed5`; I assessed this launcher change against that contract without requiring the API PR to merge first. Release order remains API #601 first, then launcher #2796.

PR #2797, opened 2026-09-16
feat(steam): detect imported game executables

Hachi-R:
## Problem
Steam-imported games can enter the local library without an executable path even when the game is installed by Steam, forcing manual setup and preventing local process tracking.

## Changes
- reads installed app manifests from every configured Steam library
- matches each active imported game by its exact Steam App ID
- searches only that app's declared install directory using the existing executable catalogue and ranking
- persists the detected executable after rechecking current local state
- keeps Steam sync successful when local manifest or executable detection fails

## Validation
- secondary Steam library detection
- mismatched manifest App ID rejection
- install-directory traversal rejection
- node and web typechecks

Stacked on #2796.


chubbygrannychaser (changes requested):
## Summary
REQUEST_CHANGES for current head `4a998724df6316f1be6d18ca97fca6aedb4c5712`.

## Blocking issue

- `src/main/services/steam-integration/link-imported-steam-executables.ts:53-56` reads `currentGame` and then writes the full snapshot with `gamesSublevel.put`. This bypasses the existing per-game serialized `updateGameRecord` path. During a Steam sync, the process watcher, Steam exit sync, or another merge can update the same game after line 42 but before line 53. The linker would then overwrite those newer fields, including playtime and unsynced deltas, tracking executable paths, or a refreshed import state. Use the shared serialized updater (for example, `updateGameRecord(gameKey, (game) => updateGameExecutablePath(game, executablePath))`) and keep the active-import/executable guard in that serialized read-modify-write operation. This preserves concurrent local state while still applying the detected path.

## Validation

- Focused manifest tests: 3 passed.
- Node typecheck: passed.
- Web typecheck: passed.
- `git diff --check`: passed.

chubbygrannychaser (approved):
## Summary

Re-reviewed PR #2797 at exact head `e0828a4ec399edca23cfab41f51a3692b9a2998a`. The prior stale-write blocker is fixed: the active-import and executable guards, together with `updateGameExecutablePath`, now run inside the shared serialized `updateGameRecord` read-modify-write callback. The detached cloud-save sync also catches and logs rejection. No blocking issues remain.

## Validation

- Focused Steam installation tests: 3 passed.
- Node typecheck: passed.
- `git diff --check`: passed.
- Hosted checks were still in progress at review time; this approval is based on the current-head code review and the local validation above.

PR #2798, opened 2026-09-16
feat(steam): launch installed games via Steam

Hachi-R:
## Problem
Launching a Steam-installed game by its executable can bypass Steam initialization, DRM, game-specific launch setup, and the Steam Proton prefix.

## Changes
- launches `shop: steam` games through the Steam protocol when the configured executable is inside the exact appmanifest installation for the same App ID
- keeps the executable path for process, playtime, achievement, and exit-sync tracking
- passes plain game arguments through `steam://run`; command wrappers remain delegated to Steam settings
- uses the Steam `compatdata/<appId>/pfx` prefix for Hydra Cloud and process matching on Linux
- falls back to direct launch if protocol dispatch fails where that fallback does not risk a Proton-prefix mismatch
- rejects sibling folders, traversal, mismatched App IDs, and symlink escapes

## Validation
- Windows, Linux, and macOS path containment
- primary and secondary Steam libraries
- launch URL arguments and wrapper behavior
- Steam compatibility root and `/pfx` process matching
- 12 focused tests
- node and web typechecks

Stacked on #2797.


chubbygrannychaser (changes requested):
## Summary

Request changes on exact head `183f572e04bd395a42f224307e215e5cb3705b21`.

The Steam installation, URL argument preservation, serialized record writes, Proton-prefix matching, and platform-gated paths look consistent. The focused validation passes. One blocking Linux fallback regression remains.

## Critical Issue

- `src/main/helpers/launch-game.ts:820-828` — If `shell.openExternal(steamProtocolLaunch.url)` fails for an eligible Linux Windows Steam game, this branch clears the cloud-save guard, restores the previous `winePrefixPath`, closes the launcher, and returns `null`. It never calls `launchResolvedGame`, so Steam being unavailable or rejecting the protocol makes the game silently not launch. The previous flow still attempted `Umu.launchExecutable` and then Wine. Preserve that fallback, using the already derived Steam compatibility prefix and its cloud-save safety state, or otherwise provide an explicit safe direct-launch fallback instead of returning success-like `null` without launching. Add a regression test for protocol-open failure and verify that the fallback receives the correct Proton prefix.

## Validation

- `node --import ./scripts/register-ts-node.mjs --test src/main/services/steam-integration/steam-installation.test.ts src/main/services/linux-process-match.test.ts` — 12 passed.
- `corepack yarn eslint src/main/helpers/launch-game.ts src/main/services/linux-process-match.test.ts src/main/services/linux-process-match.ts src/main/services/process-watcher.ts src/main/services/steam-integration/steam-installation-core.ts src/main/services/steam-integration/steam-installation.test.ts src/main/services/steam-integration/steam-protocol-launch.ts` — passed.
- `corepack yarn typecheck:node` — passed.
- `git diff --check` — passed.
- Hosted build, lint, and Windows/Ubuntu checks were still in progress at the time of review; skipped production build and text-integrity success were not treated as full green CI.

chubbygrannychaser (approved):
Re-reviewed the current head 9c0804e266b3599311c6c2508701019159a30051.

The prior Linux protocol-failure blocker is fixed. When Steam protocol dispatch rejects the URL, the fallback now calls launchResolvedGame with the already prepared Steam compatibility context. The new focused tests cover the correct Proton prefix on fallback and skipping fallback after success. Mixed %command% trailing arguments remain preserved.

No blocking issues remain.

PR #2800, opened 2026-09-16
fix(steam): deduplicate imported playtime

Hachi-R:
## Problem

Steam-imported games only avoided local Hydra playtime when their executable was inside a detected Steam library. Imported games launched from any other path counted Hydra time by default, and the negative checkbox made the behavior unclear.

## Changes

- deduplicate local playtime by default for every game with an active Steam import
- replace the per-game opt-out with an explicit opt-in to also count Hydra playtime
- keep Steam exit sync limited to executable paths inside a Steam library
- explain in English and Brazilian Portuguese that the game was imported and Steam playtime already syncs
- rename the persisted preference and IPC contract to match the positive behavior

## Validation

- 12 focused Steam policy and persistence tests passed
- node and web typechecks passed
- ESLint passed with existing warnings only
- Prettier and `git diff --check` passed
- full suite: 667/669 passed; the two existing Cloud Save custom-path tests still fail on the local macOS symlink layout

Targets `feat/steam-connection`. No API change.


chubbygrannychaser (approved):
Reviewed current head `0632c60592912d27387588659759a410c9c779dd`.

The Steam playtime policy now skips local Hydra counting for every active Steam import by default, independent of executable path. The explicit per-game opt-in enables Hydra counting without changing path-gated Steam exit sync. The stale negative preference and IPC names are removed, English and pt-BR copy describe the new behavior, and focused tests cover the policy, lookup fallback, path gating, and delta handling.

Focused tests, typechecks, ESLint, Prettier, and diff hygiene passed. Hosted build and lint checks were still in progress when reviewed.

PR #2804, opened 2026-09-16
feat(steam): upload achievements in chunks

Hachi-R:
## Problem

Steam games can expose more than 2,000 unlocked achievements, while each API request accepts at most 2,000. A single snapshot request can therefore fail and leave the game unsynced.

## Changes

- split full Steam snapshots into ordered chunks of at most 2,000 achievements
- stage every chunk and commit the snapshot atomically
- split per-game exit sync payloads into additive requests
- preserve the difference between skipped collection and a confirmed empty achievement list
- fall back when chunk routes are unavailable: publish the compatible legacy snapshot, then upload every oversized game's complete achievement list through the existing incremental game endpoint in 2,000-item chunks
- document chunk indexes, commit behavior, retry/resume semantics, and the compatibility path

The chunked path targets `hydralauncher/hydra-api#602`. The fallback keeps the launcher compatible before that API is deployed without dropping oversized achievement lists.

## Validation

- 15 focused Steam snapshot tests, including exact ordered preservation of 2,001 achievements through the fallback
- `corepack yarn typecheck:node`
- `corepack yarn typecheck:web`
- ESLint and Prettier checks
- `git diff --check`


src/main/services/steam-integration/steam-sync-orchestrator.ts
```diff
@@ -889,23 +882,40 @@ class SteamSyncOrchestrator {
...
-    await HydraApi.put(
-      `${INTEGRATION_ENDPOINT}/sync/${syncRunId}/snapshot`,
-      snapshot,
-      { signal }
+    await uploadSteamSnapshotChunks(
+      chunks,
+      async (chunk, chunkIndex) => {
+        throwIfAborted(signal);
+        await HydraApi.put(
+          `${INTEGRATION_ENDPOINT}/sync/${syncRunId}/snapshot/chunks/${chunkIndex}`,
```
chubbygrannychaser: Blocking: this endpoint is not present on the current `hydralauncher/hydra-api` main revision (`1bc65d89`). The compatible route exists only in open API PR #602 (`bd5347f290d461bcfbee80281561463bf585e280`). Merge and deploy that API contract before shipping this client call, or keep a fallback to the existing snapshot endpoint; otherwise large-achievement syncs receive 404 and never commit.

chubbygrannychaser (changes requested):
## Summary

Request changes for the current head `3a1214b471a8118b37bf4a1f759ff0db9a0abc32`.

The chunking implementation is deterministic and the local helper tests cover empty snapshots, chunk boundaries, ordering, upload sequencing, and no-commit-after-failure. However, the client now depends on a backend contract that is not available on the current `hydralauncher/hydra-api` main revision. The matching backend implementation is still an open pull request (#602), so this change is not independently shippable and large-achievement syncs will fail against the current API.

## Blocking issue

- `src/main/services/steam-integration/steam-sync-orchestrator.ts:906` calls `PUT /profile/integrations/steam/sync/:syncRunId/snapshot/chunks/:chunkIndex` and the following commit call uses `POST /profile/integrations/steam/sync/:syncRunId/snapshot/commit`. I verified that the current `hydralauncher/hydra-api` main commit `1bc65d89` exposes only the original `PUT /profile/integrations/steam/sync/:syncRunId/snapshot` route; the new routes exist only in open API PR #602 (`bd5347f290d461bcfbee80281561463bf585e280`). Merge and deploy the compatible API contract before shipping this desktop change, or retain a compatible fallback. Without that, any snapshot requiring the new path receives a 404 and never commits, so the Steam library and achievements are not uploaded.

## Validation

- `node --import ./scripts/register-ts-node.mjs --test src/main/services/steam-integration/steam-sync-snapshot.test.ts` — 13 passed.
- `corepack yarn typecheck:node` — passed.
- `corepack yarn eslint src/main/services/steam-integration/steam-game-exit-sync.ts src/main/services/steam-integration/steam-sync-orchestrator.ts src/main/services/steam-integration/steam-sync-snapshot.ts src/main/services/steam-integration/steam-sync-snapshot.test.ts src/types/steam-integration.types.ts` — passed.
- `corepack yarn prettier --check` on all five changed files — passed.
- `git diff --check` — passed.

chubbygrannychaser (changes requested):
## Review summary

Request changes for the current head `893eaf7dbd489b5fe288e46bb93c8a87ed41d680`.

### Blocking issue

- `src/main/services/steam-integration/steam-sync-orchestrator.ts:886-925` — The client now relies on the chunk and commit routes from Hydra API PR #602, but that API PR is still open and unmerged (`hydralauncher/hydra-api#602`, head `bd5347f290d461bcfbee80281561463bf585e280`). Until those routes are deployed, every sync takes the documented 404 fallback and `buildLegacySteamSnapshot` removes achievements for games with more than 2,000 unlocks. The client therefore does not deliver the advertised preservation of those achievements in the currently available API, even though library and playtime still sync. Please coordinate this change with a merged and deployed API #602 before shipping the client behavior, or otherwise make the release dependency and expected loss of oversized achievements explicit and gated.

### Validation

- Focused snapshot tests: 15 passed.
- Node typecheck: passed.
- `git diff --check`: passed.
- Hosted checks at review time: `text-integrity` passed; `lint` and both platform builds were still in progress; `build-production` was skipped.

chubbygrannychaser (approved):
Reviewed the complete diff at current head `3472ccc6739bda7b2b47b807058152167561e5f7`.

The prior oversized-achievement compatibility blocker is fixed. When chunk routes return 404, the launcher publishes the legacy-compatible snapshot first, then uploads every oversized game's complete achievement list through the existing incremental per-game PUT endpoint in ordered chunks of at most 2,000. The 2,001-achievement coverage verifies exact ordered preservation without duplicates. The stacked base `feat/steam-connection` already provides the incremental endpoint; the atomic chunk API is only needed for the primary path.

Validation passed:
- 15 focused Steam snapshot tests
- `corepack yarn typecheck:node`
- `corepack yarn typecheck:web`
- ESLint on changed TypeScript files
- Prettier checks on changed files
- `git diff --check`

No blocking issues remain.

PR #2805, opened 2026-09-16
fix(settings): show Steam playtime toggle once

Hachi-R:
## Problem

The Steam playtime opt-in appears in both General and Locations because both tabs render the same reusable settings section.

## Change

Add an explicit visibility prop and hide the Steam playtime block from Locations. The option remains available once in General.

## Validation

- `corepack yarn typecheck:web`
- full pre-push lint and typechecks
- `git diff --check`


chubbygrannychaser (approved):
Reviewed at exact head `53ad1f7754ec6f2ea1e3d97a6a69e9e2943f5a09`.

The Steam playtime opt-in is rendered once through the General tab's `GeneralSettingsSection` usage. The reused Locations usage explicitly sets `showSteamPlaytimeSection={false}`, so the control remains hidden there. The component has no other renderer usages that would be affected by the new optional prop.

Validation completed:
- Focused ESLint on both changed TSX files: passed with 16 existing warnings and no errors.
- Web TypeScript typecheck: passed.
- Prettier check on both changed TSX files: passed.
- `git diff --check`: passed.

No blockers found.

PR #2806, opened 2026-09-16
fix(playtime): apply Steam opt-in immediately

Hachi-R:
## Problem

Enabling `Also count playtime in Hydra for this game` while the game was already running saved the preference, but the active session kept its old dedup decision. Hydra only started counting after the game restarted, making the option appear broken.

## Change

- Enable Hydra tracking immediately for the active game session when the opt-in is turned on.
- Reset session tick baselines at that moment so time spent before opting in is not counted retroactively.
- Keep the stored preference for future sessions.

## Validation

- `node --import ./scripts/register-ts-node.mjs --test src/main/services/game-running-state.test.ts`
- `npm run typecheck:node`
- pre-push full lint and typechecks


chubbygrannychaser (approved):
Reviewed the complete stacked diff against base commit `53ad1f7754ec6f2ea1e3d97a6a69e9e2943f5a09` at exact head `d6c4ef98bc8e437013d608e7ca7cdbc3ee4ba93a`.

The new running-session helper preserves the session start timestamp and begins both local and sync deltas at the moment Steam playtime is enabled. It also leaves already-counting sessions unchanged and safely ignores non-running games. The IPC handler updates persisted state before applying the in-memory transition.

No blocking correctness or regression issues found.

PR #2808, opened 2026-09-16
fix(steam): clear imported state on disconnect

Hachi-R:
## Problem

Disconnecting Steam with imported-data removal could leave Steam-only games in the library after automatic executable detection. Retained games could also keep stale Steam playtime, last-session data, source, and import state, which could affect playtime deduplication.

## Change

- Treat the server's pre-disconnect Steam-only list as authoritative even when an executable was auto-linked.
- Preserve games with Hydra playtime or a Hydra library origin.
- Clear local Steam playtime, last-session state, import flag, achievement cache, and Steam-only source before the remote Hydra merge.
- Let the merge restore the retained game's Hydra last-session data.
- Keep preserved-snapshot disconnect behavior unchanged.

## Validation

- focused disconnect cleanup and playtime merge tests: 16 passed
- `npm run typecheck:node`
- pre-push full lint and node/web typechecks


chubbygrannychaser (changes requested):
REQUEST_CHANGES: disconnect cleanup can erase Hydra session data.

In `getSteamImportedDataCleanup`, `lastTimePlayed` is unconditionally set to `null` for every Steam game that has imported state. This also affects a Hydra-owned game (`source !== "steam"`) that has a real Hydra play session, because `hasImportedSteamData` includes `hasActiveSteamImport` and Steam playtime. `disconnectSteam(true)` persists that null before calling `mergeWithRemoteGames`; the merge catches and suppresses remote failures, so a failed or incomplete refresh leaves the local Hydra last-session value lost. Preserve Hydra-owned `lastTimePlayed` and clear only Steam-owned session fields, or otherwise make restoration failure-safe. Add a regression test covering a Hydra-owned game with imported Steam state and an existing `lastTimePlayed`.

chubbygrannychaser (approved):
Re-reviewed the exact current head ccab76fa361a28ab40c52bb08c3108e9047c04c5. The earlier lastTimePlayed data-loss finding on 0716b1555 is fixed: cleanup snapshots the prior value and restores it for retained non-deleted games when remote merge fails. Steam-only removal now takes precedence over local executable metadata, while Hydra-owned games with Hydra playtime are retained. Cleanup clears active Steam import, Steam playtime, achievements, and Steam source state. Focused lint, Node typecheck, Prettier, and git diff --check pass. The focused test could not start in this environment because Node v24.19.0 fails while loading the repository ts-node/esm test loader; this is an environment limitation, not a code finding.

chubbygrannychaser (approved):
Re-reviewed the exact current head 156b35a142db2e8f904d3585a6138c555f55430a. The prior disconnect cleanup data-loss finding remains fixed. The cleanup helper now accepts fixtures without lastTimePlayed, while production callers still preserve and restore Hydra-owned session timestamps when remote merge fails. The focused tests cover Steam-only removal, retained local installs, imported-state cleanup, and the Hydra fallback. No blocking issues found. Focused test, Node typecheck, ESLint, Prettier, and git diff --check pass.

